Issue #58, opened 2026-05-21
Databricks App MCP servers fail silently in `ucode configure mcp`


zjacobson1016:
## Summary

Selecting a Databricks App MCP server in `ucode configure mcp` appears to succeed (no error shown), but the server is not registered with coding agents and may not be saved to ucode state.

## Expected behavior

When a user selects a discovered Databricks App MCP server (apps whose name starts with `mcp-`), ucode should register it with configured MCP clients (Claude, Codex, etc.) and persist it in `~/.ucode/state.json`.

## Actual behavior

- User toggles an App MCP server in the picker and presses Enter
- No error is displayed
- Server is not added to agent MCP config (e.g. `~/.claude.json` for Claude Code)
- Server may be missing from `mcp_servers` in ucode state

## Suspected cause

In `configure_mcp_command()`, when `_resolve_mcp_selection()` returns `None` for an app selection, the code currently `continue`s without surfacing a warning or error:

```python
resolved = _resolve_mcp_selection(selection, workspace, available_app_mcp_servers)
if resolved is None:
    continue
```

App MCP resolution depends on matching the selection prefix against entries in `available_app_mcp_servers` keyed as `databricks-app-{app_name}`. Any mismatch between picker values and discovered app metadata causes a silent skip.

## Reproduction

1. Deploy a Databricks App with MCP enabled (app name prefixed with `mcp-`)
2. Run `ucode configure` for at least one MCP-capable agent
3. Run `ucode configure mcp`
4. Select the App MCP server from the picker
5. Observe no error, but server not registered

## Suggested fix

- Print a warning/error when `_resolve_mcp_selection()` fails instead of silently continuing
- Verify app name / URL resolution logic in `_resolve_mcp_selection()` and `build_mcp_picker_choices()` for Databricks Apps
- Add test coverage for app MCP registration in `configure_mcp_command()`

## Environment

- `ucode` (latest)
- Databricks workspace with Databricks Apps MCP
- MCP clients: Claude Code, Codex, etc.

rohita5l:
cc @AarushiShah-db

Baijack-star:
I took a quick source-level pass on current `main`. One useful narrowing point: the happy path for App MCP registration looks covered and currently passes.

What I checked:

- `app_mcp_servers()` turns an app named `mcp-my-app` into server name `databricks-app-mcp-my-app` and URL `<app-url>/mcp`.
[...]

PR #62, opened 2026-05-21
Pass --profile to databricks CLI listing commands (fixes MCP discovery with CLI v1.0.0)

bkvarda:
## Problem

`ucode configure mcp` silently skips external connections, Genie spaces, and Databricks apps with \"Skipped\" warnings, leaving the MCP picker empty or showing only \"Databricks SQL\". This is the root cause of the discovery side of #58.

Root cause: Databricks CLI v1.0.0 ([release notes](https://github.com/databricks/cli/releases/tag/v1.0.0)) changed OAuth token storage from `~/.databricks/token-cache.json` to the OS-native keychain (Keychain on macOS, Credential Manager on Windows). Previously, passing `DATABRICKS_HOST` as an env var was sufficient for the CLI to look up the cached token by host URL. With v1.0.0, tokens are keyed by profile name in the keychain — `DATABRICKS_HOST` alone returns:

```
Error: default auth: cannot configure default credentials
```

`get_databricks_token` and `ensure_databricks_auth` already handled this correctly by calling `find_profile_name_for_host` and passing `--profile`. The three listing functions used for MCP discovery missed the same pattern.

## Fix

`list_databricks_connections`, `list_genie_spaces`, and `list_databricks_apps` now call `find_profile_name_for_host` and append `--profile <name>` when a match is found, matching the pattern already used by `get_databricks_token` and `ensure_databricks_auth`.

Detecting the error and warning was considered but rejected — it requires fragile string-matching on CLI error output that can change between versions. Always passing `--profile` is simpler and correct across all CLI versions.

## Related

- Fixes discovery prerequisite of #58 (Databricks App MCP servers fail silently — apps were never reaching the picker due to this auth failure)

## Test plan

- [ ] `uv run pytest` — all 566 tests pass
- [ ] Manual: `ucode configure mcp` on a workspace with CLI v1.0.0 installed now discovers external connections, Genie spaces, and Databricks apps
- [ ] Manual: verified against external connections in a Databricks workspace - MCP server now shows in `~/.config/goose/config.yaml`

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -380,6 +380,7 @@ def _extract_connection_page(payload: object) -> tuple[list[dict], str | None]:
 
 def list_databricks_connections(workspace: str) -> list[dict]:
     env = build_databricks_cli_env(workspace)
+    profile_name = find_profile_name_for_host(workspace)
     connections: list[dict] = []
     page_token: str | None = None
     seen_page_tokens: set[str] = set()
@@ -395,6 +396,8 @@ def list_databricks_connections(workspace: str) -> list[dict]:
                 "--output",
                 "json",
             ]
+            if profile_name:
+                cmd.extend(["--profile", profile_name])
             if page_token:
                 cmd.extend(["--page-token", page_token])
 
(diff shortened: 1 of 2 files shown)


rohita5l:
@bkvarda Thanks for the contribution. Do you mind rebasing with latest master ?
@AarushiShah-db Can you review this PR ?

AarushiShah-db:
@bkvarda I think this was already resolved by https://github.com/databricks/ucode/pull/94 -> can rebase onto latest master and see if the issue is fixed?

bkvarda:
Confirmed — this is the same fix as #94 (merged), which threads `--profile` through `list_databricks_connections`, `list_genie_spaces`, and `list_databricks_apps`. Closing as redundant. Thanks @AarushiShah-db for the pointer.

PR #86, opened 2026-05-26
 Fix configure/launch when ~/.databrickscfg has duplicate hosts

rohita5l:
## Summary
  - Thread the Databricks CLI profile name through every `databricks` invocation
    (`auth token`, `auth login --no-browser` retry, `build_auth_shell_command`,
    and the web-search MCP subprocess via `DATABRICKS_CONFIG_PROFILE`). Only
    `run_databricks_login` was passing `--profile` before, so any workspace whose
    `~/.databrickscfg` had two profiles (e.g. `DEFAULT` + a named alias) pointing
    at the same host hit `"DEFAULT and <other> match ... Use --profile to specify"`
    and could never finish `configure` or `launch`.
  - Persist the chosen profile name in state under `"profile"` so subsequent
    commands and the Claude Code token-refresh hook disambiguate the same way.
    Falls back to `find_profile_name_for_host` when `--workspaces` is given a
    bare URL.
  - Add `UCODE_DEBUG` logging to `get_databricks_profiles` so the silent
    "workspace picker degrades to a blank URL prompt" case is diagnosable from
    `~/.ucode/debug.log` (logs subprocess error / non-zero rc / JSON decode /
    all-PAT counts). Bumped its timeout from 10s to 20s.

  ## Test plan
  - [x] `~/.databrickscfg` with `DEFAULT` + a named profile pointing at the same
        host: `uv run ucode configure` reaches AI Gateway verification without
        the "match" error.
  - [x] Same setup: `uv run ucode launch claude` succeeds; the auth-refresh
        command stored in state includes `--profile <name>`.
  - [x] `--workspaces https://...` (bare URL, no profile picker): profile is
        resolved post-login from `find_profile_name_for_host` and threaded
        through.
  - [x] `UCODE_DEBUG=1 uv run ucode configure` with no profiles → blank prompt
        appears AND `~/.ucode/debug.log` shows
        `get_databricks_profiles: returned=0 total=0 pat=0`.
  - [x] Web-search MCP inside Claude Code refreshes its token without the
        duplicate-profile error.
  - [x] `uv run pytest` passes (including the updated `test_e2e.py` mock that
        now returns `(workspace, None)`).

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -59,19 +59,22 @@ def _resolve_web_search_model(state: dict) -> str | None:
 WEB_SEARCH_MCP_NAME = "web_search"
 
 
-def _web_search_mcp_entry(workspace: str, search_model: str) -> dict:
+def _web_search_mcp_entry(workspace: str, search_model: str, profile: str | None = None) -> dict:
     """Stdio MCP server entry pointing at `ucode mcp web-search`. Resolves
     the absolute path to the `ucode` binary so launchers without the right
     PATH (e.g. desktop GUI launchers) still find it."""
     ucode_binary = shutil.which("ucode") or "ucode"
+    env: dict[str, str] = {
+        "DATABRICKS_HOST": workspace,
+        "UCODE_WEB_SEARCH_MODEL": search_model,
+    }
+    if profile:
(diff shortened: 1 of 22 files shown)


rohita5l:
Fixes https://github.com/databricks/ucode/issues/78

AarushiShah-db (approved):

PR #91, opened 2026-05-26
Update user permissions error message

asujithan:
Catch ServerOperationError from the SQL connector inside run_usage_query and, when the error indicates the caller can't read system.ai_gateway.usage, raise a one-line actionable RuntimeError ("Ask your workspace admin to enable READ access to system.ai_gateway.usage for your account") instead of the raw [INSUFFICIENT_PERMISSIONS] ... SQLSTATE: 42501 text the connector emits. Previously ucode usage surfaced the unmodified connector message, which named the failure but not what to do about it.

Add a private _is_usage_table_access_error helper that lowercases the error string, strips backticks, square brackets, and single/double quotes, and returns True iff the result contains both system.ai_gateway and insufficient_permissions. The quote-stripping collapses identifier-quoting variants the connector may emit (`system`.`ai_gateway`.`usage`, [system].[ai_gateway].[usage], 'system.ai_gateway.usage') to one canonical form. Matching the system.ai_gateway prefix rather than the full system.ai_gateway.usage is deliberate: schema-level denials (USE SCHEMA on system.ai_gateway) never name .usage in the error, but are still effectively the same problem from the user's perspective. Permission errors on unrelated catalogs/tables fall through unchanged to the existing Usage query failed.

**Test Plan**
- Account without SELECT on system.ai_gateway.usage: uv run ucode usage exits with the new actionable message ("Ask your workspace admin to enable READ access ...") instead of the raw [INSUFFICIENT_PERMISSIONS] ... SQLSTATE: 42501 text.
- Identifier-quoting variants from the SQL connector all trip the matcher: backticked (`system`.`ai_gateway`.`usage`), bracketed ([system].[ai_gateway].[usage]), and single-quoted ('system.ai_gateway.usage') forms each route to the friendly message.
- Non-ServerOperationError failures (network timeout, malformed SQL from a future refactor, generic RuntimeError from the connector) still surface through the generic Usage query failed: {exc} arm so existing behavior is preserved for everything outside the permissions case.
- uv run pytest tests/test_databricks.py -q passes — covers _is_usage_table_access_error (positive: table-level and schema-level denials with each quoting variant; negative: unrelated catalog/table denials) and run_usage_query (raises friendly text for matching errors, falls through for non-matching ones).
- passes uv run ruff format src/ tests/ && uv run ruff check .

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -19,6 +19,8 @@ from urllib import error as urllib_error
 from urllib import request as urllib_request
 from urllib.parse import urlparse
 
+from databricks.sql.exc import ServerOperationError
+
 from ucode.config_io import APP_DIR
 from ucode.ui import (
     err_console,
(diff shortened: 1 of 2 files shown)


AarushiShah-db (commented):
can you add a unit test for this please, ty!

AarushiShah-db (approved):

AarushiShah-db:
@asujithan can you rebase onto latest master, the e2e should pass now! good to merge after the tests pass

PR #102, opened 2026-05-27
Disable Claude Code thinking to unblock Opus 4.7 gateway calls

jxk590:
Claude Code sends `thinking.type=enabled` on outbound requests, which the Databricks AI Gateway rejects for adaptive-only models (Opus 4.7 family) with a 400. The gateway expects `thinking.type=adaptive` + `output_config.effort`. Claude Code does support that shape but only when its model-name pattern matches a known Anthropic id — the gateway exposes models as `databricks-claude-opus-4-7`, so the prefix throws off detection and Claude Code falls back to the legacy `enabled` shape. The result: `ucode claude` validation fails and the config is reverted.

Set `CLAUDE_CODE_DISABLE_THINKING=1` in the Claude Code env block so the field is omitted entirely. The gateway still applies adaptive reasoning server-side on Opus 4.7 (where it's mandatory), so reasoning quality is preserved on the new family. Older models lose client-side extended thinking through the gateway, which is the correct tradeoff for a configuration that actually works end-to-end.

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -106,6 +106,13 @@ def render_overlay(
         "ANTHROPIC_BASE_URL": base_url,
         "ANTHROPIC_CUSTOM_HEADERS": custom_headers,
         "CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS": "1",
+        # Suppress Claude Code's client-side `thinking.type=enabled` field.
+        # The Databricks AI Gateway rejects it for adaptive-only models (Opus 4.7
+        # family) which require `thinking.type=adaptive` + `output_config.effort`.
+        # Claude Code's model detection doesn't recognize the `databricks-` prefix
+        # on gateway model ids, so it sends the legacy shape and the request 400s.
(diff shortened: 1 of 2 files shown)


AarushiShah-db:
@jxk590 I'm not able to reproduce the original issue - can you please share a detailed repro of how you came across this issue?

jxk590:
@AarushiShah-db I received a 400 error when attempting to run `ucode claude`:
              
✔ Databricks auth already available for https://adb-xxxxxxxxxxxxxxxxxxx.x.azuredatabricks.net
✔ Unity AI Gateway detected      
╭──────────────────── Configuration Complete ────────────────────╮
│ Workspace: https://adb-xxxxxxxxxxxxxxxxxxx.x.azuredatabricks.net │
│ Claude Code: configured                                        │
╰────────────────────────────────────────────────────────────────╯
ERROR Claude Code: API Error: 400 {"message":"\"thinking.type.enabled\" is not supported for this model. Use \"thinking.type.adaptive\" and \"output_config.effort\" to control thinking behavior."}
ERROR Claude Code validation failed — config reverted.

PR #107, opened 2026-05-27
Add Cursor IDE support via Databricks AI Gateway

zjacobson1016:
## Summary

- Adds `ucode cursor` to configure Cursor IDE with Databricks AI Gateway foundation models (`cursor/v1/chat/completions`)
- Writes reference config to `~/.ucode/cursor-databricks.json` (base URL, OAuth token, model list) with manual setup steps for Cursor Settings → Models
- Resolves macOS launch via the app-bundle CLI when `~/.local/bin/cursor` is an agent-only shim

## Test plan

- [x] `uv run pytest tests/test_agent_cursor.py tests/test_agents_init.py`
- [ ] `ucode cursor` against a workspace with Cursor-compatible foundation-model endpoints
- [ ] Verify Cursor opens on macOS with `/Applications/Cursor.app` installed


rohita5l:
Seems like this PR is touching a lot of files that are unnecessary ?

zjacobson1016:
You are correct. Can you reject this PR and I will submit a new one. Not sure what happened here

PR #114, opened 2026-05-28
Show profile name alongside workspace URL in picker

anthonyivn2:
## Summary

Reworks the workspace picker shown during `ucode configure` / first launch so it mirrors the layout of `databricks auth profiles` (without the `Valid` column) and so the *exact* profile the user selects is what later `databricks` CLI calls use.

## Changes:
- **Remove deduping of profiles by host.** `get_databricks_profiles()` previously collapsed multiple profiles pointing at the same workspace down to the first match. It now returns one row per non-PAT profile, so each profile is selectable.
- **2-column picker with header.** `prompt_for_workspace()` renders a `Profile Name` / `Workspace URL` header via `questionary.Separator` and ljust-pads the profile names so columns line up.
- **Pass the chosen profile through.** The picker already returns `(host, profile_name)`; that tuple now travels through `configure_shared_state(profile=…)` into `state.json` and into every `--profile`-bearing CLI invocation. Previously, when a host had multiple profiles, downstream code fell back to "first profile for host" via `find_profile_name_for_host`, now this is determined solely by the user's selection

Before:
```
› https://workspace-a.cloud.databricks.com
  https://workspace-b.cloud.databricks.com
  https://workspace-c.cloud.databricks.com
  Enter a different URL
```
(only one row per host even when several profiles target it)

After:
```
    Profile Name        Workspace URL
›   profile-one         https://workspace-a.cloud.databricks.com
    profile-two         https://workspace-a.cloud.databricks.com
    profile-three       https://workspace-a.cloud.databricks.com
    profile-four        https://workspace-b.cloud.databricks.com
    profile-five        https://workspace-c.cloud.databricks.com
    Enter a different URL
```

## Test plan

- [x] `uv run pytest` (546 passed / 28 skipped) — adds `TestGetDatabricksProfiles` (duplicate-host preservation, PAT skip, trailing-slash strip, non-zero-exit) and `TestPromptForWorkspace` (header rendering, duplicate-host preservation, URL normalization).
- [x] `uv run ruff check .` clean.
- [x] `uv run ucode configure` — visually verify column alignment and that all profiles for a duplicated host show up.
- [x] Pick a non-default profile for a host that has multiple, then `uv run ucode status` — confirm `CLI profile` matches the selection (and `~/.ucode/state.json` `profile` key matches).
- [x] `uv run ucode revert && uv run ucode claude` — same picker fires on the launch path.

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -815,6 +815,11 @@ def list_profile_entries() -> list[dict]:
     """Return raw profile dicts ({"name", "host", "auth_type", ...}) from
     `databricks auth profiles`.
 
+    Each non-PAT profile is returned individually — duplicate hosts (multiple
+    profiles pointing at the same workspace) appear as separate entries so the
+    workspace picker can offer each profile by name. Order matches the CLI's
+    own ordering.
+
     Returns ``[]`` on any failure (CLI missing, timeout, non-zero exit, JSON
     decode error). When ``UCODE_DEBUG=1`` each dropout path logs *why* the
     result was empty so a silently-disappearing workspace picker is
@@ -846,8 +851,7 @@ def get_databricks_profiles() -> list[tuple[str, str]]:
     """Return [(host_url, profile_name), ...] from Databricks CLI profiles."""
     profiles = list_profile_entries()
 
-    # dict dedupes by host (first non-PAT profile wins).
-    out: dict[str, str] = {}
+    out: list[tuple[str, str]] = []
     pat = 0
     for p in profiles:
         host = (p.get("host") or "").rstrip("/")
@@ -857,13 +861,13 @@ def get_databricks_profiles() -> list[tuple[str, str]]:
         if p.get("auth_type") == "pat":
             pat += 1
             continue
-        out.setdefault(host, name)
+        out.append((host, name))
 
     _debug(
         "get_databricks_profiles",
         f"returned={len(out)} total={len(profiles)} pat={pat}",
     )
-    return list(out.items())
+    return out
 
 
 def find_profile_name_for_host(workspace: str) -> str | None:
--- a/src/ucode/ui.py
+++ b/src/ucode/ui.py
@@ -305,26 +305,34 @@ def prompt_for_workspace(
     """Ask the user for a workspace URL, offering profiles as quick-select.
 
     `profiles` is a list of (host_url, profile_name) tuples. Caller fetches
-    them — `ui.py` stays Databricks-agnostic. Returns ``(url, profile_name)``;
-    profile_name is ``None`` when the user typed a URL manually.
+    them — `ui.py` stays Databricks-agnostic. Duplicate hosts (multiple
+    profiles pointing at the same workspace) are shown separately; the picker
+    returns the exact (host, profile_name) the user selected. Returns
+    ``(url, profile_name)``; profile_name is ``None`` when the user typed a
+    URL manually.
     """
     console.print()
     console.print(Panel(description, title="ucode setup", style="bold blue", expand=False))
 
     if profiles:
-        choices = [
-            questionary.Choice(
-                title=f"{host}  ({profile_name})",
-                value=(host, profile_name),
-            )
-            for host, profile_name in profiles
+        name_header = "Profile Name"
+        url_header = "Workspace URL"
+        name_width = max(len(name_header), *(len(name) for _, name in profiles))
+        # Match the 2-char cursor gutter so the header line aligns with rows.
+        header_title = f"  {name_header.ljust(name_width)}  {url_header}"
+        choices: list[questionary.Choice | questionary.Separator] = [
+            questionary.Separator(header_title)
         ]
+        for host, profile_name in profiles:
+            row_title = f"{profile_name.ljust(name_width)}  {host}"
+            choices.append(questionary.Choice(title=row_title, value=(host, profile_name)))
         choices.append(questionary.Choice(title="Enter a different URL", value=None))
         style = questionary.Style(
             [
                 ("highlighted", "fg:cyan bold"),
                 ("pointer", "fg:cyan bold"),
                 ("answer", "fg:cyan"),
+                ("separator", "fg:white bold"),
             ]
         )
         choice = questionary.select(
--- a/src/ucode/ui.py
+++ b/src/ucode/ui.py
@@ -317,14 +317,28 @@ def prompt_for_workspace(
     if profiles:
         name_header = "Profile Name"
         url_header = "Workspace URL"
-        name_width = max(len(name_header), *(len(name) for _, name in profiles))
+        # Clamp so a single very long profile name can't push the URL column
+        # off-screen on an 80-col terminal — questionary doesn't wrap row
+        # titles cleanly, and a wrapped row breaks the picker visually.
+        max_name_width = 40
+        name_width = min(
+            max_name_width,
+            max(len(name_header), *(len(name) for _, name in profiles)),
+        )
         # Match the 2-char cursor gutter so the header line aligns with rows.
         header_title = f"  {name_header.ljust(name_width)}  {url_header}"
         choices: list[questionary.Choice | questionary.Separator] = [
             questionary.Separator(header_title)
         ]
         for host, profile_name in profiles:
-            row_title = f"{profile_name.ljust(name_width)}  {host}"
+            display_name = (
+                profile_name
+                if len(profile_name) <= name_width
+                else profile_name[: name_width - 1] + "…"
+            )
+            row_title = f"{display_name.ljust(name_width)}  {host}"
+            # Value carries the full untruncated profile name so downstream
+            # `--profile` calls always use the real name, not the display form.
             choices.append(questionary.Choice(title=row_title, value=(host, profile_name)))
         choices.append(questionary.Choice(title="Enter a different URL", value=None))
         style = questionary.Style(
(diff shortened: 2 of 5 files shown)


rohita5l:
Can we only do this when there are duplicates ? Else its a bit redundant

anthonyivn2:
@rohita5l the situation that I am facing is that I am using >5 workspaces in my local setup, and when I try to do the setup with ucode I find it really hard to know which workspace I need to select in order to target the workspace that I want to use.

The change also brings the information shown closer to what `databricks auth profiles` would show

rohita5l:
> @rohita5l the situation that I am facing is that I am using >5 workspaces in my local setup, and when I try to do the setup with ucode I find it really hard to know which workspace I need to select in order to target the workspace that I want to use.
> 
> The change also brings the information shown closer to what `databricks auth profiles` would show

Can you add a screenshot of what it looks like now ?

src/ucode/databricks.py
```diff
@@ -446,8 +443,7 @@ def get_databricks_profiles() -> list[tuple[str, str]]:
         _debug("get_databricks_profiles", f"json decode error: {exc.msg}")
         return []
 
-    # dict dedupes by host (first non-PAT profile wins).
-    out: dict[str, str] = {}
+    out: list[tuple[str, str]] = []
```
anthonyivn2, line 854: I am changing the data type used to a list so that we can have duplicated hosts, and we don't have to do the list() conversion at the end

anthonyivn2:
@rohita5l I have updated the PR based on what we discussed on Slack

src/ucode/databricks.py
```diff
@@ -457,13 +453,13 @@ def get_databricks_profiles() -> list[tuple[str, str]]:
...
 
     _debug(
         "get_databricks_profiles",
         f"returned={len(out)} total={len(profiles)} pat={pat}",
     )
-    return list(out.items())
+    return out
 
 
 def find_profile_name_for_host(workspace: str) -> str | None:
```
rohita5l, line 873: Is this being called by anywhere now ? If not can we delete it ?

anthonyivn2: Yes it is still being called in multiple places in the code outside of the function I am modifying, I think if we want to refactor this we should do it as part of a different PR

rohita5l (changes requested):
Could we add a regression test for the picker → configure_shared_state hop? Existing tests in TestConfigureAgentsSelection only exercise profile=None, so if a future refactor of _configure_shared_workspace_states
  (cli.py:232) drops the profile from the tuple unpack, nothing fails. Suggested test (drop into TestConfigureAgentsSelection in tests/test_cli.py):

  def test_picker_selected_profile_flows_to_configure_shared_state(self, monkeypatch):
      """Picker's (host, profile) tuple must reach configure_shared_state's
      `profile` kwarg, otherwise downstream --profile calls fall back to
      host-based resolution and silently pick the wrong profile."""
      import ucode.cli as cli_mod

      monkeypatch.setattr(
          cli_mod,
          "_prompt_for_configuration",
          lambda tool=None: ("https://shared.cloud.databricks.com", "picked-profile"),
      )
      captured: dict = {}

      def fake_configure_shared_state(workspace, profile=None, tools=None, force_login=False):
          captured["workspace"] = workspace
          captured["profile"] = profile
          return {**MINIMAL_STATE, "workspace": workspace, "profile": profile}

      monkeypatch.setattr(cli_mod, "configure_shared_state", fake_configure_shared_state)
      monkeypatch.setattr(cli_mod, "save_state", lambda state: None)
      monkeypatch.setattr(cli_mod, "check_gateway_endpoint", lambda state, tool: True)
      monkeypatch.setattr(cli_mod, "prompt_for_tools", lambda available: ["claude"])
      monkeypatch.setattr(cli_mod, "install_tool_binary", lambda *args, **kwargs: True)
      monkeypatch.setattr(
          cli_mod,
          "configure_selected_tools",
          lambda state, tools: {**state, "available_tools": tools},
      )
      monkeypatch.setattr(cli_mod, "validate_all_tools", lambda state: None)

      assert cli_mod.configure_workspace_command() == 0
      assert captured["profile"] == "picked-profile"

rohita5l (changes requested):
The "Enter a different URL" branch (ui.py:221-224) returns (url, None) unconditionally, which means downstream falls back to find_profile_name_for_host even when the URL the user typed happens to match a profile
  they already have. Since prompt_for_workspace already has profiles in scope, we can resolve the match inline without breaking the Databricks-agnostic boundary:

```
  while True:
      raw_value = console.input(f"  [bold]Workspace URL[/bold] {muted('›')} ").strip()
      try:
          url = normalize_workspace_url(raw_value)
      except ValueError as exc:
          print_err(str(exc))
          continue
      # If the typed URL matches a known profile, attach it so downstream
      # --profile calls disambiguate. We only auto-attach when exactly one
      # profile matches — multiple matches re-trigger the same ambiguity the
      # picker is meant to resolve, so fall back to None and let the existing
      # host-based resolution handle it (or, better, re-prompt).
      matches = [name for host, name in (profiles or []) if host == url]
      matched_profile = matches[0] if len(matches) == 1 else None
      return url, matched_profile
```
  This closes the gap for the common case (user types a URL they actually have a profile for). The ambiguous multi-match case stays on the existing find_profile_name_for_host path — fine for now, but worth a
  follow-up to re-prompt with the matching rows.

src/ucode/ui.py
```diff
@@ -181,23 +181,34 @@ def prompt_for_workspace(
...
     console.print()
     console.print(Panel(description, title="ucode setup", style="bold blue", expand=False))
 
     if profiles:
-        choices = [
-            questionary.Choice(title=host, value=(host, profile_name))
-            for host, profile_name in profiles
+        name_header = "Profile Name"
+        url_header = "Workspace URL"
+        name_width = max(len(name_header), *(len(name) for _, name in profiles))
```
rohita5l: name_width = max(len(name_header), *(len(name) for _, name in profiles))

  name_width is unbounded — whatever the longest profile name is becomes the column width. If someone has a ~/.databrickscfg entry like eng-staging-us-west-2-rohit-personal-sandbox (45 chars) plus a 40-char host
  URL, that row is ~87 chars. On an 80-column terminal:

  - questionary doesn't wrap rows cleanly — the cursor pointer and padding math assume a single line, so a wrapped row visually breaks the picker.
  - The header (Profile Name … Workspace URL) and rows can drift out of alignment.
  - Worst case the row runs off-screen and the URL is unreadable without resizing.

  Concrete fix — clamp + ellipsize:

```
  MAX_NAME_WIDTH = 40

  def _truncate(s: str, width: int) -> str:
      return s if len(s) <= width else s[: width - 1] + "…"

  name_width = min(MAX_NAME_WIDTH, max(len(name_header), *(len(name) for _, name in profiles)))
  header_title = f"  {name_header.ljust(name_width)}  {url_header}"
  choices: list[questionary.Choice | questionary.Separator] = [questionary.Separator(header_title)]
  for host, profile_name in profiles:
      row_title = f"{_truncate(profile_name, name_width).ljust(name_width)}  {host}"
      choices.append(questionary.Choice(title=row_title, value=(host, profile_name)))
```
  The value=(host, profile_name) tuple still carries the full untruncated name, so the selection passed to configure_shared_state is unchanged — only the display string is shortened.

anthonyivn2: Made the change

anthonyivn2:
@rohita5l I have made the changes to resolve the comments

anthonyivn2:
> The "Enter a different URL" branch (ui.py:221-224) returns (url, None) unconditionally, which means downstream falls back to find_profile_name_for_host even when the URL the user typed happens to match a profile they already have. Since prompt_for_workspace already has profiles in scope, we can resolve the match inline without breaking the Databricks-agnostic boundary:
> 
> ```
>   while True:
>       raw_value = console.input(f"  [bold]Workspace URL[/bold] {muted('›')} ").strip()
>       try:
>           url = normalize_workspace_url(raw_value)
>       except ValueError as exc:
>           print_err(str(exc))
>           continue
>       # If the typed URL matches a known profile, attach it so downstream
>       # --profile calls disambiguate. We only auto-attach when exactly one
>       # profile matches — multiple matches re-trigger the same ambiguity the
>       # picker is meant to resolve, so fall back to None and let the existing
>       # host-based resolution handle it (or, better, re-prompt).
>       matches = [name for host, name in (profiles or []) if host == url]
>       matched_profile = matches[0] if len(matches) == 1 else None
>       return url, matched_profile
> ```
> 
> This closes the gap for the common case (user types a URL they actually have a profile for). The ambiguous multi-match case stays on the existing find_profile_name_for_host path — fine for now, but worth a follow-up to re-prompt with the matching rows.

@rohita5l `configure_shared_state` already resolves profile=None via `find_profile_name_for_host` (cli.py:161-162), so the inline match would be duplicate logic — same end state in both single-match and multi-match cases (first match in ~/.databrickscfg order). Can we resolve this part in a different PR instead of this one?

anthonyivn2:
@rohita5l can you review this tomorrow?

rohita5l (dismissed):
LGTM but would also want @AarushiShah-db to take a look.

AarushiShah-db (dismissed):
lgtm

anthonyivn2:
@rohita5l @AarushiShah-db Can you help me run this workflow so that I can merge the code? 
https://github.com/databricks/ucode/actions/runs/26676634359

it doesn't seem I can execute the action

anthonyivn2:
@rohita5l can you approve this code and run the workflow again? I fixed the issue that the e2e test ran into (ruff formatting)

benc-vorwerk:
This would help a lot. One remaining usability gap is the “Enter a different URL” flow: after a workspace is added manually, later selection is still URL-only because there is no CLI profile name to display.
In environments with several opaque workspace URLs, it is hard to tell which workspace is which. Could this PR, or a follow-up, support a persistent optional display label for manually added workspaces?

A matching Databricks CLI profile name could remain the default label where available; otherwise, retain the user-provided alias. Keeping the URL visible alongside the label would preserve clarity and avoid ambiguity.

rohita5l (approved):

PR #118, opened 2026-05-28
Append [1m] to Opus >=4.6 and Sonnet 4.6 model for 1m context length support

AarushiShah-db:
## Summary

  - Add `[1m]` suffix when writing Claude Code settings for Opus 4.6+ and Sonnet 4.6 models.
  - Allow `databricks-claude-opus-4-8` to be discovered and selected as the default Opus model.
  - Add regression coverage for Claude settings rendering and Opus 4.8 discovery.

  ## Testing

  - `uv run pytest`
  - `uv run pytest tests/test_databricks.py`
  - `uv run ruff check src/ucode/agents/claude.py tests/test_agent_claude.py`
  - `uv run ruff check src/ucode/databricks.py tests/test_databricks.py`

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import re
 import shutil
 from pathlib import Path
 
@@ -58,6 +59,7 @@ def _resolve_web_search_model(state: dict) -> str | None:
 
 
 WEB_SEARCH_MCP_NAME = "web_search"
+_CLAUDE_MODEL_RE = re.compile(r"^databricks-claude-(opus|sonnet)-(\d+)-(\d+)(.*)$")
 
 
 def _web_search_mcp_entry(workspace: str, search_model: str, profile: str | None = None) -> dict:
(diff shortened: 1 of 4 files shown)


src/ucode/databricks.py
```diff
@@ -804,14 +804,6 @@ def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
     )
 
 
-# Model ids the AI Gateway advertises but currently rejects at launch with
-# "The provided model identifier is invalid." Filter them out at discovery so
-# agent harnesses don't try to boot with them.
```
rohita5l, line 809: Why remove this ? I added this because my ucode claude was busted today

src/ucode/agents/claude.py
```diff
@@ -58,6 +59,7 @@ def _resolve_web_search_model(state: dict) -> str | None:
 
 
 WEB_SEARCH_MCP_NAME = "web_search"
+_CLAUDE_MODEL_RE = re.compile(r"^databricks-claude-(opus|sonnet)-(\d+)-(\d+)(.*)$")
```
rohita5l, line 62: Why do we need this ?

rohita5l (approved):

src/ucode/databricks.py
AarushiShah-db: we launchd opus4.8 ~2 hours ago so this should work now

src/ucode/agents/claude.py
AarushiShah-db: need to know which version of opus/sonnet it is because 1m context window is only supported for versions 4.6 or greater

PR #127, opened 2026-05-29
user-friendly genie-space server names displayed

asujithan:
## Summary
Make Genie MCP server names readable instead of opaque space IDs. `databricks-genie-01f0b...` becomes `databricks-genie-bakehouse-genie-space` in `/mcp` inside Claude, Codex, Gemini, OpenCode, and Copilot.

`genie_mcp_servers()` now derives the name from the Genie space title (normalized to lowercase + hyphens), with a fallback to the raw `space_id` when there is no title or two titles collide in the same discovery pass. The MCP URL still carries the real `space_id`, so Genie routing is unchanged.

`_resolve_mcp_selection()` gained an `available_genie_servers` parameter and looks up the canonical URL by name (mirrors the existing apps pattern); `configure_mcp_command()` threads the discovered list through. A legacy fallback preserves the old `space_id`-in-selection path.

## Test plan
- [x] `uv run pytest tests/test_mcp.py` (39/39 pass; new test covers slug collision + empty-title fallback)
- [x] `uv run ruff check .`
- [x] On a real workspace: `ucode configure mcp` then `claude /mcp` shows friendly names.

### DIFF
--- a/src/ucode/mcp.py
+++ b/src/ucode/mcp.py
@@ -352,6 +352,31 @@ def discover_external_mcp_connection_names(workspace: str, profile: str | None =
     return external_mcp_connection_names(list_databricks_connections(workspace, profile))
 
 
+def _normalize_workspace_title(text: str) -> str:
+    """Collapse a Databricks workspace title to lowercase alphanumerics joined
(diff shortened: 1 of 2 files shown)


AarushiShah-db (approved):
ty for the fix!

PR #137, opened 2026-06-04
fix(windows): resolve Claude binary path and apiKeyHelper shell syntax on Windows

yakup-ozturk:
fix(windows): resolve Claude binary path and apiKeyHelper syntax on Windows:

Three changes that make `ucode claude` work on Windows without manual workarounds:

1. **WinError 2 — Claude binary not found** (`agents/claude.py`) Python's `subprocess` resolves PATH differently from cmd.exe and cannot find npm `.cmd` wrappers by name. Resolve `CLAUDE_BINARY` by walking from the `claude.cmd` wrapper to the actual `claude.exe` inside the npm package tree. Falls back to the wrapper path (or the bare `"claude"` string) if the .exe is absent, so non-Windows installs and non-standard npm layouts are unaffected.

2. **apiKeyHelper shell syntax error on Windows** (`databricks.py`) Claude Code runs the `apiKeyHelper` command via `cmd.exe` on Windows, which rejects the POSIX `[ -n "$VAR" ]` / `env -u` / `jq` syntax currently emitted by `build_auth_shell_command`. On `os.name == 'nt'` the function now returns a `python -c "..."` one-liner that handles the `DATABRICKS_BEARER` short- circuit and calls the Databricks CLI with `subprocess`, avoiding both the bash syntax and the `jq` dependency. POSIX behaviour is unchanged.

3. **Settings not visible to the Claude desktop / IDE extension** (`agents/claude.py`) `ucode claude` writes config to `~/.claude/ucode-settings.json` and passes `--settings` when launching the CLI, but the Claude desktop app and IDE extensions read `~/.claude/settings.json` by default.  After writing the ucode-managed file, also merge the auth/env overlay into `settings.json` so the `apiKeyHelper` and `ANTHROPIC_*` vars are available in every launch path. A backup of the pre-existing `settings.json` is created alongside the existing ucode-settings backup.
4. 
Reference: https://medium.com/@Yakup-Ozturk/getting-claude-code-to-work-with-the-databricks-ai-gateway-on-windows-three-fixes-i-wish-i-knew-f6e71ece06a9

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -16,6 +16,7 @@ from ucode.config_io import (
     backup_existing_file,
     deep_merge_dict,
     read_json_safe,
+    restore_file,
     write_json_file,
 )
 from ucode.databricks import (
@@ -30,36 +31,44 @@ from ucode.ui import print_note, print_success, print_warning
 
 CLAUDE_CONFIG_DIR = Path.home() / ".claude"
 CLAUDE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "ucode-settings.json"
-CLAUDE_DEFAULT_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "settings.json"
 CLAUDE_BACKUP_PATH = APP_DIR / "claude-ucode-settings.backup.json"
+# Claude Code's *default* settings file. ucode never touches this unless the user
+# opts in (see `state["claude_write_default_settings"]`): the `ucode claude`
+# launcher passes `--settings ucode-settings.json`, but the Claude desktop app and
+# the IDE / VS Code extension read settings.json directly and would otherwise not
+# see the Databricks gateway auth.
+CLAUDE_DEFAULT_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "settings.json"
 CLAUDE_DEFAULT_BACKUP_PATH = APP_DIR / "claude-settings.backup.json"
-
-# On Windows, subprocess can fail to locate 'claude' via PATH because npm
-# wrappers are .cmd files that Python's subprocess doesn't resolve the same
-# way as cmd.exe. Resolve the actual .exe inside the npm package tree so
-# the subprocess call is unambiguous on all platforms.
-_claude_wrapper = shutil.which("claude.cmd") or shutil.which("claude") or shutil.which("claude.bat")
-if _claude_wrapper:
-    _claude_root = Path(_claude_wrapper).parent
-    _claude_exe = (
-        _claude_root
-        / "node_modules"
-        / "@anthropic-ai"
-        / "claude-code"
-        / "bin"
-        / "claude.exe"
-    )
-    CLAUDE_BINARY = str(_claude_exe) if _claude_exe.exists() else _claude_wrapper
-else:
-    CLAUDE_BINARY = "claude"
+# State key: when truthy, write_tool_config also mirrors the overlay into the
+# default settings.json. Set by `ucode configure claude --write-default-settings`.
+WRITE_DEFAULT_SETTINGS_KEY = "claude_write_default_settings"
+# State key recording that ucode created settings.json itself (no pre-existing
+# file to back up), so revert deletes rather than leaves it behind.
+DEFAULT_SETTINGS_MANAGED_KEY = "claude_default_settings_managed"
 
 SPEC: ToolSpec = {
-    "binary": CLAUDE_BINARY,
+    "binary": "claude",
     "package": "@anthropic-ai/claude-code",
     "display": "Claude Code",
     "config_path": CLAUDE_SETTINGS_PATH,
     "backup_path": CLAUDE_BACKUP_PATH,
 }
+
+
+def claude_command(binary: str = "claude") -> list[str]:
+    """Return the argv prefix used to invoke the Claude CLI.
+
+    On Windows, npm installs `claude` as a `.cmd`/`.bat` wrapper. Python's
+    subprocess/execvp resolves these differently from cmd.exe and `CreateProcess`
+    cannot execute a batch file directly, so a bare wrapper path fails. When the
+    resolved executable is such a wrapper, run it through `cmd /c`. On POSIX (and
+    when a real executable is found) the binary is returned unchanged, so the
+    default spec and macOS/Linux behaviour stay identical."""
+    if os.name == "nt":
+        wrapper = shutil.which("claude.cmd") or shutil.which("claude.bat") or shutil.which(binary)
+        if wrapper and wrapper.lower().endswith((".cmd", ".bat")):
+            return ["cmd", "/c", wrapper]
+    return [binary]
 
 
 def is_update_available() -> tuple[str, str] | None:
@@ -236,7 +245,6 @@ def _unregister_web_search_mcp() -> None:
 
 def write_tool_config(state: dict, model: str) -> dict:
     backup_existing_file(CLAUDE_SETTINGS_PATH, CLAUDE_BACKUP_PATH)
-    backup_existing_file(CLAUDE_DEFAULT_SETTINGS_PATH, CLAUDE_DEFAULT_BACKUP_PATH)
     web_search_model = _resolve_web_search_model(state)
     overlay, managed_keys = render_overlay(
         state["workspace"],
@@ -273,13 +281,12 @@ def write_tool_config(state: dict, model: str) -> dict:
         _remove_tracing_stop_hook(merged)
     write_json_file(CLAUDE_SETTINGS_PATH, merged)
 
-    # Mirror the auth/env overlay into ~/.claude/settings.json so Claude Code
-    # finds the apiKeyHelper and ANTHROPIC_* env vars regardless of whether it
-    # is launched via `ucode claude` (which passes --settings) or directly from
-    # the Claude desktop app / IDE extension (which reads settings.json by default).
-    existing_default = read_json_safe(CLAUDE_DEFAULT_SETTINGS_PATH)
-    merged_default = deep_merge_dict(existing_default, overlay)
-    write_json_file(CLAUDE_DEFAULT_SETTINGS_PATH, merged_default)
+    # Opt-in only: mirror the same overlay into the default ~/.claude/settings.json
+    # so the Claude desktop app / IDE extension (which ignore --settings) also pick
+    # up the Databricks gateway auth. Disabled by default to keep ucode's footprint
+    # limited to its own ucode-settings.json.
+    if state.get(WRITE_DEFAULT_SETTINGS_KEY):
+        state = _write_default_settings(state, overlay)
 
     if web_search_model:
         _register_web_search_mcp(state["workspace"], web_search_model, state.get("profile"))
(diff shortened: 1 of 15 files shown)


rohita5l (changes requested):
Thanks for fixing the Windows path here. I think the intent is right, but it requires some changes

  The PR currently mixes three separate behavior changes:

  1. Windows auth command syntax.
  2. Windows Claude binary resolution.
  3. Writing ucode’s Claude overlay into the user’s default `~/.claude/settings.json`.

  The first two are Windows fixes. The third changes default behavior for all users and should either be opt-in or split into a separate PR. We definitely dont want as the default behavior and prefer to keep ucode settings separate.

  My preferred direction:

  - Keep `SPEC["binary"] = "claude"` stable.
  - Resolve a Windows-specific Claude executable only at launch/validation time.
  - Keep writing `~/.claude/ucode-settings.json` by default.
  - Make merging into `~/.claude/settings.json` explicit opt-in
  - Add focused tests for the Windows auth command and binary resolution behavior.

src/ucode/agents/claude.py
```diff
@@ -30,10 +30,31 @@
...
+        / "bin"
+        / "claude.exe"
+    )
+    CLAUDE_BINARY = str(_claude_exe) if _claude_exe.exists() else _claude_wrapper
+else:
+    CLAUDE_BINARY = "claude"
 
 SPEC: ToolSpec = {
-    "binary": "claude",
+    "binary": CLAUDE_BINARY,
```
rohita5l: I’m not convinced we should resolve `CLAUDE_BINARY` at import time or on all platforms.

  This makes `SPEC["binary"]` environment-dependent and changes macOS/Linux behavior too. In my local focused test run on this PR, `tests/test_agent_claude.py` failed because `SPEC["binary"]` became my absolute
  `~/.nvm/.../bin/claude` path instead of `"claude"`.

  Could we scope this to Windows launch/validation only, preferably behind a helper like `_resolve_claude_binary()`? That would keep the default spec stable and avoid doing PATH/package-layout probing during module import.

src/ucode/agents/claude.py
```diff
@@ -30,10 +30,31 @@
...
+_claude_wrapper = shutil.which("claude.cmd") or shutil.which("claude") or shutil.which("claude.bat")
+if _claude_wrapper:
+    _claude_root = Path(_claude_wrapper).parent
+    _claude_exe = (
+        _claude_root
+        / "node_modules"
+        / "@anthropic-ai"
+        / "claude-code"
+        / "bin"
+        / "claude.exe"
```
rohita5l: Can we verify this actually fixes the Windows launch failure?

  My concern is that npm usually installs `.cmd` / `.ps1` / shim wrappers, and I’m not sure `@anthropic-ai/claude-code` actually ships `bin/claude.exe`. If `_claude_exe.exists()` is false, this falls back to the
  `.cmd` wrapper path. Passing a `.cmd` path to `subprocess` with `shell=False` can still fail on Windows because `CreateProcess` does not execute batch files directly.

  It would be safer to either invoke the `.cmd` through `cmd /c` in the Windows path, or resolve this at the call site in a way that matches how `validate_tool` / `launch` actually execute the binary.

src/ucode/agents/claude.py
```diff
@@ -251,6 +273,14 @@ def write_tool_config(state: dict, model: str) -> dict:
         _remove_tracing_stop_hook(merged)
     write_json_file(CLAUDE_SETTINGS_PATH, merged)
 
+    # Mirror the auth/env overlay into ~/.claude/settings.json so Claude Code
+    # finds the apiKeyHelper and ANTHROPIC_* env vars regardless of whether it
+    # is launched via `ucode claude` (which passes --settings) or directly from
+    # the Claude desktop app / IDE extension (which reads settings.json by default).
+    existing_default = read_json_safe(CLAUDE_DEFAULT_SETTINGS_PATH)
```
rohita5l: Can we make writing to the default `~/.claude/settings.json` opt-in?

  `ucode claude` already launches with `--settings ~/.claude/ucode-settings.json`, so this PR broadens the behavior from “manage ucode’s Claude config” to “also mutate the user’s main Claude config.” That has a larger blast radius, especially because direct `claude` usage would start inheriting the Databricks gateway `apiKeyHelper` / `ANTHROPIC_*` settings.

  A simpler shape might be: keep writing `ucode-settings.json` by default, and add an explicit option for users who want this global behavior

rohita5l: If we do write to `~/.claude/settings.json`, we should also wire it into cleanup/rollback.

  Right now the managed config state and validation rollback are centered on `SPEC["config_path"]` / `SPEC["backup_path"]`, which still point at `ucode-settings.json`. This adds a second managed file, but I don’t see matching restore behavior for it.

src/ucode/databricks.py
```diff
@@ -956,7 +957,29 @@ def list_databricks_apps(workspace: str, profile: str | None = None) -> list[dic
 
 
 def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
-    workspace_arg = shlex.quote(workspace.rstrip("/"))
+    workspace_clean = workspace.rstrip("/")
```
rohita5l: Could we split the platform-specific command construction into small helpers instead of building both shell dialects inline here?

  The early return is not the main issue for me; the function now mixes POSIX shell, Windows `cmd.exe`, Python `-c`, profile handling, and `DATABRICKS_BEARER` fallback in one place, which makes the quoting hard to
  audit.

  For example:

  ```python
  def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
      workspace_clean = workspace.rstrip("/")
      if os.name == "nt":
          return _build_windows_auth_command(workspace_clean, profile)
      return _build_posix_auth_command(workspace_clean, profile)
```

  That keeps the public API simple while isolating the two command syntaxes. It would also make it easier to add focused tests for Windows quoting separately from the existing POSIX sh behavior.

File: src/ucode/databricks.py [hunk 5]
```diff
@@ -956,7 +957,29 @@ def list_databricks_apps(workspace: str, profile: str | None = None) -> list[dic
...
+        helper_code = (
+            "import json, os, subprocess, sys; "
+            "bearer=os.environ.get('DATABRICKS_BEARER'); "
+            "sys.exit(print(bearer) or 0) if bearer else None; "
+            f"cmd=[{databricks_exe!r},'auth','token','--host',{workspace_clean!r}"
+            f"{profile_part},'--force-refresh','--output','json']; "
+            "p=subprocess.run(cmd,capture_output=True,text=True,check=True,timeout=30); "
+            "print(json.loads(p.stdout)['access_token'])"
+        )
+        return f'"{python_exe}" -c {helper_code!r}'
```
rohita5l: This `python -c` command is hard to audit because it nests Python `repr` inside another `repr`, then hands the result to `cmd.exe`.

  Values containing quotes or `%` can behave differently under `cmd.exe`, and this is exactly the kind of code that tends to regress silently. Could we either move the Windows helper into a small dedicated helper
  function with explicit quoting/tests, or avoid the `-c` blob by invoking a stable ucode subcommand/helper instead?

rohita5l (changes requested):

rohita5l:
1.  Instead of trying to find claude.exe, just treat Windows wrappers as wrappers.

  Something like:
```
  def claude_command(binary: str = "claude") -> list[str]:
      if os.name == "nt":
          wrapper = shutil.which("claude.cmd") or shutil.which("claude.bat") or shutil.which(binary)
          if wrapper and wrapper.lower().endswith((".cmd", ".bat")):
              return ["cmd", "/c", wrapper]
      return [binary]
```
  Then in launch/validation:
```
  cmd = claude_command(SPEC["binary"])
  os.execvp(cmd[0], [*cmd, "--settings", str(CLAUDE_SETTINGS_PATH), *tool_args])
```
  and validation:
```
  return [
      *claude_command(binary),
      "--settings",
      str(CLAUDE_SETTINGS_PATH),
      "-p",
      "say hi in 5 words or less",
      "--max-turns",
      "1",
  ]
  ```
  
  2. For changing the global ~/.claude/settings.json, there should be a separate command or a flag in ucode configure claude
  
 3. Finally, keep the existing POSIX implementation unchanged, and add one Windows helper:

```
  def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
      workspace_clean = workspace.rstrip("/")
      if os.name == "nt":
          return _build_windows_auth_command(workspace_clean, profile)
      return _build_posix_auth_command(workspace_clean, profile)
```
 For Windows, I’d avoid a complicated inline python -c if possible.

rohita5l:
@yakup-ozturk I merged a change to make it work in windows. can you please give it a try ?

yakup-ozturk:
> @yakup-ozturk I merged a change to make it work in windows. can you please give it a try ?

C:\...\src\ucode\cli.py:1101 in configure

│   1098 │   │   │   # Tool binaries are installed after the user picks which agents
│   1099 │   │   │   # they want, in configure_workspace_command.
│   1100 │   │   │   if workspace_entries is None:
│ > 1101 │   │   │   │   configure_workspace_command(
│   1102 │   │   │   │   │   prompt_optional_updates=prompt_optional_updates,
│   1103 │   │   │   │   │   **skip_kwargs,
│   1104 │   │   │   │   )

│ C:\...\src\ucode\cli.py:562 in configure_workspace_command

│    559 │   │   return 0
│    560 │
│    561 │   for tool_name in picked:
│ >  562 │   │   install_tool_binary(
│    563 │   │   │   tool_name,
│    564 │   │   │   strict=False,
│    565 │   │   │   update_existing=True,

│ C:\...\src\ucode\agents\__init__.py:199 in install_tool_binary

│   196 │   print_section("Bootstrap")
│   197 │   print_warning(f"`{binary}` was not found. Installing {spec['display']}...")
│   198 │   try:
│ > 199 │   │   subprocess.run(["npm", "install", "-g", package], check=True, timeout=300)
│   200 │   except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as exc:
│   201 │   │   message = f"Failed to install {spec['display']} automatically."
│   202 │   │   if strict:

│ C:\...\Python312\Lib\subprocess.py:548 in run

│    545 │   │   kwargs['stdout'] = PIPE
│    546 │   │   kwargs['stderr'] = PIPE
│    547 │
│ >  548 │   with Popen(*popenargs, **kwargs) as process:
│    549 │   │   try:
│    550 │   │   │   stdout, stderr = process.communicate(input, timeout=timeout)
│    551 │   │   except TimeoutExpired as exc:

│ C:\...\Python312\Lib\subprocess.py:1026 in __init__

│   1023 │   │   │   │   │   self.stderr = io.TextIOWrapper(self.stderr,
│   1024 │   │   │   │   │   │   encoding=encoding, errors=errors)
│   1025 │   │   │
│ > 1026 │   │   │   self._execute_child(args, executable, preexec_fn, close_fds,
│   1027 │   │   │   │   │   │   pass_fds, cwd, env,
│   1028 │   │   │   │   │   │   startupinfo, creationflags, shell,
│   1029 │   │   │   │   │   │   p2cread, p2cwrite,

│ C:\...\Python312\Lib\subprocess.py:1538 in _execute_child

│   1535 │   │   │
│   1536 │   │   │   # Start the process
│   1537 │   │   │   try:
│ > 1538 │   │   │   │   hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
│   1539 │   │   │   │   │               None, None,
│   1540 │   │   │   │   │               int(not close_fds),
│   1541 │   │   │   │

FileNotFoundError: [WinError 2] The system cannot find the specified file.

PR #151, opened 2026-06-10
Add UCODE_USE_UC_SECURABLES opt-in for UC model-services and MCP-services discovery

AarushiShah-db:
## Summary

When `UCODE_USE_UC_SECURABLES` is set, ucode discovers two kinds of UC
securables that are otherwise invisible:

1. **Models** via the Unity Catalog model-services API, addressed as
   `system.ai.<model-name>` instead of the per-family AI Gateway listings.
   Base URLs are unchanged — only the model name differs.

2. **Curated MCP servers** via the Unity Catalog mcp-services API
   (`system.ai.github`, `system.ai.slack`, `system.ai.atlassian`, …),
   surfaced in `ucode configure mcp` alongside the existing UC connection-
   based MCP entries. Selecting one registers an HTTP MCP server pointing at
   `/<workspace>/ai-gateway/mcp-services/<full-name>`.

The flag is sticky — persisted into state at configure time so launches
honor it without re-exporting the env var (an explicit env var still wins).
With it unset, ucode behavior is unchanged: legacy AI Gateway model lists
and the existing MCP picker.

Discovery uses a small page size with per-page retries because the metastore listing is slow and frequently 504s, and filters to system.ai.* so model services from other schemas don't leak into the family buckets.

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run pytest tests/test_databricks.py`
- [x] `UCODE_USE_UC_SECURABLES=1 uv run ucode configure` on e2-dogfood
      pulls Claude/Codex models as `system.ai.*`.
- [x] `UCODE_USE_UC_SECURABLES=1 uv run ucode configure mcp` surfaces
      `system.ai.github`, etc.; selection registers correctly and
      `tools/list` works after one-time managed-OAuth authorize.
- [x] With the flag unset, model listings and MCP picker are unchanged.
- [x] `uv run pytest` (full suite)

**MCP services surfaced:**

Screenshot 2026-06-11 at 4 50 15 PM

**System.ai models surfaced:**

In Claude:
Screenshot 2026-06-11 at 4 53 15 PM
Screenshot 2026-06-11 at 4 51 54 PM

In Pi (all 3 model families supported shown):
Screenshot 2026-06-11 at 4 53 37 PM

Co-authored-by: Isaac

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -977,19 +977,22 @@ def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
     )
 
 
-def use_uc_securables(default: bool = False) -> bool:
+def uc_enabled(default: bool = False) -> bool:
     """True when the opt-in UC-securables discovery path is enabled.
 
-    Set ``UCODE_USE_UC_SECURABLES=1`` (or true/yes/on) to:
-      - discover models via UC `model-services` (`system.ai.<model-name>`)
-      - surface curated `system.ai.*` MCP services in `ucode configure mcp`
+    Three input precedences, callers handle the highest one first:
+      1. ``ucode configure --enable-uc / --no-enable-uc`` (resolved by the
+         CLI before this function is called and passed in via ``default``,
+         since it overrides everything).
+      2. ``UCODE_ENABLE_UC=1`` (or true/yes/on) env var.
+      3. The value persisted in state (sticky, also passed via ``default``).
 
-    The env var, when set to any value, wins. ``default`` is the fallback used
-    when the env var is unset — callers pass the value persisted in state so a
-    workspace configured with the flag stays on the UC path without the env
-    var being re-exported each time.
+    Enabling UC discovery makes ucode:
+      - resolve models via UC `model-services` as `system.ai.<model-name>`
+        instead of the per-family AI Gateway listings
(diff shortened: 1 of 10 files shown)


rohita5l (commented):
@asujithan can you add some e2e tests

src/ucode/databricks.py
```diff
@@ -977,6 +977,225 @@ def build_auth_shell_command(workspace: str, profile: str | None = None) -> str:
...
+    Set ``UCODE_USE_UC_SECURABLES=1`` (or true/yes/on) to:
+      - discover models via UC `model-services` (`system.ai.<model-name>`)
+      - surface curated `system.ai.*` MCP services in `ucode configure mcp`
+
+    The env var, when set to any value, wins. ``default`` is the fallback used
+    when the env var is unset — callers pass the value persisted in state so a
+    workspace configured with the flag stays on the UC path without the env
+    var being re-exported each time.
+    """
+    raw = os.environ.get("UCODE_USE_UC_SECURABLES")
```
rohita5l: just make it UCODE_ENABLE_UC

also it should something people can pass in with configure command like --enable-uc

rohita5l (approved):
@AarushiShah-db can you also PTAL

PR #155, opened 2026-06-12
fix gemini mcp registration to write to correct settings.json

asujithan:
## Summary

Before:
`ucode configure mcp` would silently register Gemini MCP servers into a file the launcher never reads, so `/mcp` inside `ucode gemini` always showed "No MCP servers configured." even after a successful save.

`add_gemini_mcp_server` / `remove_gemini_mcp_server` ran `gemini mcp add/remove` without setting `GEMINI_CLI_HOME`, so entries went to `$HOME/.gemini/settings.json`. But `agents/gemini.py` launches Gemini with `GEMINI_CLI_HOME=~/.ucode/.gemini-home`, which makes the session read from a different settings.json. The two never intersected.

Fix: pin `GEMINI_CLI_HOME` to the same `gemini.GEMINI_HOME_DIR` for both subprocess calls so registrations write to the file the launcher reads.

## Test plan

- [x] `uv run ruff check src/ucode/mcp.py tests/test_mcp.py`
- [x] `uv run pytest tests/test_mcp.py` (39/39)
- [x] `ucode configure mcp` → select an entry → `ucode gemini` → `/mcp` lists the registered servers and `tools/list` works.
- [x] `~/.ucode/.gemini-home/.gemini/settings.json` contains the new `mcpServers` entries; `~/.gemini/settings.json` is no longer written by ucode.

Screenshot 2026-06-11 at 5 20 51 PM

### DIFF
--- a/src/ucode/mcp.py
+++ b/src/ucode/mcp.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import os
 import shutil
 import string
 import subprocess
@@ -23,7 +24,7 @@ from questionary.prompts.common import InquirerControl
 from questionary.question import Question
 from questionary.styles import merge_styles_default
 
-from ucode.agents import copilot, opencode
+from ucode.agents import copilot, gemini, opencode
 from ucode.config_io import restore_file
 from ucode.databricks import (
     ensure_databricks_auth,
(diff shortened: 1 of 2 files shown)


AarushiShah-db (approved):
thank you!!

PR #156, opened 2026-06-12
Support non-interactive configure with `--profiles`

bbqiu:
ucode configure forced an interactive databricks auth login even when a static credential was available, and PAT profiles in ~/.databrickscfg were invisible to ucode (databricks auth token is OAuth-only). Now:

- resolve_pat_token picks up DATABRICKS_BEARER, a DATABRICKS_TOKEN with matching DATABRICKS_HOST, or a PAT-type CLI profile (token read from ~/.databrickscfg), and configure skips the forced login when one exists; workspace access is still verified via the AI Gateway probe.
- find_profile_name_for_host resolves PAT profiles (OAuth wins per host).
- New `ucode configure --profiles <names>` uses already-authenticated CLI profiles with no interactive login; mutually exclusive with --workspaces.
- The persisted agent auth shell command falls back to `databricks auth describe --sensitive` for PAT profiles so launched agents can refresh tokens.

tested manually:
--profiles and --workspaces still work

screenshot 2026-06-11-23-51-04-iTerm2

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,20 @@ ucode configure --workspaces https://first.databricks.com,https://second.databri
 
 When multiple workspaces are provided, `ucode` logs into and saves state for each workspace. Launch commands such as `ucode codex` use the first workspace in the list.
 
+Alternatively, pass existing Databricks CLI profiles (from `~/.databrickscfg`) instead of workspace URLs — each profile's host supplies the workspace URL:
+
+```bash
+ucode configure --profiles DEFAULT --agents claude,codex
+```
+
+Auth behaves the same as `--workspaces`: an OAuth `databricks auth login` is forced by default.
+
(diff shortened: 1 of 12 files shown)


rohita5l:
@bbqiu Im okay with adding the `configure --profiles` as an analogous to `configure --workspaces` however I dont want ucode to automatically start using PAT token if its available. We generally want to heavily discourage use of that.

So to force ucode to use a PAT token can the contract be the following.
1. `DATABRICKS_BEARER` should be available as env var. Thats the only key we respect
2. Users need to pass in --skip-oauth

That can work in a Lakebox / CI like environment imo but for interactive use cases the default should force you to oauth.

rohita5l (dismissed):

rohita5l (approved):

PR #165, opened 2026-06-16
Dedup Claude /model list

asujithan:
## Problem
ucode claude → /model showed duplicate "Opus 4.8" rows (one from ANTHROPIC_MODEL, one from ANTHROPIC_DEFAULT_OPUS_MODEL) that both resolved to the same gateway-routable id. Visually noisy and confusing — users couldn't tell which row to pick.

## Fix
- Stop setting ANTHROPIC_MODEL in render_overlay. The "Default (recommended)" row now resolves through the pinned ANTHROPIC_DEFAULT_OPUS_MODEL family alias instead of standing as its own duplicate row.
- Actively prune stale ucode-managed model env keys in write_tool_config. deep_merge_dict doesn't remove keys absent from the overlay, so without this any ANTHROPIC_MODEL (or short-lived _NAME companions) written by older ucode versions would linger in ~/.claude/settings.json forever and the dupe would never go away. Pruning is scoped to a fixed allowlist (CLAUDE_MANAGED_MODEL_ENV_KEYS) so user-owned env vars are untouched.

## Test plan

[x] uv run pytest tests/test_agent_claude.py (53 passed)
[x] uv run pytest (762 passed; 2 pre-existing env-only failures in pi/opencode user-agent e2e tests unrelated to this change)
[x] uv run ruff check .
[x] Manual: ucode claude shows 4 rows in /model; "Default (recommended)" launches in opus; /model sonnet switches families; ucode claude -- --model haiku starts in haiku

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -76,6 +76,18 @@ CLAUDE_TRACING_ENV_KEYS = (
     "MLFLOW_EXPERIMENT_ID",
     "MLFLOW_TRACING_SQL_WAREHOUSE_ID",
 )
+# Model-selection env keys ucode owns end-to-end. Anything in this tuple that
+# isn't written by render_overlay gets actively pruned from settings.json on
+# every launch, so stale values from older ucode versions never linger.
+CLAUDE_MANAGED_MODEL_ENV_KEYS = (
+    "ANTHROPIC_MODEL",
+    "ANTHROPIC_DEFAULT_OPUS_MODEL",
+    "ANTHROPIC_DEFAULT_OPUS_MODEL_NAME",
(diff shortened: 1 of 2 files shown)


src/ucode/agents/claude.py
```diff
@@ -135,13 +147,23 @@ def render_overlay(
         ]
     )
     env: dict[str, str] = {
-        "ANTHROPIC_MODEL": _maybe_add_1m_suffix(model),
```
AarushiShah-db, line 138: what is the default model now that we remove this?

AarushiShah-db (approved):

src/ucode/agents/claude.py
asujithan: default is opus-4-8 model still

PR #168, opened 2026-06-18
Fix SSL cert verification failures behind enterprise inspection proxies

edurdevic:
## Problem

`ucode init` fails with `CERTIFICATE_VERIFY_FAILED: self-signed certificate in certificate chain` on networks with a corporate TLS inspection proxy, even though `curl` to the same AI Gateway endpoint succeeds.

Root cause: Python's `urllib.request.urlopen` does not load the OS trust store on all platforms. The proxy injects a self-signed CA that `curl` finds via the system keychain, but Python's default SSL context misses it. The error surfaces as:

```
Databricks Unity AI Gateway probe failed on this workspace
(network error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
self-signed certificate in certificate chain (ssl.c:1001)).
```

## Fix

Add `_make_ssl_context()` in `databricks.py` that creates a default SSL context and additionally loads any CA bundle pointed to by `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, or `SSL_CERT_FILE` — the same env vars honored by `curl` and the `requests` library. All three `urlopen` call sites (`_http_get_json`, `_http_post_json`, `discover_sql_warehouse_http_path`) now use this context.

## Workaround for affected customers (before this ships)

```sh
export REQUESTS_CA_BUNDLE=/path/to/enterprise-ca.pem
ucode init
```

## Test plan

- [x] All existing unit tests pass (`uv run pytest`)
- [x] Lint clean (`uv run ruff check .` + `uv run ruff format --check`)
- [ ] Manual test on a workspace behind an SSL inspection proxy


src/ucode/databricks.py
```diff
@@ -194,6 +195,27 @@ def _log_auth_diagnostics() -> None:
...
+    at their enterprise CA bundle without patching the system Python install."""
+    ctx = ssl.create_default_context()
+    for env_var in ("REQUESTS_CA_BUNDLE", "CURL_CA_BUNDLE", "SSL_CERT_FILE"):
+        ca_bundle = os.environ.get(env_var, "").strip()
+        if ca_bundle and Path(ca_bundle).is_file():
+            try:
+                ctx.load_verify_locations(cafile=ca_bundle)
+            except ssl.SSLError:
+                pass
+            break
```
rohita5l: Bug — break fires even when load_verify_locations fails (_make_ssl_context, line ~214)

              try:
                  ctx.load_verify_locations(cafile=ca_bundle)
              except ssl.SSLError:
                  pass
              break   # ← outside try/except, always executes

src/ucode/databricks.py
```diff
@@ -194,6 +195,27 @@ def _log_auth_diagnostics() -> None:
         _debug(f"databrickscfg ({cfg_path})", f"read error: {exc}")
 
 
+def _make_ssl_context() -> ssl.SSLContext:
```
rohita5l, line 234: Minor — _make_ssl_context() called on every request, should be cached

  It hits the env vars + filesystem on every urlopen call. The rest of the file already uses @functools.cache for this pattern. Adding `@functools.cache` makes it a one-shot per process.

src/ucode/databricks.py
```diff
@@ -194,6 +195,27 @@ def _log_auth_diagnostics() -> None:
...
+    env vars that curl and the `requests` library use lets customers point ucode
+    at their enterprise CA bundle without patching the system Python install."""
+    ctx = ssl.create_default_context()
+    for env_var in ("REQUESTS_CA_BUNDLE", "CURL_CA_BUNDLE", "SSL_CERT_FILE"):
+        ca_bundle = os.environ.get(env_var, "").strip()
+        if ca_bundle and Path(ca_bundle).is_file():
+            try:
+                ctx.load_verify_locations(cafile=ca_bundle)
+            except ssl.SSLError:
+                pass
```
rohita5l: OSError (e.g. permission denied) is not caught here — Path.is_file() returns True for files the current user cannot read, but `load_verify_locations` will then raise OSError: Permission denied, which propagates uncaught and crashes the process. 

Suggest broadening the except catch

rohita5l (changes requested):

src/ucode/databricks.py
edurdevic: Updated with 
            else:
                break

src/ucode/databricks.py
edurdevic: added @functools.cache

src/ucode/databricks.py
edurdevic: broadened to except OSError

PR #189, opened 2026-07-07
claude: merge a caller's --settings instead of clobbering ucode's

Edwinhe03:
## What

`ucode claude` prepends its own `--settings <ucode-file>` to every launch. Claude Code honors only **one** `--settings` flag, so a caller that *also* passes `--settings '{…}'` (e.g. an integration injecting hooks) has exactly one of the two silently dropped — either ucode's gateway config or the caller's settings. This broke omnigent's managed Claude sessions: its transcript/permission hooks were dropped and the web UI rendered empty.

## Change

When the caller supplies `--settings`, compose instead of clobber: deep-merge the caller's settings with ucode's — ucode's gateway keys (`apiKeyHelper`, `env`) win, and `hooks` from both sides are **unioned** — then hand Claude a single merged `--settings` (inline JSON). The merge is per-launch and is **never written back** to the shared `ucode-settings.json`, so concurrent launches can't accumulate one another's hooks. No caller `--settings` → unchanged (ucode's settings file is passed directly).

This is the general "ucode should compose with any prior `claude` command out of the box" fix — it helps every caller of `ucode claude --settings`, not just omnigent.

## Test plan

- `uv run pytest` — green.
- New `TestBuildClaudeArgv` covers: merge into a single `--settings`, the `--settings=` form, ucode-wins-on-conflict, file-path value, and unparseable pass-through.
- `uv run ruff check .` + `ruff format --check` clean.

## Related

Split out of #184 (which now carries only the codex `--profile`-on-`app-server` fix). Same underlying principle: ucode's injected flags must compose with the caller's command — merged where a flag would clobber (`--settings`, here), conditional where a subcommand rejects it (`--profile`, in #184).

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -522,23 +522,42 @@ def _extract_caller_settings(tool_args: list[str]) -> tuple[list[str], list[str]
     return values, remaining
 
 
-def _load_caller_settings(value: str) -> dict | None:
+def _load_caller_settings(value: str) -> dict:
     """Resolve a ``--settings`` value (inline JSON or file path) to a dict.
 
-    Returns ``None`` when it is neither parseable JSON nor an existing file, so
-    the caller can pass such a value through untouched instead of dropping it.
+    Claude Code accepts either inline JSON or a path to a JSON file. Raises
+    ``RuntimeError`` (surfaced by the CLI as an actionable error) when the value
+    is neither, rather than silently dropping it: a dropped value would also be
+    passed through as a second ``--settings`` flag, and Claude Code honors only
+    one — so either the caller's settings or ucode's gateway config would be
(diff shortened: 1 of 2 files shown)


src/ucode/agents/claude.py
```diff
@@ -495,12 +497,130 @@ def default_model(state: dict) -> str | None:
...
+
+    Returns ``None`` when it is neither parseable JSON nor an existing file, so
+    the caller can pass such a value through untouched instead of dropping it.
+    """
+    text = value.strip()
+    if text.startswith("{"):
+        try:
+            parsed = json.loads(text)
+        except json.JSONDecodeError:
+            return None
```
lilly-luo: should this throw more loudly

lilly-luo (dismissed):

lilly-luo (approved):

PR #191, opened 2026-07-08
Force token refresh in the auth-token apiKeyHelper

brodrigu:
## Problem

Long-running Claude Code / Codex sessions die when the original Databricks OAuth token expires (~1h), even though the `apiKeyHelper` is configured to refresh every 15 min via `CLAUDE_CODE_API_KEY_HELPER_TTL_MS: 900000`.

## Root cause

`ucode auth-token` is the command Claude Code's `apiKeyHelper` (and Codex's auth command) re-invokes on that 15-min interval. It called `get_databricks_token(...)` **without** `force_refresh`, so the underlying `databricks auth token` only renewed the token when the CLI *independently* judged the cached token "close to expiry":

> Get authentication token from the local cache… Refresh the access token if it is expired or close to expiry. Use `--force-refresh` to bypass expiry checks.

That heuristic can lapse between helper invocations, so the session dies at the original token's expiry. Notably, this is inconsistent with the gemini/opencode/copilot/pi agents, whose background refreshers already pass `force_refresh=True`.

## Fix

Pass `force_refresh=True` from the `auth-token` helper so each refresh cycle bypasses the expiry heuristic and hands back a token with a full fresh lifetime.

## Tests

- `tests/test_cli.py` — the `auth-token` command forwards `force_refresh=True`.
- `tests/test_databricks.py` — `--force-refresh` is (and by default isn't) forwarded to the real `databricks auth token` argv.

Full suite passes (`853 passed`); the only failures are two pre-existing network-dependent e2e tests that also fail on clean `main`.

## Note for reviewers

PR #137 (Windows support) touches this same auth code and is stale against #177 — the two will conflict.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -776,7 +776,15 @@ def auth_token_cmd(
             )
             raise typer.Exit(1)
     try:
-        token = get_databricks_token(workspace, profile)
+        # force_refresh: this helper is re-invoked by Claude Code's apiKeyHelper
+        # (and Codex's auth command) on the CLAUDE_CODE_API_KEY_HELPER_TTL_MS
+        # interval (15 min). Forcing a refresh each cycle hands back a token with
+        # a full fresh lifetime, so a long session never dies at the original
+        # token's ~1h expiry. Without it, `databricks auth token` only refreshes
+        # when it independently judges the cached token "close to expiry", which
+        # can lapse between invocations. Mirrors the force_refresh=True used by
+        # the gemini/opencode/copilot/pi background refreshers.
+        token = get_databricks_token(workspace, profile, force_refresh=True)
     except RuntimeError as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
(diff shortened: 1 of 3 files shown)


brodrigu:
Thanks — good catch. Addressed in 08bd1d4: the three `--use-pat` test mocks now declare `force_refresh` as keyword-only (`lambda w, p, *, force_refresh=False`), mirroring the real `get_databricks_token` signature so the tests fail if the call site ever passes it positionally.

brodrigu:
Pushed 54b4432 on top of the accepted quoting suggestion:

- **Path quoting** — extended the `>> "$argv_log"` quoting to the third (pre-existing) occurrence in `test_passes_profile_flag_when_provided` so the fake shim is consistent.
- **`DATABRICKS_BEARER` short-circuit** — the remaining Copilot concern. Rather than patching each test, I strip `DATABRICKS_BEARER` once in the shared `_fake_databricks` helper, since `get_databricks_token` short-circuits on it and *every* test in `TestGetDatabricksToken` needs the fake CLI shim to actually run. A leaked CI/developer bearer would otherwise silently bypass the shim.

All 216 tests in `test_databricks.py` + `test_cli.py` pass; lint/format clean.

PR #195, opened 2026-07-08
ucode: add --skip-preflight to skip per-launch auth/gateway re-validation

Edwinhe03:
## What

Every `ucode <agent>` launch re-runs `configure_shared_state`, which does an auth login check, a token fetch, an AI Gateway v2 probe, and (outside provider mode) model discovery — **~5–10s of network round-trips** — even when a prior `ucode configure` already validated all of it. Managed/headless launchers (e.g. omnigent) that run `configure` once and then launch repeatedly pay this on every turn.

## Change

Add a launch-only `--skip-preflight` flag (distinct from the configure-only `--skip-validate`, which skips the post-configure model smoke test). It threads through `_launch_tool` as `configure_shared_state(skip_preflight=True)`:

- `configure_shared_state` assembles the shared workspace state (profile, base URLs, auth mode) up front, then **returns early** when `skip_preflight` is set; otherwise it falls through into the preflight block (auth + AI Gateway validation, then model discovery). One function, one shared assembly — no parallel state builder and no inverted `not skip_preflight` guard.
- The profile is resolved from `~/.databrickscfg` **at most once per launch** — in the skip branch (trusting the prior configure) or after `databricks auth login` on the full path, never both.
- The PAT/bearer is already exported by `apply_pat_environment`, and the AI Gateway was verified by the earlier `configure`, so the skipped work is redundant. Previously-discovered model lists are preserved (not clobbered with empties).

Wired into every launch command (`codex` / `claude` / `gemini` / `opencode` / `copilot` / `pi`). Default off — behavior is unchanged unless a caller opts in.

## `--help`

```
$ ucode codex --help

 Usage: ucode codex [OPTIONS]

 Launch Codex via Databricks.

╭─ Options ─────────────────────────────────────────────────────────────────────╮
│ --provider        TEXT  Route through a Unity Catalog Model Provider Service    │
│                         (<catalog>.<schema>.<name>). Skips Databricks model     │
│                         pinning; pass before any `--` separator.                │
│ --skip-preflight        Skip the per-launch Databricks auth + AI Gateway        │
│                         re-validation, trusting a prior `ucode configure`.       │
│ --help                  Show this message and exit.                             │
╰────────────────────────────────────────────────────────────────────────────────╯
```

## Tests

New tests in `tests/test_cli.py`:

- `TestConfigureSharedStateSkipPreflight` — `configure_shared_state(..., skip_preflight=True)` makes **no** `ensure_databricks_auth` / `ensure_ai_gateway_v2` / `get_databricks_token` / model-discovery calls (each patched to raise if hit), still rebuilds `base_urls` + persists state, and resolves the profile when `None`.
- `TestSkipPreflightFlag` — `--skip-preflight` on every launch command threads through `_launch_tool` to `configure_shared_state` as `skip_preflight=True` (and defaults to `False` without the flag).

```
$ uv run pytest tests/test_cli.py -k preflight
10 passed, 80 deselected

$ uv run pytest         # full unit suite
858 passed, 8 skipped
```

`ruff` + `ty` clean.

## Manual verification

Configured against a real staging workspace (`ucode configure --profiles <p> --agents claude --use-pat`), then launched the same agent with and without the flag (`claude --version` execs the real binary and exits, so the launch is observable end-to-end):

```
$ ucode claude --version                    # DEFAULT — preflight runs
✔ Databricks auth already available for https://…staging.cloud.databricks.com
✔ Unity AI Gateway detected
  Model: system.ai.claude-opus-4-5
✔ Starting Claude Code
2.1.206 (Claude Code)
elapsed: 7.39s

$ ucode claude --skip-preflight --version   # preflight skipped
  Model: system.ai.claude-opus-4-5
✔ Starting Claude Code
2.1.206 (Claude Code)
elapsed: 0.79s
```

`--skip-preflight` drops the auth + AI Gateway round-trips (**7.39s → 0.79s, ~7s saved**), and the agent still launches on the correctly-resolved model — confirming the trusted state (base URLs rebuilt, model lists preserved) is intact.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -216,6 +216,7 @@ def configure_shared_state(
     force_login: bool = False,
     use_pat: bool | None = None,
     skip_model_discovery: bool = False,
+    skip_preflight: bool = False,
 ) -> dict:
     """Log into Databricks, enforce AI Gateway v2, fetch model lists, persist state.
 
@@ -229,6 +230,12 @@ def configure_shared_state(
     ``--profile`` to every CLI invocation so ambiguous `~/.databrickscfg`
     entries (e.g. DEFAULT and a named profile both pointing at the same host)
     don't error out. If ``None``, we resolve it from the host after login.
+    If skip_preflight is True, skip the entire preflight block below — auth
+    validation, the AI Gateway probe, and model discovery — trusting a prior
+    ``ucode configure``. The PAT/bearer is already exported (``apply_pat_environment``
+    in ``_launch_tool``) and the gateway was verified by that earlier configure.
+    Only the local profile resolution and the shared state assembly still run;
+    the saved model lists are preserved.
     """
     workspace = normalize_workspace_url(workspace)
     prior_state = load_state()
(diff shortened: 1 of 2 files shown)


src/ucode/cli.py
```diff
@@ -284,39 +301,45 @@ def configure_shared_state(
...
-        # `system.ai.<model-name>` ids, bucketed by name. If a family comes back
-        # empty (workspace without UC model-services, or the listing failed), fall
-        # back to the per-family AI Gateway listing for that family only.
-        with spinner("Fetching available models..."):
-            ms_claude, ms_codex, ms_gemini, ms_oss, ms_reason = discover_model_services(
-                workspace, token
-            )
+    # Under no_preflight no discovery runs at all — the prior configure's
+    # model lists are trusted (and preserved below, not overwritten).
+    if not no_preflight:
```
lilly-luo: nit can you do something like

```
if no_preflight: 
    return 
else: 
    xxx 
``` 

not no_preflight is confusing

src/ucode/cli.py
```diff
@@ -216,6 +216,7 @@ def configure_shared_state(
     force_login: bool = False,
     use_pat: bool | None = None,
     skip_model_discovery: bool = False,
+    no_preflight: bool = False,
```
lilly-luo: nit no_preflight -> skip_preflight to be consistent? 

you don't need to address in this PR; however i think we need to clean up the stpes and create some utils to make it clearer what the diff steps are, what's happening at configure time vs agent-launch time, and make it clearer exactly what's being skipped

lilly-luo:
this is great! can you put the --help outputs on the PR description and have some examples to confirm this is working? before merge

lilly-luo (dismissed):
ideally also rename to --skip-preflight

lilly-luo (approved):

PR #197, opened 2026-07-09
fix: serialize token refresh across processes and fix OpenCode's dynamic Authorization plugin (#190)

### LINKED ISSUE #190: Concurrent ucode opencode/gemini sessions race on the shared OAuth cache during --force-refresh, revoking the token family → Forbidden: Invalid Token
## Summary

Running multiple concurrent `ucode opencode` (or `ucode gemini`) sessions against the same workspace/profile intermittently produces `Forbidden: Invalid Token` on all sessions, immediately after a background token refresh that appeared to succeed. The freshly written token is valid in isolation (a direct `curl` to the AI Gateway with it returns 200), but requests still 403 shortly after.

Root cause is a cross-process race on the shared Databricks OAuth cache during `--force-refresh`, with no locking anywhere in ucode.

This is distinct from #98 (which was about `x-api-key` vs `Authorization` header routing). Here the header routing is correct; the token itself gets revoked by a concurrent refresh.

## Environment

- ucode installed via `uv tool`
- OpenCode `1.17.10` (stock), macOS (arm64)
- `~/.databrickscfg` uses `auth_type = databricks-cli` with `auth_storage = secure` (macOS Keychain, single `databricks-cli` item)
- Multiple `ucode opencode` sessions running simultaneously (normal daily workflow — 4–5 terminals)

## Mechanism

`src/ucode/agents/opencode.py` launches a per-process daemon thread `_refresh_forever`, which every `TOKEN_REFRESH_INTERVAL_SECONDS` (1800s) calls `_refresh_token_once(..., force_refresh=True)` → `get_databricks_token(..., force_refresh=True)` → `databricks auth token --force-refresh` (`src/ucode/databricks.py:811-812`). `agents/gemini.py` has the identical pattern.

Every concurrent `ucode` process runs its own independent timer against the same OAuth cache (the Keychain item). `grep -rn "fcntl|flock|Lock|filelock" src/ucode/` returns nothing — there is no cross-process (or in-process) serialization around the refresh.

Databricks OAuth rotates the refresh token on redemption. When two sessions' 30-minute timers land close together, both redeem the same refresh token concurrently; the second presents an already-rotated (consumed) refresh token, which trips the OAuth server's refresh-token-reuse detection and revokes the whole token family — invalidating the access token that was just issued and written into `opencode.json`. Hence "Forbidden" right after a "successful" refresh.

## Reproduction

1. Configure ucode for opencode against a v2 AI Gateway workspace (Keychain-backed `databricks-cli` profile).
2. Launch 4–5 concurrent `ucode opencode` sessions and leave them running past the 30-minute refresh boundary so multiple refresh timers overlap.
3. Observe intermittent `Forbidden: Invalid Token` across sessions after a refresh tick. Recovery requires `databricks auth logout && databricks auth login` to reset the token family.

(The race is timing-dependent; overlapping timers across ≥2 long-lived sessions make it likely within a few refresh cycles.)

## Impact

Multiple concurrent sessions is a common workflow. Today the only recovery is a manual re-login, and it recurs.

## Proposed fix (two independent axes)

**Axis 1 — serialize refreshes (fixes the revocation).** Wrap the `--force-refresh` path in `get_databricks_token` (`databricks.py`) with a cross-process `fcntl.flock` on a per-workspace+profile lockfile (e.g. under the ucode app dir). The winner refreshes; others block briefly, then read the now-fresh cache and skip their own force-refresh if it was refreshed within the last few seconds. Benefits every agent (opencode + gemini share the pattern).

**Axis 2 — stop baking a static, expiring token into a shared file for opencode.** OpenCode has no `apiKeyHelper` and resolves `{env:}`/`{file:}` only at startup, so the 30-min file rewrite never reaches a running session anyway. A ucode-managed OpenCode plugin using the per-request `chat.headers` hook can set `Authorization` from a freshly sourced token on each request (verified empirically against stock 1.17.10: a plugin-set `Authorization` overrides the static `provider.options.headers.Authorization` on the wire). Combined with Axis 1's lock, the token source stays serialized. This removes the shared-file-rewrite mechanism for opencode entirely.

Happy to contribute a PR for both axes (with tests + ruff) if the maintainers agree with the direction. Wanted to confirm awareness and preferred approach first.

jesco39:
## Summary

Fixes #190. Running multiple concurrent `ucode opencode` sessions against the same workspace/profile intermittently produces `Forbidden: Invalid Token` on all sessions, and a single long-lived session eventually hits the same error after its token outlives the process's in-memory startup config. Two independent root causes, fixed separately:

**Axis A — concurrent `--force-refresh` races on the shared Databricks OAuth cache.** Every `ucode opencode`/`ucode gemini` process runs its own background timer that calls `databricks auth token --force-refresh` every 30 minutes, with no locking anywhere. When two sessions' timers land close together, both redeem the same rotating refresh token concurrently; the loser presents an already-rotated refresh token, tripping the OAuth server's reuse detection and revoking the whole token family — invalidating the access token that was just issued.

**Axis B — OpenCode reads a static token once at startup and never again.** OpenCode has no `apiKeyHelper` equivalent (unlike Claude Code) and only resolves `{env:}`/`{file:}` config placeholders at process start, so a long-running session keeps sending whatever token it read on launch even after ucode's background thread rewrites the file with a fresh one.

## Changes

- **`databricks.py`** — wrap `--force-refresh` in a cross-process `flock` (per workspace+profile lockfile under `~/.ucode`), with a coalescing sentinel: if a peer refreshed within the last 60s, drop `--force-refresh` and reuse the cached token instead of redeeming again. Degrades gracefully to no-lock on Windows (no `fcntl`) or on `OSError`. The non-force read path is unchanged. The sentinel is marked on *attempt*, not success — Databricks rotates the refresh token on the redemption attempt reaching the server, not on our ability to parse a token out of the response, so marking only on success would leave a coalescing gap on a slow-but-successful peer.
- **`agents/opencode.py`** — ship a ucode-managed OpenCode plugin (`chat.headers` hook, auto-loaded from the plugin dir) that sets a fresh `Authorization` header on every chat request for our Databricks providers, sourced from `ucode auth-token` with an in-memory TTL cache (60s) + an in-flight promise guard so it doesn't shell out on every single request or spawn concurrent subprocesses. The static `apiKey`/`headers` in `opencode.json` are left as-is as a bootstrap/fallback the plugin overrides live and fails open to on any refresh error. `write_tool_config` now also writes the plugin, scoped to the provider ids actually configured.
- **`cli.py`** — `ucode revert` deletes the (fully-generated, no backup needed) plugin file.

## A note on how Axis B's bug was actually found

The generated plugin's provider-scoping guard originally read `input.provider?.info?.id`. Every unit test I wrote for it passed, because those tests only asserted on the *generated JS's string content* — one test literally asserted `"input.provider?.info?.id" in js` as the expected string. Against a real `opencode` process, `input.provider` is the runtime `Provider` record itself (`id`/`name`/`env`/`options`/`source`/`models`) — `id` lives directly on it, not nested under `.info`. The guard's `!providerId` check was therefore always true, so the hook returned immediately on every request and never refreshed anything — Axis B was a complete no-op in production.

This only surfaced once I ran the plugin end-to-end against the real opencode 1.17.10 binary in an isolated environment (separate `$HOME`/XDG dirs, real Databricks auth, no contact with any other live install) with a deliberately expired-but-well-formed JWT baked into the static config. It still failed with `Bad Request: Invalid Token` with the plugin loaded; diagnostic instrumentation showed the guard's `providerId` was always `undefined`. Fixed to `input.provider?.id` and reverified against the same harness — the plugin now fetches a fresh token and the request succeeds despite the stale static config.

Added `TestRenderAuthPluginRuntimeBehavior`, which shells out to a real `node` process and invokes the generated plugin's `chat.headers` hook against the actual runtime input shape (skips if `node` isn't on `PATH`). Verified it fails against the old `input.provider?.info?.id` line and passes against the fix, so this class of bug — assumed hook shape vs. real hook shape — now has coverage that isn't just a string-content assertion.

## Testing

- `uv run pytest`: 881 passed, 37 skipped, 1 pre-existing unrelated failure (`test_e2e_user_agent.py::TestClaudeUserAgent::test_user_agent_arrives_at_gateway`, a real-`claude`-binary e2e test that fails identically on `main` — verified via `git stash`).
- `uv run ruff check .` / `uv run ruff format --check` / `ty check src/`: all clean.
- Full local install (`uv tool install --reinstall .`) against a real Databricks workspace, plus an isolated end-to-end harness (separate `$HOME`, real auth, no contact with any live install) used to find and confirm the Axis B fix above.
- Live soak test: 3 concurrent long-lived `ucode opencode` sessions run for ~18 hours against a real workspace, crossing the ~60-minute token lifetime boundary roughly 17+ times each. Zero `Forbidden`/`Invalid Token`/`Unauthorized`/`Bad Request` auth failures in that entire window (verified via log inspection), with the background refresh cycling cleanly across all three sessions without collision.

## Follow-ups considered out of scope for this PR

- The plugin's 60s token cache is time-based, not response-aware — if a token is revoked mid-window, up to 60s of requests could fail before the cache expires and refetches. Bounded and strictly better than today's indefinite staleness, but a response-driven cache invalidation would need a different hook.
- `ucode revert`'s plugin-file deletion doesn't check dry-run mode (only relevant if `revert` ever runs under `--dry-run`).

### DIFF
--- a/src/ucode/agents/opencode.py
+++ b/src/ucode/agents/opencode.py
@@ -247,7 +247,13 @@ async function getToken() {{
 
 export const UcodeDatabricksAuth = async (_ctx) => ({{
   "chat.headers": async (input, output) => {{
-    const providerId = input.provider?.info?.id;
+    // NOTE: the hook's `input.provider` is the runtime Provider record
+    // itself (id/name/env/options/source/models), not the ProviderContext
+    // shape ({{source, info, options}}) some type surfaces suggest -- `id`
+    // lives directly on `input.provider`, not `input.provider.info.id`.
+    // Confirmed empirically against opencode 1.17.10; verify against a real
+    // hook invocation before changing this path again (#190 follow-up).
+    const providerId = input.provider?.id;
     if (!providerId || !MANAGED_PROVIDER_IDS.has(providerId)) {{
       return;
     }}
--- a/src/ucode/agents/opencode.py
+++ b/src/ucode/agents/opencode.py
@@ -222,7 +222,14 @@ let inflight = null;
 async function fetchToken() {{
   const [cmd, ...args] = UCODE_AUTH_TOKEN_ARGV;
   const {{ stdout }} = await execFileAsync(cmd, args);
-  return stdout.trim();
+  const token = stdout.trim();
+  if (!token) {{
+    // `ucode auth-token` exited 0 but printed nothing usable -- treat as a
+    // failure so the caller's catch block fails open to the static
+    // bootstrap token instead of caching/sending an empty bearer.
+    throw new Error("ucode auth-token returned empty output");
+  }}
+  return token;
 }}
 
 async function getToken() {{
@@ -259,6 +266,11 @@ export const UcodeDatabricksAuth = async (_ctx) => ({{
     }}
     try {{
       const token = await getToken();
+      // Defensive: ensure `output.headers` exists even if this opencode
+      // version ever invokes the hook without a pre-populated headers
+      // object, so the header still gets set rather than throwing (which
+      // would be caught below and silently no-op the refresh).
+      output.headers = output.headers || {{}};
       output.headers["Authorization"] = "Bearer " + token;
     }} catch (err) {{
       console.error("[ucode] failed to refresh Databricks token:", err);
--- a/src/ucode/agents/opencode.py
+++ b/src/ucode/agents/opencode.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import os
 import signal
 import subprocess
@@ -15,9 +16,11 @@ from ucode.config_io import (
     deep_merge_dict,
     read_json_safe,
     write_json_file,
+    write_text_file,
 )
 from ucode.databricks import (
     TOKEN_REFRESH_INTERVAL_SECONDS,
+    build_auth_token_argv,
     build_opencode_base_urls,
     get_databricks_token,
     model_token_limits,
@@ -30,6 +33,18 @@ OPENCODE_CONFIG_DIR = OPENCODE_XDG_CONFIG_HOME / "opencode"
 OPENCODE_CONFIG_PATH = OPENCODE_CONFIG_DIR / "opencode.json"
 OPENCODE_BACKUP_PATH = APP_DIR / "opencode-config.backup.json"
 OPENCODE_MCP_AUTH_HEADER_VALUE = "Bearer {env:OAUTH_TOKEN}"
+
+# Axis B of #190: opencode bakes provider.options.headers.Authorization into
+# opencode.json once at `ucode configure`/refresh time and never re-reads it
+# for the life of the process, so a long-lived session keeps sending a
+# bearer token that may since have been rotated or revoked. This
+# ucode-managed plugin's `chat.headers` hook overrides that static header on
+# every chat request with a freshly fetched (and short-TTL-cached) token.
+# The static opencode.json header stays in place as the bootstrap/fallback
+# the plugin fails open to if the refresh call errors.
+OPENCODE_PLUGIN_DIR = OPENCODE_CONFIG_DIR / "plugin"
+OPENCODE_PLUGIN_PATH = OPENCODE_PLUGIN_DIR / "ucode-databricks-auth.js"
+AUTH_PLUGIN_TOKEN_TTL_SECONDS = 60
 
 SPEC: ToolSpec = {
     "binary": "opencode",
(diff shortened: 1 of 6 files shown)


jesco39:
Addressed both review comments in `d259934`:

- `fetchToken()` now throws if `ucode auth-token` exits 0 with empty/whitespace stdout, so the hook fails open to the static bootstrap token instead of caching/sending an empty `Authorization: Bearer `.
- The `chat.headers` hook now defensively initializes `output.headers` before writing to it, so a missing headers object doesn't silently no-op the refresh.

Both are covered by new Node-executed runtime tests (`test_empty_token_output_fails_open_instead_of_setting_empty_bearer`, `test_initializes_missing_output_headers_instead_of_throwing`) — each verified to fail against the pre-fix code and pass against the fix.

Also rebased onto latest `main` to clear the out-of-date branch status.

jesco39:
Two more comments came in after the rebase (`7e0084b`) — both legitimate portability nits in the test harness, not production code:

- `_run_plugin()` hard-coded `python3` when patching `build_auth_token_argv` for the fake ucode binary — not guaranteed to be on PATH. Now uses `sys.executable`.
- The Node harness imported the generated plugin via `plugin_path.as_posix()`, not a valid ESM specifier on Windows. Now uses `Path.as_uri()`.

The other two comments on this thread (opencode.py lines 233/275) are stale duplicates of the ones already fixed in `d259934` — same comment IDs, diff_hunk frozen at review-creation time before that fix landed. No action needed there.

jesco39:
Addressed the 3 new comments in `d814390`:

- `_touch_sentinel()` now creates its own parent directory before touching — it could raise a swallowed `FileNotFoundError` whenever `fcntl` is unavailable (Windows), since `_refresh_lock` returns before ever creating `APP_DIR` on that branch, silently disabling coalescing on every such platform.
- `_refresh_lock()`'s acquisition loop now only retries on `BlockingIOError` (actual lock contention) instead of any `OSError`. An unexpected error (EBADF, or a filesystem without flock support) used to spin for the full 30s acquisition timeout before falling back — confirmed by timing the pre-fix behavior (~30s) vs. post-fix (<2s).
- `_remove_opencode_auth_plugin()` now raises `RuntimeError` on a real removal failure instead of silently swallowing it, matching `restore_file`'s existing convention elsewhere in `revert`.

All three have new tests, each verified to fail against the pre-fix code and pass against the fix.

PR #211, opened 2026-07-14
Add ucode codex-app to configure + launch the Codex desktop app

AarushiShah-db:
Adds `ucode codex-app` to point the Codex **desktop app** at Databricks and open it.

  ### Why it edits `config.toml` (not `ucode.config.toml`)
  The CLI (`ucode codex`) writes an isolated `~/.codex/ucode.config.toml` and launches with
  `codex --profile ucode`, so it never touches your real config. The **desktop app has no
  `--profile`** — the only file it reads on startup is the root `~/.codex/config.toml`.
  So to steer the app we have to edit that shared file in place. To make that safe, ucode backs it
  up first and uses a distinct provider name (`ucode-databricks-app`) so it coexists with the CLI's
  provider on the same file.

  ### What it does
  - Backs up `~/.codex/config.toml`, then merges in a Databricks provider block (base URL +
    `ucode auth-token` refresh command).
  - Opens Codex in the directory you ran the command from; prompts to restart it if it's already running.

  ### Undo (both restore from the backup)
  - `ucode codex-app --restore`
  - `ucode revert` (wired in too)

  macOS + Windows; Linux gives a clear error pointing to `ucode codex`. Tests added for the config
  write, revert, and launch/restart paths.

### DIFF
--- a/src/ucode/agents/codex.py
+++ b/src/ucode/agents/codex.py
@@ -523,20 +523,29 @@ def _quit_app() -> None:
         )
 
 
-def _open_app() -> bool:
-    """Open the Codex desktop app. Returns False when it couldn't be launched."""
+def _open_app(workdir: str | None = None) -> bool:
+    """Open the Codex desktop app, in ``workdir`` when given.
+
+    A GUI app launched via ``open``/``start`` does not inherit the terminal's
+    working directory (the launch is handed to launchd / the shell), so Codex
+    would otherwise open its own default folder. Passing the directory as a
+    positional argument tells Codex which repo to open.
+    """
     system = platform.system()
     try:
         if system == "Darwin":
-            result = subprocess.run(
-                ["open", "-b", CODEX_APP_BUNDLE_ID], capture_output=True, timeout=15
-            )
+            argv = ["open", "-b", CODEX_APP_BUNDLE_ID]
+            if workdir:
+                argv.append(workdir)
+            result = subprocess.run(argv, capture_output=True, timeout=15)
             return result.returncode == 0
         if system == "Windows":
-            # `start` is a cmd builtin; launch the Codex app by name.
-            result = subprocess.run(
-                ["cmd", "/c", "start", "", "Codex.exe"], capture_output=True, timeout=15
-            )
+            # `start` is a cmd builtin; launch the Codex app by name, passing the
+            # directory as an argument so it opens there.
+            argv = ["cmd", "/c", "start", "", "Codex.exe"]
+            if workdir:
+                argv.append(workdir)
+            result = subprocess.run(argv, capture_output=True, timeout=15)
             return result.returncode == 0
     except (OSError, subprocess.SubprocessError):
         return False
@@ -560,7 +569,7 @@ def _wait_for_app_exit(timeout: float) -> bool:
     return not _app_is_running()
 
 
-def _restart_app(prompt: str) -> None:
+def _restart_app(prompt: str, workdir: str | None = None) -> None:
     """Quit and reopen the app (with confirmation) so it re-reads config.toml."""
     if not prompt_yes_no(prompt):
         print_note("Quit and reopen Codex when you're ready for the change to take effect.")
@@ -569,7 +578,7 @@ def _restart_app(prompt: str) -> None:
     if not _wait_for_app_exit(CODEX_APP_QUIT_TIMEOUT):
         print_warning("Codex did not quit; quit it manually, then reopen it to apply the change.")
         return
-    if _open_app():
+    if _open_app(workdir):
         print_success("Codex restarted.")
     else:
         _warn_app_not_found()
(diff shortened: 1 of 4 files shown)


src/ucode/agents/codex.py
```diff
@@ -408,3 +433,187 @@ def validate_cmd(binary: str) -> list[str]:
...
+            )
+            return result.stdout.strip() == "true"
+        if system == "Windows":
+            result = subprocess.run(
+                ["tasklist", "/FI", "IMAGENAME eq Codex.exe"],
+                capture_output=True,
+                text=True,
+                timeout=10,
+            )
+            return "Codex.exe" in result.stdout
```
max-rozen-oss-db, line 504: nit - it would be cleaner to define an interface called `SystemCodex` with functions such as `start`, `quit`, is_running`, etc. Then you have 2 objects that implement this interface - `WindowsCodex` and `DarwinCodex` and those have OS-specific logic to operate on the Codex app. Now all OS-specific logic is in those classes/interfaces and adding another OS (Linux) requires only changing one place.

https://refactoring.guru/design-patterns/factory-method

src/ucode/cli.py
```diff
@@ -948,6 +954,105 @@ def _launch_tool(
...
+        profile=state.get("profile"),
+        tools=["codex"],
+        skip_model_discovery=bool(provider),
+        skip_preflight=skip_preflight,
+    )
+    if provider:
+        resolved_model = None
+    else:
+        state, resolved_model = resolve_launch_model("codex", state, None)
+    state = codex.write_app_config(state, resolved_model, provider=provider)
```
max-rozen-oss-db, line 990: Can we use different variables/subfunctions to manage 'state', instead of reassigning 'state' multiple times? This would make the code easier to follow

src/ucode/agents/codex.py
```diff
@@ -34,6 +39,27 @@
...
+CODEX_APP_MODEL_PROVIDER_NAME = "ucode-databricks-app"
+CODEX_APP_BUNDLE_ID = "com.openai.codex"
+# Seconds to wait for the app to quit before reopening it (mirrors Ollama).
+CODEX_APP_QUIT_TIMEOUT = 5.0
+
+CODEX_APP_MANAGED_KEYS: list[list[str]] = [
+    ["model_provider"],
+    ["model"],
+    ["model_providers", CODEX_APP_MODEL_PROVIDER_NAME],
+    ["model_providers", CODEX_APP_MODEL_PROVIDER_NAME, "http_headers"],
```
lilly-luo, line 60: i don't understand what a list of strings means for a key. could you add a comment?

src/ucode/agents/codex.py
```diff
@@ -408,3 +433,187 @@ def validate_cmd(binary: str) -> list[str]:
...
+
+def _quit_app() -> None:
+    system = platform.system()
+    if system == "Darwin":
+        subprocess.run(
+            ["osascript", "-e", 'tell application "Codex" to quit'],
+            capture_output=True,
+            timeout=10,
+        )
+    elif system == "Windows":
```
lilly-luo, line 518: ideally: add a case for Linux. if that's too much work, add an else case

src/ucode/agents/codex.py
```diff
@@ -408,3 +433,187 @@ def validate_cmd(binary: str) -> list[str]:
...
+    """
+    system = platform.system()
+    try:
+        if system == "Darwin":
+            argv = ["open", "-b", CODEX_APP_BUNDLE_ID]
+            if workdir:
+                argv.append(workdir)
+            result = subprocess.run(argv, capture_output=True, timeout=15)
+            return result.returncode == 0
+        if system == "Windows":
```
lilly-luo, line 542: if -> elif

src/ucode/agents/codex.py
```diff
@@ -408,3 +433,187 @@ def validate_cmd(binary: str) -> list[str]:
...
+            # `start` is a cmd builtin; launch the Codex app by name, passing the
+            # directory as an argument so it opens there.
+            argv = ["cmd", "/c", "start", "", "Codex.exe"]
+            if workdir:
+                argv.append(workdir)
+            result = subprocess.run(argv, capture_output=True, timeout=15)
+            return result.returncode == 0
+    except (OSError, subprocess.SubprocessError):
```
lilly-luo, line 550: add case for ubuntu or add an else throwing b/c it's not handled

src/ucode/cli.py
```diff
@@ -948,6 +954,105 @@ def _launch_tool(
...
 
+def _configure_codex_app(provider: str | None, skip_preflight: bool) -> None:
+    """Point the Codex desktop app at Databricks, then open/restart it.
+
+    Unlike `ucode codex` (which writes an isolated ~/.codex/ucode.config.toml and
+    passes --profile), the desktop app reads only the root ~/.codex/config.toml,
+    so we back that file up and edit it in place. `ucode codex-app --restore`
+    (or `ucode revert`) undoes it.
+    """
+    from ucode.agents import codex
```
lilly-luo, line 965: nit move this import up

lilly-luo (commented):
is it possible to run the restore_codex step any time the customer exits codex? would be nice to restore their .toml in case they're going to then use codex separately

hayescode:
Looking forward to this. I want to give lots of business users the Codex desktop app and having them in the terminal and editing the config.toml is proving to be a major challenge.

Ideally Codex would support Databricks SSO + workspace picker, but I understand that's on OpenAI.

Issue #215, opened 2026-07-16
AI Gateway MLflow chat-completions omits finish_reason on natural stop (breaks inkling in Pi)


dgokeeffe:
## Summary

The AI Gateway MLflow chat-completions endpoint (`/ai-gateway/mlflow/v1/chat/completions`) omits the OpenAI-spec-required `finish_reason` field on the terminal chunk of a **streaming** response when a model completes **naturally** (i.e. not truncated by `max_tokens`). The stream ends with only `data: [DONE]`.

This breaks any OpenAI-compatible streaming client that (correctly, per spec) requires a `finish_reason` before the stream terminates. Concretely it makes `databricks-inkling` unusable through the [Pi coding agent](https://pi.dev), which errors `Stream ended without finish_reason`.

## Affected model

- `databricks-inkling` (a reasoning model — emits `reasoning_content` deltas, then `content` deltas)

Non-reasoning endpoints on the same route (e.g. `databricks-glm-5-2`) **do** send `finish_reason` and work correctly, so this appears specific to inkling's serving stack.

## Reproduction (natural completion → NO finish_reason)

```
curl -sN "$WS/ai-gateway/mlflow/v1/chat/completions" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"model":"databricks-inkling","max_tokens":4000,"stream":true,
       "messages":[{"role":"user","content":"reply with exactly the two letters: ok"}]}'
```

Observed tail of the stream:

```
data: {"id":"chatcmpl-...","choices":[{"delta":{"role":"assistant","content":"ok"},"index":0,"logprobs":null}], ...}

data: [DONE]
```

`grep -c finish_reason` over the full stream returns **0** — no chunk carries it.

## Contrast: truncated completion DOES send finish_reason

With a small budget the model is cut off and the gateway *does* emit `finish_reason:"length"`:

```
curl -sN ... -d '{"model":"databricks-inkling","max_tokens":32,"stream":true, ...}'
...
data: {"...","choices":[{"delta":{"...","reasoning_content":" friendly"},"finish_reason":"length","index":0,...}], ...}
data: [DONE]
```

So the field is wired up — it's just missing on the **natural stop** path (should be `finish_reason:"stop"`).

## Expected

Per the OpenAI Chat Completions streaming spec, the final content chunk (or a final empty-delta chunk) before `[DONE]` should carry `finish_reason:"stop"` on normal completion.

## Impact

[...]

PR #217, opened 2026-07-16
feat(pi): add validated GLM/Kimi MLflow provider

### LINKED ISSUE #204: ucode pi automatic model detection
1. launching ucode pi automatically detects available models
2. the automatically detected models are not comprehensive
      "models": [
        {
          "id": "system.ai.claude-haiku-4-5"
        },
        {
          "id": "system.ai.claude-opus-4-8"
        },
        {
          "id": "system.ai.claude-sonnet-5"
        }
      ]
      "models": [
        {
          "id": "system.ai.gpt-oss-120b"
        },
        {
          "id": "system.ai.gpt-oss-20b"
        }
      ]
3. if you try to modify the providers manually, ucode just overwrites them on next launch.

dgokeeffe:
## Summary

Adds Pi support for the validated OSS coding-model cohort already used by ucode: **GLM and Kimi**.

Closes #204. Supersedes the earlier broad-cohort/proxy approach in this PR.

## Changes

- Add Pi's `databricks-mlflow` provider (`openai-completions` → `/ai-gateway/mlflow/v1`).
- Configure the gateway compatibility flags Pi needs: `supportsStore: false` and `supportsStrictMode: false`.
- Share GLM/Kimi context, output, and reasoning metadata across Pi and OpenCode.
- Let Pi select discovered OSS models and use them as the final default fallback.
- Keep OSS discovery as an explicit product allowlist: GLM and Kimi only.
- Keep the independent `gpt-oss` exclusion from the Codex/Responses bucket so it cannot be routed to an incompatible API.
- Include Pi in OSS discovery diagnostics.

## Review follow-up

- **Removed the local SSE-repair proxy entirely.** The missing `finish_reason` is an AI Gateway defect tracked by #215 and should not add an HTTP proxy to ucode's launch path.
- **Inkling is deferred until #215 is fixed.** It is not offered to Pi or OpenCode by this PR.
- Llama, Qwen, GPT-OSS, Gemma, and other chat models remain unlisted until they are explicitly validated for coding harnesses and accepted into ucode's supported model policy.
- Removed the unrelated version bump and `uv.lock` changes.

## Validation

- `uv run pytest tests/test_agent_pi.py tests/test_agent_opencode.py tests/test_databricks.py tests/test_cli.py -q`
- `uv run ruff check` on all changed Python files
- Final review-loop validation: `588 passed, 29 skipped` across the focused changed-area suite.
- The full stacked branch passes `1062 passed, 36 skipped` when installed-agent capture tests are excluded. Those tests depend on locally installed agent state rather than the unit under review.

## Scope

This is the base of the stack:

1. #217 — Pi GLM/Kimi provider and shared OSS metadata
2. #223 — Claude/GPT long-context metadata
3. #239 — OpenCode GPT routing and serving-endpoint fallback

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -1402,9 +1402,23 @@ _MODEL_SERVICE_REQUIRED_PREFIX = "system.ai."
 # is the only server-side narrowing that works.
 _MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"
 
-# Supported OSS chat families, matched by name substring. Add an entry to
-# support a new family.
+# OSS families validated as coding models in ucode, matched by name substring.
+# Keep this as an explicit product allowlist rather than exposing every model on
+# the chat-completions route. Inkling remains excluded until gateway issue #215
+# is fixed; other families require coding-harness validation before inclusion.
 _OSS_MODEL_FAMILIES = ("kimi-", "glm-")
+
+# Non-chat services must never be offered to a chat agent if a future supported
+# family also uses one of these substrings.
+_OSS_NON_CHAT_SUBSTRINGS = ("embedding", "embed", "rerank")
+
+
+def _is_oss_chat_model(model_id: str) -> bool:
+    """True if the id matches an OSS chat family and isn't a non-chat service."""
+    if any(bad in model_id for bad in _OSS_NON_CHAT_SUBSTRINGS):
+        return False
+    return any(family in model_id for family in _OSS_MODEL_FAMILIES)
+
 
 # Claude model families ucode buckets, newest tier first. Each maps to a
 # Claude Code family alias (ANTHROPIC_DEFAULT_<FAMILY>_MODEL). Add an entry to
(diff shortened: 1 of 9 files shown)


dgokeeffe:
@rohita5l @AarushiShah-db — would appreciate a review when you have a moment. This builds directly on your OSS work (#178 Kimi, #186 GLM): adds a Pi `databricks-mlflow` provider (Pi had none), broadens the OSS family bucket to the rest of the chat-completions cohort (inkling/llama/qwen/gpt-oss/gemma, excluding embeddings), corrects the GLM limits to the probed gateway caps (1M context / 65536 output), and adds a reasoning flag. One heads-up: it reverses the GLM output cap from #186's 25k to a verified 65536 — details in the PR description. A code-review pass was already applied (proxy framing/error-handling/SSRF-guard hardening). CI hasn't run — could use a maintainer approve-and-run on this fork PR.

src/ucode/agents/_mlflow_proxy.py
```diff
@@ -0,0 +1,201 @@
+"""Local SSE-repair proxy for the MLflow chat-completions gateway.
```
rohita5l: This is a gateway spec-compliance bug (missing `finish_reason` on the terminal SSE chunk) — it should be fixed in the gateway API, which is where #215 already tracks it. Shipping a ~200-line localhost HTTP proxy in the client to paper over it is the wrong layer: it only helps Pi (OpenCode already tolerates the omission), only on this ucode version, and adds real launch-path surface (threads, SSRF guard, the missing `urlopen` timeout, SSE byte-parsing) that we'll carry indefinitely — "delete once the gateway is fixed" shims rarely get deleted.

Since inkling appears to be the only model that triggers this, can we just not offer inkling until the gateway emits `finish_reason`, rather than merge the proxy? If there's a hard requirement to ship inkling now, let's at least make the proxy's removal contingent on #215 with an owner.

src/ucode/databricks.py
```diff
@@ -1162,8 +1162,31 @@ def build_auth_shell_command(
...
+# support a new family. Covers the full chat-completions-only cohort on the
+# `/ai-gateway/mlflow/v1` route (verified 2026-07-16): kimi, glm, inkling,
+# llama, qwen, gpt-oss, gemma.
+_OSS_MODEL_FAMILIES = (
+    "kimi-",
+    "glm-",
+    "inkling",
+    "llama-",
+    "qwen",
+    "gpt-oss-",
```
rohita5l: Push back on including `gpt-oss` `llama` and `qwen`.... gpt-oss isn't a frontier coding model — it's well behind the pack on coding benchmarks, so I'm not inclined to surface it as a selectable OSS model in a coding harness. Our OSS story is deliberately GLM / Kimi, which we're actively validating against coding harnesses; gpt-oss doesn't clear that bar.

Is there a specific customer requirement or blocker driving its inclusion? If not, I'd drop `"gpt-oss-"` from `_OSS_MODEL_FAMILIES` (and the corresponding `_MODEL_TOKEN_LIMITS` / `_OSS_REASONING_FAMILIES` entries). Note the codex-bucket exclusion below (`"gpt-oss" not in m`) can stay regardless — that guard is about not mis-routing it to the Responses API, independent of whether we offer it at all.

src/ucode/agents/_mlflow_proxy.py
dgokeeffe: Agreed. I removed the local SSE-repair proxy and its tests entirely. Pi now talks directly to the MLflow gateway, and Inkling is not offered while #215 remains open. The refreshed PR has no proxy threads, HTTP forwarding, timeout, or SSE parsing surface.

src/ucode/databricks.py
dgokeeffe: Agreed on keeping the coding-model product allowlist deliberate. The refreshed PR now offers only the validated GLM/Kimi cohort. GPT-OSS, Llama, Qwen, Gemma, and Inkling are excluded from selectable OSS discovery. I retained the independent `gpt-oss` Codex exclusion so it still cannot be misrouted to the Responses API.

dgokeeffe:
@rohita5l I refreshed this PR around the review feedback: removed the local proxy, deferred Inkling pending #215, narrowed OSS selection to GLM/Kimi, removed version/lockfile noise, and resolved both threads with details. The branch is rebased on current main and locally validated. Could you re-review and approve the fork CI run when convenient?

dgokeeffe:
Post-review update: a 3-round fresh-context review loop found and fixed two additional foundation issues in this layer: GLM 5.2 now has model-specific 1M/65536 limits while older/unknown GLM keeps 200k/25k, and Pi OSS-only workspaces now pass shared availability/diagnostic checks. Focused and full-stack validation are updated in the PR body.

dgokeeffe:
Rebased this PR onto current `main` (`67429ec`) and force-pushed the existing branch. New head: `036572c`.

### Stack order

This is **1 of 3** and must land first:

1. **#217** — Pi MLflow OSS provider (this PR)
2. #223 — shared Claude/GPT metadata + Pi GPT thinking-off fix
3. #239 — OpenCode GPT routing + OSS discovery fallback

#243 is independent.

If these are squash-merged, please rebase the next PR onto the updated `main` before merging it; the semantic order remains the same.

### Validation

- CI-parity suite: 1,648 passed, 6 skipped
- Only local failures: the two pre-existing CLI capture-server tests (`TestClaudeUserAgent`, `TestPiUserAgent`), which skip in CI's unit-test job
- `ruff check` and `ruff format --check`: clean
- No committed `uv.lock` churn or conflict markers

### Deliberately deferred managed-config integration

This PR adds the lower-level Pi OSS provider, but the managed-config resolver still treats managed Pi Kimi/GLM models as unservable and does not intersect admin allowlists with live gateway discovery. That broader policy/integration work is tracked separately in #290 rather than expanding this provider PR.

PR #218, opened 2026-07-16
Fix validate_tool hang: pass stdin=DEVNULL to the validation subprocess

Edwinhe03:
## What

`validate_tool` runs each tool's smoke test (`pi --print …`, `codex exec …`) via `subprocess.run(..., timeout=60)` without setting `stdin`, so the child inherits ucode's stdin. pi and codex open stdin in print/exec mode and block on it. When ucode is launched from a non-interactive parent whose stdin is an open pipe with no EOF (e.g. an agent harness spawning `ucode <tool>`), validation hangs the full 60s and returns a spurious `timed out` → `<tool> validation failed — config reverted`, even though the tool is configured correctly.

## Fix

Pass `stdin=subprocess.DEVNULL` — the smoke test is one-shot and never needs input. Benign for every tool.

## Verified

Open-stdin-pipe condition, before (`main`) vs after:

| tool | before | after |
|---|---|---|
| pi (real `validate_tool`) | `timed out`, 60s | ok, 0s |
| codex (real `validate_tool`) | `timed out`, 60s | ok, 0s |
| real `pi --print` on sandbox | hang, 60s | reply, 2s |
| real `codex exec` on sandbox | hang, 60s | reply, 3s |

Added `TestValidateTool` (asserts `stdin=DEVNULL`; `TimeoutExpired` → `timed out`). Full suite: 875 passed, ruff clean.

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -465,14 +465,6 @@ def validate_tool(tool: str) -> tuple[bool, str]:
         except RuntimeError:
             env = None
     try:
-        # stdin=DEVNULL: the validate command is a one-shot smoke test (e.g.
-        # `pi --print "say hi..."`) that never needs interactive input. Some
-        # agent CLIs still open stdin in print mode and block on it; without
-        # this, `subprocess.run` inherits our stdin, so when ucode itself is
-        # launched from a non-interactive parent whose stdin is an open pipe
-        # with no EOF (e.g. an agent harness spawning `ucode <tool>`), the
-        # validation subprocess hangs until the 60s timeout and reports a
-        # spurious "timed out" / "validation failed".
         result = subprocess.run(
             cmd,
             check=False,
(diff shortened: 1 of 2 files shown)


src/ucode/agents/__init__.py
```diff
@@ -465,8 +465,22 @@ def validate_tool(tool: str) -> tuple[bool, str]:
...
             env = None
     try:
+        # stdin=DEVNULL: the validate command is a one-shot smoke test (e.g.
+        # `pi --print "say hi..."`) that never needs interactive input. Some
+        # agent CLIs still open stdin in print mode and block on it; without
+        # this, `subprocess.run` inherits our stdin, so when ucode itself is
+        # launched from a non-interactive parent whose stdin is an open pipe
+        # with no EOF (e.g. an agent harness spawning `ucode <tool>`), the
+        # validation subprocess hangs until the 60s timeout and reports a
+        # spurious "timed out" / "validation failed".
```
lilly-luo: can u rm all this

lilly-luo (dismissed):

lilly-luo (approved):

PR #223, opened 2026-07-18
fix(pi): omit unsupported GPT off-reasoning effort

### LINKED ISSUE #286: pi: gpt-5/mini/nano/5-5-pro always 400 — reasoning.effort "none" sent because thinkingLevelMap is unset
## Summary

`ucode pi` with `databricks-openai/system.ai.gpt-5`, `-gpt-5-mini`, `-gpt-5-nano` or `-gpt-5-5-pro` fails on every request with:

```
OpenAI API error (400): 400 status code (no body)
```

Cause: `_pi_gpt_model_entry` declares `reasoning: True` but never sets `thinkingLevelMap`, so when thinking is off Pi sends `reasoning: {effort: "none"}`, which those four models reject.

## Root cause

`src/ucode/agents/pi.py:157-172`:

```python
def _pi_gpt_model_entry(model_id: str) -> dict:
    entry: dict = {"id": model_id, "contextWindow": ..., "maxTokens": ...}
    if "gpt-5" in model_id.lower().replace(".", "-"):
        entry["reasoning"] = True
        entry["input"] = ["text", "image"]
    return entry
```

Pi's contract (`pi-ai/dist/api/openai-responses.js:243-247`) is that the model entry declares its own off-state; absent `thinkingLevelMap` it defaults to `"none"`:

```js
else if (model.provider !== "github-copilot" && model.thinkingLevelMap?.off !== null) {
    params.reasoning = { effort: (model.thinkingLevelMap?.off ?? "none") };
}
```

`"none"` is only valid on gpt-5.1+. We never set the override, so older ids get an invalid value.

## Measured, against `/ai-gateway/codex/v1/responses`

| Model | `effort:"none"` | `effort:"minimal"` | omitted |
|---|---|---|---|
| `gpt-5`, `gpt-5-mini`, `gpt-5-nano` | **400** | 200 | 200 |
| `gpt-5-5-pro` | **400** | **400** | 200 |
| `gpt-5-1/-2/-3-codex/-4/-4-mini/-4-nano/-5/-6-luna/-6-sol/-6-terra` | 200 | 200 | 200 |

Gateway response on failure:

```json
{"error_code":"BAD_REQUEST","message":"{\n  \"error\": {\n    \"message\": \"Unsupported value: 'none' is not supported with the 'gpt-5' model. Supported values are: 'minimal', 'low', 'medium', and 'high'.\"..."}
```

I bisected the rest of Pi's Responses payload against the gateway and it is all fine: `store:false`, `prompt_cache_key`, `prompt_cache_retention:"24h"`, `prompt_cache_options`, `include:["reasoning.encrypted_content"]`, `developer` role, flat tool schemas with `additionalProperties:false`, and the underscore `session_id` / `x-client-request-id` affinity headers — all 200.

## Why the error message is useless

Pi's `error-body.js` recovery only reads `error.error`, and the gateway wraps its payload as `{"error_code", "message"}`. So `e.error === undefined`, the body is discarded, and *every* gateway 400 renders identically. Filed upstream separately. That is why this took so long to pin down, but the config bug is ours.

## Proposed fix

Set the off-state explicitly in `_pi_gpt_model_entry`. `thinkingLevelMap: {"off": None}` makes Pi omit `reasoning` entirely (the `?.off !== null` branch above) — verified 200 on all 14 ids, and it's the same pattern we already use for the Gemini 3.x entries.

```python
if "gpt-5" in model_id.lower().replace(".", "-"):
    entry["reasoning"] = True
    entry["input"] = ["text", "image"]
    entry["thinkingLevelMap"] = {"off": None}
```

`{"off": "minimal"}` is the alternative if we want to keep an explicit low-reasoning signal, but it still 400s on `gpt-5-5-pro`, so it would need per-model handling.

Worth noting `databricks-openai` is the only provider block in `pi.py` with no `compat`/capability hardening — `databricks-claude` needed `supportsEagerToolInputStreaming: False` and `databricks-mlflow` needed `supportsStore`/`supportsStrictMode: False`. Same class of omission. The docstrings at `pi.py:12-22` and the note at `databricks.py:2446` ("required for `oss` only") should be updated too.

Happy to put up the PR.

## Environment

- ucode 0.1.0, pi 0.83.0
- workspace: `dbc-a5d4177a-49dc.cloud.databricks.com`
- config: `~/.ucode/pi-home/.pi/agent/models.json`

dgokeeffe:
## Issue and stack

Closes #286.

Depends on #217 and #333. Review only the concern-isolated two-file incremental range:

https://github.com/dgokeeffe/ucode/compare/f299e50...94fe107

Do not merge before #333. This PR was previously described as the context-window PR; #333 now owns that concern, while this existing head is retained for the GPT thinking-off compatibility commit without rewriting history.

---
# Unit 02 — Pi GPT reasoning compatibility

## Objective and user-visible behavior

Prevent Pi's thinking-off mode from emitting unsupported `reasoning.effort: "none"` for GPT gateway models. Explicit `thinkingLevelMap: {"off": null}` makes Pi omit the reasoning field when thinking is off.

Candidate: `review/pi-gpt-reasoning` (`94fe107`), mandatory PR target `review/shared-context-policy` (`f299e50`). This is a concern-isolated two-file stacked unit and must not remain conflated with context-policy PR #223.

## Exact scope

Production: `src/ucode/agents/pi.py`.
Tests: `tests/test_agent_pi.py`.
Non-goals: token/context windows (unit 03), OpenCode, OSS models, MLflow proxying, cache retention.

## Before / after reproduction

```bash
git show f299e50:src/ucode/agents/pi.py | rg 'thinkingLevelMap' || echo '<absent>'
# <absent>

git show review/pi-gpt-reasoning:src/ucode/agents/pi.py | rg 'thinkingLevelMap'
# entry["thinkingLevelMap"] = {"off": None}
```

Focused validation:

```bash
uv run --frozen pytest tests/test_agent_pi.py -q
# 63 passed
uv run --frozen ruff check .
# All checks passed
```

Full suite at the stacked candidate:

```bash
uv run --frozen pytest -q
# 2 failed, 1616 passed, 36 skipped
```

Failures are the installed Claude and Pi User-Agent capture tests. This older stack does not contain unit 01; the integrated `dev` baseline passes the Pi capture test, while independently reproducing the Claude capture failure.

## Live Databricks evidence

Retained prior redacted gateway evidence: GPT-5-family requests with explicit effort `none` returned `BAD_REQUEST`; omitting `reasoning` was accepted on `/ai-gateway/codex/v1`. No bearer tokens or workspace identifiers are stored. The deterministic tests cover GPT-5, mini, nano, 5.5-pro, and 5.6 IDs plus a non-GPT control.

## Impact map

- Agent/provider: Pi `databricks-openai` / Responses only.
- Config: one per-model metadata field in generated `models.json`.
- Compatibility: affects only models already marked `reasoning`; non-GPT Responses entries remain unchanged.
- Managed config/MCP/skills: no changes.

## Rollback and residual risk

Revert `94fe107`; rollback risks restoring gateway 400s when users select thinking off. Residual risk: future Pi semantics for a null off mapping could change, so the generated model-entry contract should remain covered by tests.

## Hygiene

`git diff --check review/shared-context-policy..review/pi-gpt-reasoning` passes. The PR must target `review/shared-context-policy`; its target-relative diff is the review contract. Two-file range; no generated files, credentials, `uv.lock`, `.pi-subagents/`, or `goal.md`; no merge markers or unresolved index entries.

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -27,6 +27,7 @@ from ucode.config_io import (
 from ucode.databricks import (
     build_auth_shell_command,
     build_tool_base_url,
+    claude_model_supports_1m,
     get_databricks_token,
 )
 from ucode.launcher import exec_or_spawn
@@ -83,11 +84,6 @@ def _resolve_web_search_model(state: dict) -> str | None:
 
 
 WEB_SEARCH_MCP_NAME = "web_search"
-# Matches both the AI Gateway form (`databricks-claude-opus-4-8`) and the UC
-# model-services form (`system.ai.claude-opus-4-8`).
-_CLAUDE_MODEL_RE = re.compile(
-    r"^(?:system\.ai\.)?(?:databricks-)?claude-(opus|sonnet)-(\d+)-(\d+)(.*)$"
-)
 
 # Env keys the MLflow Stop hook reads to route traces. Written into the
 # settings `env` block alongside the hook itself.
@@ -390,19 +386,9 @@ def render_overlay(
 
 
 def _maybe_add_1m_suffix(model: str) -> str:
-    if model.endswith("[1m]"):
+    if model.endswith("[1m]") or not claude_model_supports_1m(model):
         return model
-    match = _CLAUDE_MODEL_RE.match(model)
-    if not match:
-        return model
-
-    family, major_raw, minor_raw, _ = match.groups()
-    major = int(major_raw)
-    minor = int(minor_raw)
-    should_suffix = (family == "opus" and (major, minor) >= (4, 6)) or (
-        family == "sonnet" and (major, minor) >= (4, 6)
-    )
-    return f"{model}[1m]" if should_suffix else model
+    return f"{model}[1m]"
 
 
 def _register_web_search_mcp(workspace: str, search_model: str, profile: str | None = None) -> bool:
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -1402,9 +1402,23 @@ _MODEL_SERVICE_REQUIRED_PREFIX = "system.ai."
 # is the only server-side narrowing that works.
 _MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"
 
-# Supported OSS chat families, matched by name substring. Add an entry to
-# support a new family.
+# OSS families validated as coding models in ucode, matched by name substring.
+# Keep this as an explicit product allowlist rather than exposing every model on
+# the chat-completions route. Inkling remains excluded until gateway issue #215
+# is fixed; other families require coding-harness validation before inclusion.
 _OSS_MODEL_FAMILIES = ("kimi-", "glm-")
+
+# Non-chat services must never be offered to a chat agent if a future supported
+# family also uses one of these substrings.
+_OSS_NON_CHAT_SUBSTRINGS = ("embedding", "embed", "rerank")
+
+
+def _is_oss_chat_model(model_id: str) -> bool:
+    """True if the id matches an OSS chat family and isn't a non-chat service."""
+    if any(bad in model_id for bad in _OSS_NON_CHAT_SUBSTRINGS):
+        return False
+    return any(family in model_id for family in _OSS_MODEL_FAMILIES)
+
 
 # Claude model families ucode buckets, newest tier first. Each maps to a
 # Claude Code family alias (ANTHROPIC_DEFAULT_<FAMILY>_MODEL). Add an entry to
@@ -1440,21 +1454,46 @@ def classify_model_family(model_id: str) -> str | None:
 # config dialect. Both fields are provided because agents like OpenCode require
 # context and output together. Keyed by family substring; add an entry to bound
 # a new model.
+#
+# Output caps probed from the gateway 2026-07-16 (it 400s with "max_tokens (N)
+# cannot exceed <cap>"); context windows from each model's docs/description
+# (conservative when unstated). If the gateway raises a cap or ships a new
+# model, update this table.
 _MODEL_TOKEN_LIMITS: dict[str, dict[str, int]] = {
-    # GLM-4.6: 200k context, but the gateway caps output well below the model's
-    # native 128k — pin 25k so requests aren't rejected.
+    # Keep the version-specific entry before the family fallback: GLM 5.2 has
+    # materially higher probed gateway limits than earlier/unknown variants.
+    "glm-5-2": {"context": 1_000_000, "output": 65_536},
     "glm": {"context": 200_000, "output": 25_000},
+    "kimi": {"context": 128_000, "output": 65_536},
 }
+
+# Conservative fallback for a future variant that matches a validated family
+# but has no specific entry. Pinning a low output ceiling risks truncation, not
+# a gateway 400, so it is the safe failure direction.
+_OSS_FALLBACK_LIMITS = {"context": 128_000, "output": 8_192}
+
+# Validated families that emit reasoning. Pi renders their streamed
+# reasoning_content as thinking when the model entry sets reasoning:true.
+_OSS_REASONING_FAMILIES = ("glm", "kimi")
+
+
+def model_is_reasoning(model_id: str) -> bool:
+    """True if the OSS model reports reasoning output (family-matched)."""
+    return any(family in model_id for family in _OSS_REASONING_FAMILIES)
 
 
 def model_token_limits(model_id: str) -> dict[str, int] | None:
     """Return ``{"context": ..., "output": ...}`` limits for ``model_id``, or None.
 
-    Matches by family substring (e.g. any ``*glm*`` id). None means the model
-    has no known limits and the agent should not pin any."""
+    Prefers a specific `_MODEL_TOKEN_LIMITS` family entry (e.g. any ``*glm*``
+    id). Any other OSS chat model falls back to a conservative floor so it is
+    never offered uncapped (which would 400). None only for non-OSS ids, where
+    the agent should not pin any limit."""
     for family, limits in _MODEL_TOKEN_LIMITS.items():
         if family in model_id:
             return dict(limits)
+    if _is_oss_chat_model(model_id):
+        return dict(_OSS_FALLBACK_LIMITS)
     return None
 
 
@@ -1737,10 +1776,16 @@ def discover_model_services(
     # newest-wins once the router accepts opus-5 (PR databricks-eng/universe#2365446).
     _prefer_opus_4_8(claude_models, ids)
 
-    codex_models = sorted([m for m in ids if "gpt-" in m], key=model_version_sort_key)
+    # `gpt-oss-*` also contains "gpt-" but is a chat-completions-only OSS model
+    # (served via /ai-gateway/mlflow/v1), NOT an openai-responses codex model —
+    # exclude it here so it isn't offered under the codex provider (which 400s).
+    codex_models = sorted(
+        [m for m in ids if "gpt-" in m and "gpt-oss" not in m],
+        key=model_version_sort_key,
+    )
     gemini_models = sorted([m for m in ids if "gemini-" in m], key=model_version_sort_key)
 
-    oss_models = [m for m in ids if any(family in m for family in _OSS_MODEL_FAMILIES)]
+    oss_models = [m for m in ids if _is_oss_chat_model(m)]
 
     if not (claude_models or codex_models or gemini_models or oss_models):
         sample = ", ".join(ids[:5])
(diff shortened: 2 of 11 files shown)


src/ucode/databricks.py
```diff
@@ -1149,24 +1172,130 @@ def build_auth_shell_command(
...
+    for family, limits in _GPT_TOKEN_LIMITS:
+        if tail.startswith(family):
+            return dict(limits)
+    return dict(_GPT_FALLBACK_LIMITS)
+
+
+_CLAUDE_FALLBACK_LIMITS = {"context": 200_000, "output": 64_000}
+
+
+def claude_model_token_limits(model_id: str) -> dict[str, int]:
```
rohita5l: The "which Claude versions get 1M" policy is now encoded in two places with two different regexes: here (Opus ≥4.6, Sonnet ≥4.6→128k, Sonnet ≥4.5→64k) and in `claude.py:_maybe_add_1m_suffix` (Opus ≥4.6, Sonnet ≥4.5). They agree today, but the next model tier (e.g. Opus 5, or Sonnet gaining 128k output) has to be updated in both, and nothing fails if you only touch one. Consider a single shared predicate/table in `databricks.py` that both the Claude Code suffix logic and Pi's metadata consume, so the version floors live in one spot.

Minor, related: the branch order here is asymmetric — Sonnet has an explicit `>=(4,5)` branch but Opus 4.5 silently falls through to `_CLAUDE_FALLBACK_LIMITS`. That's correct (Opus 4.5 predates 1M), but a one-line comment on the fallback would save the next reader from wondering if the Opus 4.5 case was forgotten.

src/ucode/agents/claude.py
```diff
@@ -251,8 +251,12 @@ def _maybe_add_1m_suffix(model: str) -> str:
     family, major_raw, minor_raw, _ = match.groups()
     major = int(major_raw)
     minor = int(minor_raw)
+    # 1M-token context window support: Opus from 4.6, Sonnet from 4.5 (Sonnet
+    # 4.5 gained the 1M beta window before Opus did). Without the `[1m]`
+    # suffix the gateway serves these models with the default 200k context, so
+    # gate each family at its own floor rather than a shared version.
     should_suffix = (family == "opus" and (major, minor) >= (4, 6)) or (
-        family == "sonnet" and (major, minor) >= (4, 6)
+        family == "sonnet" and (major, minor) >= (4, 5)
```
rohita5l: Flagging for the agent owner's awareness: this lowers the Sonnet `[1m]` floor from 4.6→4.5, which changes `ANTHROPIC_DEFAULT_SONNET_MODEL` for the **Claude Code** agent — not Pi. The change is correct (Sonnet 4.5 has the 1M beta window and it's test-covered), but since this PR is framed as "Pi long-context metadata," it's easy to miss that it also alters Claude Code's default model selector.

src/ucode/databricks.py
dgokeeffe: Addressed. Claude family/version parsing and capabilities now live in one shared helper in `databricks.py`. Pi limits, Pi adaptive-thinking compatibility, and Claude Code `[1m]` suffix selection consume the same result. Table-driven tests cover Opus 4.5/4.6/5, Sonnet 4.4/4.5/4.6/5, Haiku, Fable fallback, prefixes, malformed IDs, and existing suffixes. The fallback docstring explicitly calls out Opus 4.5.

src/ucode/agents/claude.py
dgokeeffe: Acknowledged. The Sonnet 4.5 change is intentional and affects Claude Code as well as Pi. I updated the PR description with a dedicated Claude Code behavior section and retained the positive/negative coverage for the Sonnet 4.5 and Opus 4.6 family-specific floors.

dgokeeffe:
@rohita5l I refreshed this stacked PR: Claude capability/version policy is now shared by Pi and Claude Code, the Sonnet 4.5 Claude Code behavior is explicit in the description, and both threads have detailed replies and are resolved. Could you re-review and approve the fork CI run when convenient?

dgokeeffe:
Post-review update: shared capability policy now reflects Sonnet 4.6 at 1M/64k and Fable 5 at native 1M/128k with adaptive thinking but no `[1m]` suffix. GPT matching is boundary-safe/case-insensitive, excludes GPT-OSS from Responses selection, and preserves generic Responses endpoints while preferring the newest semantic GPT. PR body/comparison link updated.

dgokeeffe:
Pushed `8d66253` to fix a 400 this PR would otherwise introduce.

`_pi_gpt_model_entry` set `reasoning: True` without a `thinkingLevelMap`. Pi's Responses builder treats a missing off-state as `effort: "none"` (`pi-ai/dist/api/openai-responses.js`, the `thinkingLevelMap?.off !== null` branch):

```js
else if (model.provider !== "github-copilot" && model.thinkingLevelMap?.off !== null) {
    params.reasoning = { effort: (model.thinkingLevelMap?.off ?? "none") };
}
```

`"none"` is only valid on gpt-5.1+, so with thinking off, four ids rejected every request:

```
BAD_REQUEST: Unsupported value: 'none' is not supported with the 'gpt-5' model.
Supported values are: 'minimal', 'low', 'medium', and 'high'.
```

Measured against `/ai-gateway/codex/v1/responses`:

| Model | `effort:"none"` | `effort:"minimal"` | omitted |
|---|---|---|---|
| `gpt-5`, `gpt-5-mini`, `gpt-5-nano` | **400** | 200 | 200 |
| `gpt-5-5-pro` | **400** | **400** | 200 |
| `gpt-5-1/-2/-3-codex/-4/-4-mini/-4-nano/-5/-6-luna/-6-sol/-6-terra` | 200 | 200 | 200 |

Fix is `entry["thinkingLevelMap"] = {"off": None}`, which makes Pi omit `reasoning` entirely — 200 on all 14 ids. `{"off": "minimal"}` was the other option but `gpt-5-5-pro` rejects it too, so it would need a hardcoded per-model exclusion list against a dynamically discovered model set. Same `{"off": None}` pattern we already use for the Gemini 3.x entries.

Also bisected the rest of Pi's Responses payload against the gateway; all 200, no other compat flags needed: `store:false`, `prompt_cache_key`, `prompt_cache_retention:"24h"`, `prompt_cache_options`, `include:["reasoning.encrypted_content"]`, `developer` role, flat tool schemas with `additionalProperties:false`, and the underscore `session_id` / `x-client-request-id` affinity headers.

Added two regression tests in `tests/test_agent_pi.py` (every gpt-5 entry pins `{"off": None}`; non-gpt codex entries stay bare) plus a docstring note on why the override is mandatory whenever `reasoning: True` is declared.

`uv run pytest --ignore=tests/test_e2e.py --ignore=tests/test_e2e_tracing.py` → 1102 passed, 6 skipped. The two `test_e2e_user_agent` failures I see locally are pre-existing on this branch (identical at `4984d37` without my commit) and need local agent binaries, so they skip in the `test` job.

Two things for reviewers:

1. **`_pi_claude_model_entry` in this same diff has the same shape** — `reasoning: True`, no `thinkingLevelMap`. It doesn't bite today because Claude goes through `anthropic-messages`, which never reaches the Responses reasoning branch. Left alone deliberately, but it's a latent trap if that ever moves.
2. The trade in this fix: thinking-**off** is now delegated to the gateway default rather than explicitly requested as zero reasoning. That seemed clearly better than 400ing, but flagging it in case exact off-semantics matter.

Worth noting this was painful to diagnose because the gateway returns `{"error_code","message"}` instead of OpenAI's `{"error":{...}}`, so Pi's `error-body.js` recovery no-ops and every 400 renders as `OpenAI API error (400): 400 status code (no body)`. Reported upstream: earendil-works/pi#7748. Tracking issue: #286.

dgokeeffe:
Rebased this PR stack onto current `main` (`67429ec`) and force-pushed the existing branch. New head: `94fe107`.

### Stack dependency

This is **2 of 3** and must land after #217:

1. #217 — Pi MLflow OSS provider
2. **#223** — shared Claude/GPT metadata + Pi GPT thinking-off fix (this PR)
3. #239 — OpenCode GPT routing + OSS discovery fallback

The rebase preserves main's #282 contract: Pi pins `codex_models[0]` verbatim; it does not re-sort for the newest GPT. The `preferred_gpt_model` helper is not part of this PR anymore—it is introduced with its OpenCode consumer in #239.

`94fe107` contains the `thinkingLevelMap: {"off": None}` fix discussed above and tracked in #286. It prevents Pi from sending unsupported `reasoning: {effort: "none"}` to gpt-5/-mini/-nano/-5-5-pro.

If #217 is squash-merged, please rebase this branch onto the updated `main` before merging.

### Validation

- CI-parity suite: 1,648 passed, 6 skipped
- Only local failures: the two pre-existing CLI capture-server tests (`TestClaudeUserAgent`, `TestPiUserAgent`), which skip in CI's unit-test job
- `ruff check` and `ruff format --check`: clean
- No committed `uv.lock` churn or conflict markers

Managed allowlist × live-discovery intersection is deliberately outside this PR and tracked in #290.

PR #224, opened 2026-07-20
Add --model launch flag for custom UC model services

grundprinzip:
Add a launch-time `--model <catalog>.<schema>.<name>` flag to both `ucode claude` and `ucode codex` that pins a custom (non-`system.ai`) Unity Catalog model-service as the agent's default model, while still discovering and listing the `system.ai` models in the picker. Sibling to the existing `--provider` flag and mutually exclusive with it.

- databricks.py: scope `list_model_services` with `parent=schemas/system.ai` (keep page_size + client-side filter as safety nets); add `get_model_service` for an existence-only singular lookup with a schema-scoped listing fallback.
- cli.py: `--model` option on both commands, threaded through `_launch_tool`; mutual-exclusion guard vs `--provider`; validate the id before launch while keeping normal system.ai discovery.
- agents/claude.py: pin `ANTHROPIC_MODEL` to the custom id alongside the existing family aliases so system.ai rows remain in the picker.
- agents/codex.py: pin the custom id verbatim, bypassing the OpenAI-id rewrite.
- Tests: +15 across databricks/claude/codex/cli covering discovery scope, singular lookup, per-agent pinning, and CLI validation/mutual-exclusion.

Co-authored-by: Isaac

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -287,17 +287,23 @@ def configure_tool(
     model: str | None = None,
     provider: str | None = None,
     provider_models: dict[str, str] | None = None,
+    custom_model: str | None = None,
 ) -> dict:
     result: dict | tuple[dict, str]
     if tool == "codex":
-        result = codex.write_tool_config(state, model, provider=provider)
+        result = codex.write_tool_config(state, model, provider=provider, custom_model=custom_model)
     elif tool == "claude":
         # A Model Provider Service routes by header and pins no Databricks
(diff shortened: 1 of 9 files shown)


lilly-luo:
Hi @grundprinzip - #306 already added this functionality for Claude and the behavior was native for Codex. Against workspace https://eng-ml-inference.staging.cloud.databricks.com for example: 

```
ucode claude --model main.default.lilly
ucode codex --model main.alex_large_schema_2.alex_custom_ms198
``` 

You'll need to `rm -rf /etc/claude-code/managed-settings.json` to test claude.

PR #225, opened 2026-07-20
Improve `configure mcp` UX: search wizard, faster/robust discovery & apply

nachill-databricks:
## Summary

Overhauls the MCP setup experience in ucode. Fixes crashes and slowness that show up on large workspaces (100s of MCP servers), reworks the interactive picker into a clearer wizard, and surfaces MCP setup as a natural step of `ucode configure` so users don't have to know `configure mcp` exists.

### 1. Fixed a crash during discovery
A network read timeout raises a bare `TimeoutError` (an `OSError`, **not** a `URLError`), which slipped past the error handling in `_http_get_json` and killed the whole command mid-discovery (e.g. "timed out Discovering Vector Search…"). Timeouts are now caught and treated as a skippable failure, like every other discovery error.

### 2. New two-step "search wizard" for `configure mcp`
The interactive flow is now:

1. **`Search for:` screen** — pick which sources to search. External connections, Databricks apps, MCP services, and Genie spaces are **pre-checked**; Vector Search and UC functions are **off by default** because they walk the whole workspace and are slow. Just press Enter to accept the defaults.
2. **Server picker** — choose servers from what was found.

Extras:
- Press **← (Left arrow)** in the picker to go **back** to the search screen without restarting the command.
- **Ctrl-A** toggles select-all / select-none in the picker.

### 3. Better feedback while searching and saving
- **Live progress counters** on the slow, walk-based sources: the spinner now shows e.g. `Searching Vector Search… 3/8 endpoints, 2 found` instead of a blank spinner.
- **Post-save summary**: `Added 2, removed 1 MCP servers across Claude Code, Codex` instead of a bare `Saved`.
- Picker labels are now consistently prefixed: `MCP:` for MCP services and `Connection:` for external connections (matching the existing `App:` / `Genie:` / `Vector Search:` labels).

### 4. Fixed the "frozen" apply for many servers
Saving 100+ servers appeared to hang because ucode invokes each tool's CLI once per server (claude/codex/gemini), serially. `apply_mcp_server_changes` now runs those operations **concurrently across clients** (serial *within* a client, since each mutates one shared config file) with a live `Configuring N/M` progress spinner. It no longer looks frozen. (Per-server CLI spawns remain — collapsing them into direct config-file writes is a possible future follow-up, noted in code.)

### 5. MCP offered at the end of `ucode configure`
After workspace + agents are set up in the **fully-interactive** `ucode configure` (no `--agent`/`--agents`/`--workspaces` flags), it now asks **`Configure MCP servers now?`** and runs the wizard if accepted. Flag-driven / scripted runs and `--dry-run` are unaffected — they never prompt.

### Also added
- `list_all_mcp_services`: a workspace-wide MCP-services walk (catalogs → schemas → mcp-services) under a wall-clock budget, feeding the "MCP services" source.

## Testing
- `uv run pytest` — green except 3 `tests/test_cli.py::TestStatus`/`TestRevert` failures that are **pre-existing on `main`** (verified by stashing this branch's changes) and unrelated to this work.
- `uv run ruff check .`, `uv run ruff format --check`, and `uv run ty check src/` all pass.


src/ucode/databricks.py
```diff
@@ -1542,6 +1554,9 @@ def map_bedrock_claude_models(targets: list[str]) -> dict[str, str]:
 _UC_FUNCTION_PROBE_TIMEOUT = 5
 _VECTOR_SEARCH_DEADLINE_SECONDS = 15.0
 _UC_FUNCTIONS_DEADLINE_SECONDS = 20.0
+# This walk runs on every `configure mcp` (not opt-in), so keep the budget tight:
```
sunishsheth2009: Should do this for `-mcp` as well along with `configure mcp` unless this already works for both?

src/ucode/databricks.py
```diff
@@ -1784,6 +1821,111 @@ def collect_pair(has_fn, pair):
     return sorted(pairs), None
 
 
+def list_all_mcp_services(
```
sunishsheth2009, line 1932: In the long run maybe we can only MCP servers from system.ai as default via this interactive flow? If we add 100+ MCP service, this list will be long :D 
So might be easier to focus on system.ai maybe and let customers manually add other?

src/ucode/databricks.py
```diff
@@ -1784,6 +1821,111 @@ def collect_pair(has_fn, pair):
...
+    workspace: str,
+    token: str,
+    *,
+    deadline_seconds: float = _MCP_SERVICES_WALK_DEADLINE_SECONDS,
+    on_progress: Callable[[int, int, int], None] | None = None,
+) -> tuple[list[str], str | None]:
```
sunishsheth2009, line 1938: If it always returns back None for the reason, Do we need to return that back. Basically can this just be:

`list[str] | None:`

src/ucode/mcp.py
```diff
@@ -1010,24 +1086,146 @@ def _discover_mcp_source(label: str, discover: Callable[[], list[Any]]) -> list[
...
+        _discover_mcp_source_with_progress(
+            "UC functions",
+            "schemas",
+            lambda on_progress: discover_uc_functions_mcp_servers(
+                workspace, profile, on_progress=on_progress
+            ),
+        )
+        if "uc-functions" in sources
+        else []
+    )
```
sunishsheth2009, line 1238: Should we deprecate Vector Search and UC Functions from here to stop promoting it?

sunishsheth2009 (dismissed):

src/ucode/databricks.py
nachill-databricks: This only affects the interactive picker. The `--mcp` flag takes exact service names, so it skips the workspace-wide scan entirely and this budget never applies to it. So it's already fine for both.

src/ucode/databricks.py
nachill-databricks: Most MCP services actually live outside `system.ai` for our users, so defaulting to `system.ai`-only would hide the majority of them — that's why the full scan stays the default. I've bumped the scan budget to 30s so it has time to enumerate them on larger workspaces. The long-list problem is real, but I'd rather solve it with better search/filtering than by narrowing the default scope.

src/ucode/databricks.py
nachill-databricks: The reason isn't always `None` — it's set when the list comes back empty (`"no UC catalogs found"`, `"deadline exceeded…"`), and a couple of tests check those strings. It also matches the `tuple[list, str | None]` shape of the other `list_*`/`discover_*` functions here. The caller ignores it today, but I'd keep it for consistency and in case we want to surface the reason later. Happy to change it if you'd rather.

src/ucode/mcp.py
nachill-databricks: That's basically what this PR does. Both are now unchecked by default on the search screen (labeled "slower") since they scan the whole workspace — you have to opt in. This stops promoting them without removing them. We can fully remove them later if no one uses them.

sunishsheth2009 (approved):

PR #231, opened 2026-07-22
Install AI Tools once per agent for users who previously configured ucode

lennartkats-db:
Stacked on #230. #230 installs Databricks AI Tools when an agent is configured, but users who already set up ucode before it shipped won't get them, since their agents are configured and configuration won't re-run. This backfills them: a persisted ai_tools_installed marker means any already-configured agent gets a one-time install on its next ucode <agent> launch, then is skipped without shelling out again.

- [ ] target branch of this stacked PR must be set to main before merging

# Tests

Unit tests for the marker (recorded on success, skipped when marked, retried on failure), plus end-to-end: an already-configured user's first ucode claude installs AI Tools and the second launch skips it.

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -20,6 +20,7 @@ from ucode.config_io import ToolSpec
 from ucode.databricks import (
     BEDROCK_PROVIDER_TYPES,
     get_databricks_token,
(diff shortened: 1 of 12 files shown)


lennartkats-db:
Closing this given that we landed on a prompt-based journey for #230

PR #236, opened 2026-07-23
ucode configure skills: MCP-connection mode (add/remove/replace) + status

xsh310:
## Summary

Adds `ucode configure skills` for managing a single Unity Catalog **Skills** MCP connection (`databricks-skill-registry`) whose scope is a set of `catalog.schema` locations, rendered as a multi-schema `?schema=` URL (aligned to the backend route in databricks-eng/universe#2256555). Also adds a top-level **Skills** section to `ucode status`.

Self-contained PR off `main` (a small `_setup_mcp_clients` / `build_mcp_http_entry` refactor that was originally split into a separate PR has been folded in here).

## What's in this PR

- `build_skills_mcp_url(workspace, locations)` — multi-schema `?schema=a&schema=b`; empty list → bare `/ai-gateway/skills/` (utility tools only).
- Single skills connection (`kind:"skills"`, fixed name `databricks-skill-registry`) with `skill_locations` as the source of truth; the URL is derived, never parsed back. Every run rebuilds exactly one skills entry (single-connection invariant); non-skills entries are preserved.
- `ucode configure skills --location a.b,c.d --mcp` — sets the connection's location set to **exactly** the listed schemas, replacing any prior set. Override-only for now: to add or drop a schema, re-pass the full list you want. (`--remove`/`--replace` mutation flags are intentionally out of scope to keep the surface minimal.)
- MCP mode is required for now: a bare `ucode configure skills --location a.b` (no `--mcp`) raises a clear "Download mode is not available yet; pass --mcp for now." error rather than silently accepting a no-op. Download mode is a planned follow-up.
- `alwaysLoad` is set on the skills entry for the Claude client only (rides the existing per-client apply); other clients get the URL only.
- Behavior-preserving prep folded in: `_setup_mcp_clients` extracted from `configure_mcp_command` (shared by the mcp and skills commands), and an `always_load` keyword added to `build_mcp_http_entry` (default off).
- `ucode status` gains a **Skills** section (`Skill MCP Locations` + `Configured` agents); the skills entry is excluded from each agent's per-client MCP-servers line to avoid double-listing.

### Unit tests

Unit tests across `tests/test_databricks.py`, `tests/test_mcp.py`, `tests/test_cli.py`:
- `build_skills_mcp_url` for zero / one / many locations (order preserved).
- `_resolve_skills_mcp_servers` single-entry rebuild, URL derived from `skill_locations`, mcp-service entries preserved, `alwaysLoad` set.
- `configure_skills_mcp_command` flows: set-on-empty (registers connection), `--location` replaces a prior set, multiple locations set in order, mcp-service entries preserved across a set.
- CLI dispatch: `--mcp` sets the location set, comma `--location` parsing, `--mcp` required (download-mode error without it), 3-part `.skill` rejected under `--mcp`, malformed → error, missing `--location` → usage error; `status` Skills-section renderings.

Verified locally with `uv run ruff check .` (clean) and the full `pytest` suite: **915 passed, 39 skipped, 1 failed** — the single failure is a pre-existing, environment-only `test_e2e_user_agent.py` case (the local `claude` binary is Claude Code, not the standalone Anthropic CLI) that also fails on `main`.

### Manual E2E verification (live workspace)

Ran the CLI end to end against a real workspace, `eng-ml-inference.staging` (profile `eng-ml-inference-staging`), with Claude Code and Codex configured as agents. For each step I checked the `~/.ucode/state.json` skills entry, `ucode status`, and the agent registrations via `claude mcp list` / `codex mcp list`. (The runs below were against the earlier add/remove/replace surface; the override semantics are what this PR now ships — the set-and-render, no-clobber, endpoint-reachability, and error-path behaviors they exercised are unchanged, and the override behavior is covered by the updated unit tests. A fresh manual pass on the override-only surface is planned before undrafting.)

| Step | Command | Result |
|---|---|---|
| Set one | `configure skills --location xsh.uc_skill_test --mcp` | `skill_locations: ["xsh.uc_skill_test"]`, url `.../ai-gateway/skills/?schema=xsh.uc_skill_test`, entry named `databricks-skill-registry`, clients `[claude, codex]`; both agents registered. A pre-existing old-format `kind:"skills"` entry was swept into the canonical shape. |
| Set two (multi-schema) | `--location xsh.uc_skill_test,xsh.xsh_oncall_skills --mcp` | Set is `["xsh.uc_skill_test", "xsh.xsh_oncall_skills"]`; URL carries both `&schema=` params in order. |
| Override with one | `--location xsh.uc_skill_test --mcp` | Set becomes exactly `["xsh.uc_skill_test"]`; both agent registrations updated. |
| Coexistence / no-clobber | `configure mcp --location system.ai` with a skills entry present | The 10 added `system.ai` mcp-service entries and the skills entry coexist; the skills entry survives in state, in `ucode status`, and in the Claude registration. |
| Error paths | no `--mcp`; 3-part `a.b.c --mcp`; malformed `foo --mcp` | All exit 1 with the expected messages ("Download mode is not available yet; pass --mcp for now.", "--location entries must be `<catalog>.<schema>`..."); a valid set exits 0. |

**`ucode status` Skills section** rendered as expected throughout, e.g.:

```
Skills
  Skill MCP Locations: xsh.uc_skill_test, xsh.xsh_oncall_skills
  Configured: Claude Code, Codex
```

**Endpoint reachability.** Posting a `tools/list` JSON-RPC body to the URL ucode renders returned real tools with HTTP 200, directly against the plain workspace endpoint (the multi-schema route from databricks-eng/universe#2256555 is already deployed to this staging workspace, so no LiteSwap traffic header was needed):
- Single scope `?schema=xsh.uc_skill_test` → 3 skill tools plus the 6 utility tools (`load_skill`, `get_skill_files`, `list_skills`, `create_skill`, `update_skill`, `delete_skill`).
- Two-scope set → 3 + 5 = 8 skill tools plus the 6 utility tools.
- Bare `/ai-gateway/skills/` → 0 skill tools, the 6 utility tools only.

Note: the deployed handler currently emits full-FQN skill tool names (`skill_<catalog>.<schema>.<leaf>`) for both single and multi-scope requests. This is a gateway-side detail; ucode only renders the connection URL, so it does not affect this PR.

Screenshot 2026-07-23 at 12 49 52 PM

Screenshot 2026-07-23 at 12 46 02 PM

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -764,10 +764,8 @@ def status() -> int:
         print_kv("Skills", "not configured")
     else:
         locations = skills_entry.get("skill_locations") or []
-        print_kv("Connection", str(skills_entry.get("name")))
-        print_kv("Mode", "mcp" if locations else "download")
         print_kv(
-            "Locations",
+            "Skill MCP Locations",
             ", ".join(locations) if locations else "none — utility tools only",
         )
         displays = [
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1322,24 +1322,16 @@ def configure_skills(
             "--mcp", help="Manage the skills MCP connection (required until download mode lands)."
         ),
     ] = False,
-    remove: Annotated[
-        bool, typer.Option("--remove", help="With --mcp: remove the listed schemas from the set.")
-    ] = False,
-    replace: Annotated[
-        bool,
-        typer.Option(
-            "--replace", help="With --mcp: replace the set with exactly the listed schemas."
-        ),
-    ] = False,
 ) -> None:
-    """Configure Databricks Skills for your coding tools."""
+    """Configure Databricks Skills for your coding tools.
+
+    ``--location`` sets the skills MCP connection's scope to exactly the listed
+    schemas, replacing any previous set.
+    """
     try:
         if not mcp:
             raise RuntimeError("Download mode is not available yet; pass --mcp for now.")
-        if remove and replace:
-            raise RuntimeError("Use either --remove or --replace, not both.")
-        mode = "remove" if remove else "replace" if replace else "add"
-        configure_skills_mcp_command(_parse_skill_locations(location), mode=mode)
+        configure_skills_mcp_command(_parse_skill_locations(location))
     except RuntimeError as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 
-from typing import Annotated, NamedTuple
+from typing import Annotated
 
 import typer
 from rich.panel import Panel
@@ -55,7 +55,6 @@ from ucode.mcp import (
     MCP_CLIENTS,
     SKILLS_MCP_KIND,
     configure_mcp_command,
-    configure_skills_command,
     configure_skills_mcp_command,
     purge_cross_workspace_mcp_residue,
     revert_mcp_configs,
@@ -145,47 +144,22 @@ def _parse_agents_option(agents: str) -> list[str]:
     return tools
 
 
-class SkillLocation(NamedTuple):
-    catalog: str
-    schema: str
-    skill: str | None
-
-    @property
-    def schema_ref(self) -> str:
-        return f"{self.catalog}.{self.schema}"
-
-
-def _parse_skill_locations(location: str, *, allow_skill: bool) -> list[SkillLocation]:
-    """Parse a comma-separated `--location` into `SkillLocation`s.
-
-    Each entry is a 2-part `<catalog>.<schema>` or 3-part
-    `<catalog>.<schema>.<skill>` dotted ref with non-empty parts. When
-    ``allow_skill`` is False the 3-part form is rejected. Duplicates are
-    dropped, preserving order.
-    """
-    locations: list[SkillLocation] = []
-    seen: set[SkillLocation] = set()
+def _parse_skill_locations(location: str) -> list[str]:
+    """Parse a comma-separated `--location` into `<catalog>.<schema>` refs,
+    dropping duplicates while preserving order."""
+    locations: list[str] = []
     for raw in location.split(","):
         raw = raw.strip()
         if not raw:
             continue
         parts = raw.split(".")
-        if len(parts) not in (2, 3) or not all(part.strip() for part in parts):
-            raise RuntimeError(
-                f"--location entries must be `<catalog>.<schema>[.<skill>]`, got `{raw}`."
-            )
-        if len(parts) == 3 and not allow_skill:
-            raise RuntimeError(
-                f"`.{parts[2]}` names a single skill, which is download-only and not valid "
-                f"with --mcp; drop it to scope the connection to `{parts[0]}.{parts[1]}`."
-            )
-        parsed = SkillLocation(parts[0], parts[1], parts[2] if len(parts) == 3 else None)
-        if parsed not in seen:
-            seen.add(parsed)
-            locations.append(parsed)
+        if len(parts) != 2 or not all(part.strip() for part in parts):
+            raise RuntimeError(f"--location entries must be `<catalog>.<schema>`, got `{raw}`.")
+        if raw not in locations:
+            locations.append(raw)
     if not locations:
         raise RuntimeError(
-            "No skills provided for --location. Use `<catalog>.<schema>[.<skill>]`, "
+            "No schemas provided for --location. Use `<catalog>.<schema>`, "
             "comma-separated for multiple."
         )
     return locations
@@ -759,21 +733,21 @@ def status() -> int:
         console.print()
 
     print_heading("Skills")
-    skills_entry = next((s for s in mcp_servers if s.get("kind") == SKILLS_MCP_KIND), None)
-    if not skills_entry:
+    skill_mcp_entry = next((s for s in mcp_servers if s.get("kind") == SKILLS_MCP_KIND), None)
+    if not skill_mcp_entry:
         print_kv("Skills", "not configured")
     else:
-        locations = skills_entry.get("skill_locations") or []
+        locations = skill_mcp_entry.get("skill_locations") or []
         print_kv(
             "Skill MCP Locations",
             ", ".join(locations) if locations else "none — utility tools only",
         )
-        displays = [
+        configured_agents = [
             str(MCP_CLIENTS[client]["display"])
-            for client in (skills_entry.get("clients") or [])
+            for client in (skill_mcp_entry.get("clients") or [])
             if client in MCP_CLIENTS
         ]
-        print_kv("Configured", ", ".join(displays) if displays else "none")
+        print_kv("Configured", ", ".join(configured_agents) if configured_agents else "none")
 
     print_heading("Tracing")
     tracing = state.get("tracing") or {}
@@ -1340,53 +1314,32 @@ def configure_mcp(
 def configure_skills(
     location: Annotated[
         str,
-        typer.Option(
-            "--location",
-            help="Comma-separated `<catalog>.<schema>` skill scopes (a single "
-            "`<catalog>.<schema>.<skill>` is download-only). Required.",
-        ),
+        typer.Option("--location", help="Comma-separated `<catalog>.<schema>` skill scopes."),
     ],
     mcp: Annotated[
         bool,
-        typer.Option("--mcp", help="Mutate the skills MCP connection instead of downloading."),
+        typer.Option(
+            "--mcp", help="Manage the skills MCP connection (required until download mode lands)."
+        ),
     ] = False,
     remove: Annotated[
-        bool, typer.Option("--remove", help="(--mcp) Remove the listed schemas from the set.")
+        bool, typer.Option("--remove", help="With --mcp: remove the listed schemas from the set.")
     ] = False,
     replace: Annotated[
         bool,
-        typer.Option("--replace", help="(--mcp) Replace the set with exactly the listed schemas."),
-    ] = False,
-    path: Annotated[
-        str | None, typer.Option("--path", help="(download) Project root; default user scope (~/).")
-    ] = None,
-    yes: Annotated[
-        bool,
-        typer.Option("--yes", "-y", help="(download) Overwrite on collision without prompting."),
+        typer.Option(
+            "--replace", help="With --mcp: replace the set with exactly the listed schemas."
+        ),
     ] = False,
 ) -> None:
     """Configure Databricks Skills for your coding tools."""
     try:
+        if not mcp:
+            raise RuntimeError("Download mode is not available yet; pass --mcp for now.")
         if remove and replace:
             raise RuntimeError("Use either --remove or --replace, not both.")
-        if (remove or replace) and not mcp:
-            raise RuntimeError("--remove/--replace require --mcp.")
-        if path is not None and mcp:
-            raise RuntimeError("--path is not valid with --mcp.")
-        if mcp:
-            mode = "remove" if remove else "replace" if replace else "add"
-            locs = _parse_skill_locations(location, allow_skill=False)
-            configure_skills_mcp_command([loc.schema_ref for loc in locs], mode=mode)
-        else:
-            # Download mode lands in a later PR; until then, reject every
-            # download-only affordance rather than silently ignoring it.
-            locs = _parse_skill_locations(location, allow_skill=True)
-            if path is not None or yes or any(loc.skill for loc in locs):
-                raise RuntimeError(
-                    "Download mode is not available yet; use --mcp to configure the skills "
-                    "connection for the given schemas."
-                )
-            configure_skills_command([loc.schema_ref for loc in locs])
+        mode = "remove" if remove else "replace" if replace else "add"
+        configure_skills_mcp_command(_parse_skill_locations(location), mode=mode)
     except RuntimeError as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -773,6 +773,10 @@ def status() -> int:
     print_note(
         "Use `ucode configure mcp` to add Databricks MCP servers to configured coding tools."
     )
+    print_note(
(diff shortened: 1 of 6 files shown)


src/ucode/cli.py
```diff
@@ -142,6 +145,52 @@ def _parse_agents_option(agents: str) -> list[str]:
...
+
+    @property
+    def schema_ref(self) -> str:
+        return f"{self.catalog}.{self.schema}"
+
+
+def _parse_skill_locations(location: str, *, allow_skill: bool) -> list[SkillLocation]:
+    """Parse a comma-separated `--location` into `SkillLocation`s.
+
+    Each entry is a 2-part `<catalog>.<schema>` or 3-part
```
xsh310: Let's add the 3-part handling in the later download mode PR instead

src/ucode/cli.py
```diff
@@ -1268,6 +1336,65 @@ def configure_mcp(
...
+        bool, typer.Option("--remove", help="(--mcp) Remove the listed schemas from the set.")
+    ] = False,
+    replace: Annotated[
+        bool,
+        typer.Option("--replace", help="(--mcp) Replace the set with exactly the listed schemas."),
+    ] = False,
+    path: Annotated[
+        str | None, typer.Option("--path", help="(download) Project root; default user scope (~/).")
+    ] = None,
+    yes: Annotated[
```
xsh310: Let's not add this in later download mode PR

src/ucode/cli.py
```diff
@@ -1268,6 +1336,65 @@ def configure_mcp(
...
+        str,
+        typer.Option(
+            "--location",
+            help="Comma-separated `<catalog>.<schema>` skill scopes (a single "
+            "`<catalog>.<schema>.<skill>` is download-only). Required.",
+        ),
+    ],
+    mcp: Annotated[
+        bool,
+        typer.Option("--mcp", help="Mutate the skills MCP connection instead of downloading."),
```
xsh310: Can you make this required and change this to optional when we add support for download?

src/ucode/cli.py
```diff
@@ -1268,6 +1336,65 @@ def configure_mcp(
...
+    ] = False,
+    remove: Annotated[
+        bool, typer.Option("--remove", help="(--mcp) Remove the listed schemas from the set.")
+    ] = False,
+    replace: Annotated[
+        bool,
+        typer.Option("--replace", help="(--mcp) Replace the set with exactly the listed schemas."),
+    ] = False,
```
xsh310: Should we note that remove and replace is currently only supported for mcp mode?

src/ucode/cli.py
```diff
@@ -1268,6 +1336,65 @@ def configure_mcp(
...
+        if (remove or replace) and not mcp:
+            raise RuntimeError("--remove/--replace require --mcp.")
+        if path is not None and mcp:
+            raise RuntimeError("--path is not valid with --mcp.")
+        if mcp:
+            mode = "remove" if remove else "replace" if replace else "add"
+            locs = _parse_skill_locations(location, allow_skill=False)
+            configure_skills_mcp_command([loc.schema_ref for loc in locs], mode=mode)
+        else:
+            # Download mode lands in a later PR; until then, reject every
```
xsh310: Should we just return not implemented error here for now?

Also is `configure_skills_command([loc.schema_ref for loc in locs])` correct? I thought we only provide static tools when in download mode

src/ucode/cli.py
```diff
@@ -707,6 +758,23 @@ def status() -> int:
...
+    skills_entry = next((s for s in mcp_servers if s.get("kind") == SKILLS_MCP_KIND), None)
+    if not skills_entry:
+        print_kv("Skills", "not configured")
+    else:
+        locations = skills_entry.get("skill_locations") or []
+        print_kv(
+            "Skill MCP Locations",
+            ", ".join(locations) if locations else "none — utility tools only",
+        )
+        displays = [
```
xsh310: nit: why this is variable called displays? Is there a more intuitive variable name for this?

src/ucode/cli.py
```diff
@@ -707,6 +758,23 @@ def status() -> int:
         print_kv("Config file", str(config_path) if config_path.exists() else "missing")
         console.print()
 
+    print_heading("Skills")
+    skills_entry = next((s for s in mcp_servers if s.get("kind") == SKILLS_MCP_KIND), None)
```
xsh310: Can we call this skill_mcp_entry

src/ucode/databricks.py
```diff
@@ -1329,6 +1329,18 @@ def build_mcp_service_url(workspace: str, full_name: str) -> str:
     return f"{workspace}/ai-gateway/mcp-services/{full_name}"
 
 
+def build_skills_mcp_url(workspace: str, locations: list[str]) -> str:
+    """Skills route with repeated ``?schema=`` scopes (universe PR #2256555).
```
xsh310: Tighten this cmd to only provide 2 examples, 1 for empty location, 1 for multiple location. Don't mention universe PR number

src/ucode/mcp.py
```diff
@@ -79,6 +80,9 @@
         "list_command": "copilot mcp list",
     },
 }
+SKILLS_MCP_KIND = "skills"
+SKILLS_MCP_SERVER_NAME = "databricks-skill-registry"
+SKILLS_MCP_SCHEMA_SOFT_CAP = 10
```
xsh310: I think it's fine that we drop soft cap for now and have the backend return the errors directly

src/ucode/mcp.py
```diff
@@ -1037,7 +1044,10 @@ def apply_mcp_server_changes(
         url = server.get("url")
         if not isinstance(url, str) or not url:
             continue
-        entry = build_mcp_http_entry(url)
+        # alwaysLoad is Claude-only and rides the per-entry apply; other clients
```
xsh310: Tighten this comment

src/ucode/mcp.py
```diff
@@ -1103,6 +1113,12 @@ def purge_cross_workspace_mcp_residue(state: dict, workspace: str) -> None:
         )
 
 
+def _skills_entries(servers: list[dict]) -> list[dict]:
+    """Skills connections carry their own lifecycle (``configure skills``); the
```
xsh310: remove this docstring, the function name self explantory

src/ucode/mcp.py
```diff
@@ -1335,3 +1438,40 @@ def configure_mcp_command(location: str | None = None, services: set[str] | None
         # User submitted the picker without toggling anything --> make it clear nothing was selected
         print_note("No MCP servers selected. Press space to toggle an item, then enter to save.")
     return 0
+
+
+def _apply_skills_connection(
```
xsh310: nit: does it make sense to call this _update_skills_mcp instead?

src/ucode/mcp.py
```diff
@@ -1335,3 +1438,40 @@ def configure_mcp_command(location: str | None = None, services: set[str] | None
...
+    if len(new_locations) > SKILLS_MCP_SCHEMA_SOFT_CAP:
+        print_warning(
+            f"{len(new_locations)} schemas exceeds the recommended limit of "
+            f"{SKILLS_MCP_SCHEMA_SOFT_CAP}; registering anyway."
+        )
+    _apply_skills_connection(state, workspace, clients, new_locations)
+    return 0
+
+
+def configure_skills_command(locations: list[str]) -> int:
```
xsh310: Why do we need this?

src/ucode/mcp.py
```diff
@@ -1174,28 +1191,87 @@ def _resolve_location_mcp_servers(
...
+
+
+def resolve_skill_location_set(current: list[str], requested: list[str], *, mode: str) -> list[str]:
+    """Apply a set mutation to the skill-location list. ``mode`` is one of
+    ``add`` (union), ``remove`` (difference), ``replace`` (deduped requested).
+    Order is stable throughout; backend assigns tool indices by URL position."""
+    if mode == "remove":
+        return [c for c in current if c not in requested]
+    if mode == "replace":
+        return _dedupe_stable(requested)
```
xsh310: hmm, is there not a built-in way to get the distinct value in a python list?

src/ucode/mcp.py
```diff
@@ -1174,28 +1191,87 @@ def _resolve_location_mcp_servers(
...
+    return current + [r for r in requested if r not in current]
+
+
+def _current_skill_locations(state: dict) -> list[str]:
+    """Read ``skill_locations`` off the single ``kind:"skills"`` entry, else ``[]``."""
+    for server in state.get("mcp_servers") or []:
+        if server.get("kind") == SKILLS_MCP_KIND:
+            return list(server.get("skill_locations") or [])
+    return []
```
xsh310: Can we move these skill helper functions to be right above configure_skills_mcp_command?

src/ucode/cli.py
xsh310: Done (ba1f1b6) — dropped 3-part `.skill` parsing entirely; `_parse_skill_locations` now takes only `<catalog>.<schema>`. It'll come back with download mode.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — removed `--yes` (and `--path`); both move to the download-mode PR.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — `--mcp` is now effectively required: a non-`--mcp` invocation errors with "Download mode is not available yet; pass --mcp for now." When download lands it becomes the default and `--mcp` turns optional.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — help text now reads "With --mcp: …" for both --remove and --replace.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — non-`--mcp` now returns the not-implemented error. And you're right about static tools: I removed the bare 'register schemas' path (`configure_skills_command`) entirely, so there's no non-`--mcp` behavior to get wrong.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — renamed to `configured_agents`.

src/ucode/cli.py
xsh310: Done (ba1f1b6) — renamed to `skill_mcp_entry`.

src/ucode/databricks.py
xsh310: Done (ba1f1b6) — docstring now shows just two examples (empty + multi-location) and drops the PR number.

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — removed `SKILLS_MCP_SCHEMA_SOFT_CAP` and the warning; the backend enforces the limit.

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — tightened to one line.

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — docstring removed.

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — renamed to `_update_skills_mcp`.

src/ucode/mcp.py
xsh310: Removed (ba1f1b6) — this was the bare-default 'register schemas' entry point; with download deferred and the non-`--mcp` path erroring, it's no longer needed.

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — replaced `_dedupe_stable` with `list(dict.fromkeys(requested))` (order-preserving dedupe built-in).

src/ucode/mcp.py
xsh310: Done (ba1f1b6) — moved `_current_skill_locations`, `resolve_skill_location_set`, and `_update_skills_mcp` to just above `configure_skills_mcp_command`.

src/ucode/mcp.py
```diff
@@ -1335,3 +1405,46 @@ def configure_mcp_command(location: str | None = None, services: set[str] | None
...
+
+
+def resolve_skill_location_set(current: list[str], requested: list[str], *, mode: str) -> list[str]:
+    """Apply a set mutation to the skill-location list. ``mode`` is one of
+    ``add`` (union), ``remove`` (difference), ``replace`` (deduped requested).
+    Order is stable throughout; backend assigns tool indices by URL position."""
+    if mode == "remove":
+        return [c for c in current if c not in requested]
+    if mode == "replace":
+        return list(dict.fromkeys(requested))
```
xsh310: Do we still need this dedup step given that we've dropped dup in _parse_skill_locations?

src/ucode/mcp.py
```diff
@@ -1174,28 +1187,58 @@ def _resolve_location_mcp_servers(
...
-                f"`system.ai.<name>`): {', '.join(bare)}"
-            )
-        if len(schemas) != 1:
-            raise RuntimeError(
-                "--services without --location must all share one `<catalog>.<schema>` "
-                f"(got: {', '.join(sorted(schemas)) or 'none'}); pass --location instead."
-            )
-        location = next(iter(schemas))
-    state = load_state()
+def _merge_clients(prior: list[str] | None, new: list[str]) -> list[str]:
```
xsh310: Can we move _merge_clients, _build_skills_entry, and _resolve_skills_mcp_servers to be on top of configure_skills_mcp_command as well?

src/ucode/mcp.py
xsh310: Good catch — dropped it (f7628f4). `resolve_skill_location_set` now returns `requested` as-is for replace; `_parse_skill_locations` is the single place that de-dupes. Updated the test accordingly.

src/ucode/mcp.py
xsh310: Done (f7628f4) — moved `_merge_clients`, `_build_skills_entry`, and `_resolve_skills_mcp_servers` down next to the other skills helpers, just above `configure_skills_mcp_command`. (`_skills_entries` stays up top since it's used by `configure_mcp_command`, not the skills path.)

AarushiShah-db (approved):

PR #237, opened 2026-07-24
skills: add download mode to `ucode configure skills` (client, writer, CLI wiring)

xsh310:
## What

Completes **download mode** for `ucode configure skills`, end to end. Downloads every Unity Catalog skill in a schema to local disk and registers a schema-less skills MCP connection. This PR covers the whole path: the read-only client, the on-disk writer, the CLI wiring, and docs.

> **Stacked on #236.** This branch is rebased on top of `skills-mcp-mode` (PR #236, MCP mode). Because that branch lives in a fork, GitHub cannot set it as this PR's base, so the base stays `main` and the diff shown here includes #236's commits until it merges. Review only the download commits (`b2ffa9a` onward); the first six commits are #236. Once #236 merges into `main`, this diff self-cleans.

**Download client + writer + orchestration** (`src/ucode/skills_download.py`, new module):

- **`list_schema_skills(...)`** returns finalized skill leaf names in a schema via the UC skills API (`GET /api/2.1/unity-catalog/skills?parent=schemas/<cat>.<sch>`, paginated). Uses `bundle_name` as the leaf and the presence of `finalize_time` as the readiness signal.
- **`list_skill_files(...)`** returns a bundle's files as paths relative to the skill directory, by recursively walking its UC Volume directory (`GET /api/2.0/fs/directories/Volumes/<cat>/<sch>/<leaf>`, paginated).
- **`fetch_skill_file(...)`** returns one file's raw bytes; **`fetch_skill_bundle(...)`** composes list + fetch into `{relpath: bytes}`, all or nothing (any failure returns a reason and no bundle, so a partial skill is never written).
- **`skill_dir_roots(path)`** returns the `.claude/skills` + `.agents/skills` roots. `path` is optional: given, it must be an existing absolute directory; omitted, roots default to the home directory (user scope).
- **`write_skill(...)`** writes a skill's `{relpath: bytes}` bundle into every root, one flat `<root>/<leaf>/` dir per skill. It prompts before overwriting any pre-existing skill dir (ucode does not track which schema wrote a dir, so it always asks). Rejects invalid leaf names and `..`/absolute paths in the bundle.
- **`download_skills(...)`** fetches a schema's bundles concurrently (with a `k/n` progress bar), then writes them sequentially so overwrite prompts do not interleave. A per-skill failure warns and skips without aborting the batch, and it prints a `Downloaded k/n skill(s)` summary.
- **`configure_skills_download_command(locations, *, path)`** is the command entry point: resolves clients + token, calls `download_skills`, then registers or keeps the schema-less MCP connection.

**Shared transport** (`src/ucode/databricks.py`): `_http_get_bytes`, a raw-bytes GET helper next to `_http_get_json`, since bundles can contain binary files.

**MCP glue** (`src/ucode/mcp.py`): `setup_mcp_clients` (promoted from private) and `register_schemaless_skills_connection`, which preserves any prior `--mcp` `skill_locations` and otherwise registers the bare schema-less route. `mcp.py` holds no download logic; the dependency flows one way (`cli` to `skills_download` to `mcp` to `databricks`).

**CLI** (`src/ucode/cli.py`): `--mcp` is optional and `--path` is download-only. Download is the default mode; `--mcp` opts into the connection-only mode; `--path` with `--mcp` is rejected. `skill_locations` is never touched by download, so a prior `--mcp` scope survives a download run.

**Docs** (`README.md`, `AGENTS.md`): a Skills subsection and Other Commands rows covering both modes, the optional `--path` with its home-dir default, the flat layout, the collision prompt, and the schema-less connection; module-boundary note for `skills_download.py`.

## Test plan

**Unit tests** (no network, mocked HTTP / patched `prompt_yes_no`):

- `tests/test_skills_download.py`: the client (`list_schema_skills` finalized-only filter + pagination, `list_skill_files` recursive walk + pagination, `fetch_skill_file`, `fetch_skill_bundle` all-or-nothing); the writer (`skill_dir_roots` project + home-default + reject relative/missing, `write_skill` into both roots, nested files, collision prompt keep/overwrite, invalid-leaf skip, `..`/absolute path rejection); the orchestrator (`download_skills` fetch-and-write, list-failure skip, per-skill bundle-failure skip, `k/n` summary counts only written skills); and the command wiring (`configure_skills_download_command` downloads then registers, threads a `None` path).
- `tests/test_mcp.py`: `_skill_mcp_locations`, `register_schemaless_skills_connection` (registers the bare route, preserves a prior `--mcp` set), plus revert/purge lifecycle coverage for the skills entry.
- `tests/test_cli.py`: two-mode dispatch (default downloads with and without `--path`, `--mcp` sets the connection, `--path` + `--mcp` rejected, malformed `--location`).

Full suite: **952 passing**, 39 skipped, 1 pre-existing unrelated `test_e2e_user_agent` failure (spawns a live `claude` subprocess; also fails on the base commit). `ruff check`, `ruff format --check`, and `ty check src/` all clean (these are the CI gates in `test_lint.py`).

**Live end-to-end against a staging workspace** (`eng-ml-inference.staging`, user OAuth token, location `xsh.uc_skill_test`). Drove the real `configure_skills_download_command` with only the agent-config writes and state persistence patched out, so the network fetch, the write, and the connection registration all ran for real. Three scenarios on the latest commit:

1. **Project scope** (`--path <dir>`, fresh): `pii-handling` downloaded with its full nested bundle (`SKILL.md` + 5 `references/*.md` + 2 `templates/*`) into both `.claude/skills/` and `.agents/skills/` (16 files total); `SKILL.md` had real frontmatter, not a stub. The progress bar advanced `0/3` to `3/3` during the concurrent fetch and cleared. Summary: `Downloaded 1/3 skill(s) from xsh.uc_skill_test`. Schema-less connection registered (URL `.../ai-gateway/skills/`, `skill_locations: []`, clients `[claude, codex]`).
2. **Project scope, re-run**: prompted `A skill named 'pii-handling' already exists. Overwrite it with 'xsh.uc_skill_test.pii-handling'?` before writing, confirming the collision path.
3. **Home scope** (`--path` omitted, `Path.home()` pointed at a temp dir): the same 16 files landed under the home-dir roots, confirming the optional-`--path` default.

In all three, the two other listed leaves (`xsh-format`, `dataset-boundary`) have no Volume in staging and returned a real `HTTP 404 Volume ... does not exist`; each was skipped with a warning and the batch continued (exit 0). Scratch dirs were removed afterward and real ucode state was never written.

The earlier client-only e2e also resolved a real design question: ucode's client-side OAuth token reads all three endpoints directly, so no `GenerateTemporaryStorageCredentials` vending is needed. It corrected two assumptions unit tests alone would miss: the deployed endpoint is `/api/2.1/unity-catalog/skills` (not the `managed-catalog-aigov` path), and readiness is `finalize_time` (there is no `status` field).

Screenshot 2026-07-24 at 1 52 13 PM
Screenshot 2026-07-24 at 1 52 27 PM
Screenshot 2026-07-24 at 1 53 24 PM

## Notes

- Draft until reviewed.


src/ucode/skills_download.py
```diff
@@ -0,0 +1,101 @@
+"""Write downloaded Unity Catalog skills to disk for coding agents to load.
```
xsh310: Tighten

src/ucode/skills_download.py
```diff
@@ -0,0 +1,101 @@
...
+import re
+from pathlib import Path
+
+from ucode.ui import print_warning, prompt_yes_no
+
+# Cross-agent skill directories: `.claude/skills` (Claude) and `.agents/skills`
+# (the Agent Skills alias the other agents read). Both get the same skills.
+SKILL_DIR_NAMES = (".claude/skills", ".agents/skills")
+
+# Agent Skills spec: a skill's directory name is its `name`, lowercase a-z 0-9 -.
```
xsh310: remove this and name this SKILL_NAME_PATTERN

src/ucode/skills_download.py
```diff
@@ -0,0 +1,101 @@
...
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from ucode.ui import print_warning, prompt_yes_no
+
+# Cross-agent skill directories: `.claude/skills` (Claude) and `.agents/skills`
+# (the Agent Skills alias the other agents read). Both get the same skills.
+SKILL_DIR_NAMES = (".claude/skills", ".agents/skills")
```
xsh310: Rename SKILL_BASE_DIR_NAMES

src/ucode/skills_download.py
```diff
@@ -0,0 +1,87 @@
...
+        safe_path = _safe_relative_path(relative_path)
+        if safe_path is None:
+            print_warning(f"Skipping unsafe path in `{leaf}`: {relative_path}")
+            continue
+        destination = skill_dir / safe_path
+        destination.parent.mkdir(parents=True, exist_ok=True)
+        destination.write_bytes(content)
+
+
+def write_skill(roots: list[Path], leaf: str, files: dict[str, bytes], *, schema_ref: str) -> str:
```
xsh310: Is schema_ref essentially the location string? Can we name it location if that's the case?

src/ucode/skills_download.py
```diff
@@ -0,0 +1,87 @@
...
+    """Write one skill's bundle into every root, prompting before overwriting.
```
xsh310: Tight this docstring

src/ucode/skills_download.py
xsh310: Done in 50c6f9c — condensed the module docstring to one line, and tightened `write_skill`'s docstring too.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/skills_download.py
xsh310: Done in 50c6f9c — removed the comment and renamed `_LEAF_PATTERN` → `SKILL_NAME_PATTERN`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/skills_download.py
xsh310: Done in 50c6f9c — renamed `SKILL_DIR_NAMES` → `SKILL_BASE_DIR_NAMES`.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/skills_download.py
xsh310: Yes — it's the `<catalog>.<schema>` location string the download loop iterates. Renamed the param to `location` in 50c6f9c.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/skills_download.py
xsh310: Done in 50c6f9c — tightened to a short docstring covering the prompt-before-overwrite behavior and the `location` arg.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/cli.py
```diff
@@ -1349,6 +1315,44 @@ def configure_mcp(
...
+    ``--path`` and registers a schema-less MCP connection. With ``--mcp``,
+    instead sets the skills MCP connection's scope to exactly the listed schemas.
+    """
+    try:
+        if mcp:
+            if path is not None:
+                raise RuntimeError("--path is not valid with --mcp.")
+            configure_skills_mcp_command(_parse_skill_locations(location))
+        else:
+            if path is None:
```
xsh310: Why is path required for download? It uses user scope skill dir when not provided according to the spec.

src/ucode/databricks.py
```diff
@@ -1365,6 +1358,136 @@ def build_mcp_service_url(workspace: str, full_name: str) -> str:
...
+        []                        -> ``.../ai-gateway/skills/``
+        ["main.default", "ml.a"]  -> ``.../ai-gateway/skills/?schema=main.default&schema=ml.a``
+    """
+    base = f"{workspace}/ai-gateway/skills/"
+    if not locations:
+        return base
+    return base + "?" + urlencode([("schema", loc) for loc in locations])
+
+
+# --- Skill download (UC skills API + Files API) ----------------------------
```
xsh310: Given that we have a src/ucode/skills_download.py? Would it be possible to move all of these helper skill download helper function to that file?

src/ucode/mcp.py
```diff
@@ -1335,3 +1365,104 @@ def configure_mcp_command(location: str | None = None, services: set[str] | None
...
+    return 0
+
+
+def _skill_mcp_locations(state: dict) -> list[str]:
+    """The skills MCP connection's ``skill_locations``, or ``[]`` if none exists."""
+    entry = next(iter(_skills_entries(list(state.get("mcp_servers") or []))), None)
+    return list((entry or {}).get("skill_locations") or [])
+
+
+def configure_skills_download_command(locations: list[str], *, path: str) -> int:
```
xsh310: I don't think the download logic should live in the mcp.py? Can we only leave the mcp related logic here and move the download related logic to a skill.py or maybe skill_download.py directly?

src/ucode/mcp.py
```diff
@@ -1335,3 +1365,104 @@ def configure_mcp_command(location: str | None = None, services: set[str] | None
...
+    ``<path>/.agents/skills`` (prompting before overwriting an existing dir), then
+    registers/keeps the schema-less MCP connection. ``skill_locations`` is never
+    touched, so a prior ``--mcp`` set survives a download run.
+    """
+    state = load_state()
+    workspace, profile, clients = _setup_mcp_clients(state, "Skills")
+    token = get_databricks_token(workspace, profile)
+    roots = skill_dir_roots(path)
+
+    for location in locations:
```
xsh310: 2 main feedback on the download loops:
1. would be possible to parallelize a bit? eg. maybe parallelize the download. Maybe write is a bit more difficult given that there could be conflicts and would require interactions
2. can we add some progress tracking messages to be printed so that users can know that skills are being downloaded? We can keep this relatively simple for now

src/ucode/skills_download.py
```diff
@@ -0,0 +1,76 @@
...
+        destination.parent.mkdir(parents=True, exist_ok=True)
+        destination.write_bytes(content)
+
+
+def write_skill(roots: list[Path], leaf: str, files: dict[str, bytes], *, location: str) -> str:
+    """Write ``leaf``'s bundle (``{relpath: bytes}``) into every root.
+
+    Prompts before overwriting an existing skill dir. ``location`` is the source
+    ``<catalog>.<schema>``, shown in that prompt. Returns ``"written"``,
+    ``"overwritten"``, ``"kept"``, or ``"skipped"``.
```
xsh310: Are these output string used anywhere? Do we need them?

src/ucode/cli.py
xsh310: Good catch, that contradicted the spec. Made `--path` optional in 67bb1c6: when omitted, `skill_dir_roots(None)` defaults the roots to the home dir (`~/.claude/skills` + `~/.agents/skills`, user scope). Updated the CLI help, README, AGENTS, the spec, and tests to match.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/databricks.py
xsh310: Done in 67bb1c6. Moved the whole download client (`list_schema_skills`, `list_skill_files`, `fetch_skill_file`, `fetch_skill_bundle`, `_skill_bundle_name`) into `skills_download.py`. `databricks.py` now keeps only the shared transport helpers (`_http_get_json`, `_http_get_bytes`), which `skills_download` imports. The client tests moved to `test_skills_download.py` too.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/mcp.py
xsh310: Agreed. Moved the download orchestration into `skills_download.download_skills` in 67bb1c6. `mcp.py`'s `configure_skills_download_command` now only does the MCP glue: set up clients, get the token, call `download_skills`, then register/keep the schema-less connection.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/mcp.py
xsh310: Both done in 67bb1c6:
1. Per-skill bundle fetches now run concurrently via a `ThreadPoolExecutor` (`_fetch_bundles`, up to 8 workers). Writes stay sequential as you suggested, since the overwrite prompt needs to interact with the user one skill at a time.
2. Added progress messages: `Downloading N skill(s) from <location>...` before a schema, then per skill `Downloaded <leaf> (N file(s))` / `Overwrote ...` / `Kept existing ...`, plus a `No skills found` note for an empty schema.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/skills_download.py
xsh310: No, only tests read them. Dropped the return value in 67bb1c6; `write_skill` now returns None and prints its own outcome (`Downloaded` / `Overwrote` / `Kept existing`), which also covers the progress-message ask. Updated the tests to assert on the files written rather than the status string.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

src/ucode/mcp.py
xsh310: Follow-up: my earlier reply only moved the download *loop*; the `configure_skills_download_command` wrapper was still in `mcp.py`. Fixed in a6f5535 — the command now lives in `skills_download.py` too. `mcp.py` keeps only the MCP-connection glue it exposes as `setup_mcp_clients` and `register_schemaless_skills_connection`, and no longer imports `skills_download`, so the dependency flows one way (`cli -> skills_download -> mcp -> databricks`).

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

AarushiShah-db (approved):

PR #238, opened 2026-07-24
fix: consolidate Pi and OpenCode model routing

dgokeeffe:
## Summary

Consolidates the Pi and OpenCode model fixes into one self-contained change.

## Changes

- Add Pi support for MLflow chat-completions-only OSS models.
- Add OSS reasoning, context-window, and output-token metadata.
- Harden the Pi MLflow SSE repair proxy.
- Prevent gpt-oss models from being routed through the Codex Responses API.
- Add long-context Claude and GPT metadata for Pi.
- Restore OpenCode access to GPT-5, GPT-5.6, and Codex through `databricks-openai`.
- Fall back to foundation-model serving endpoints when UC model-services does not expose OSS models.

## Validation

- Focused suite: 362 passed.
- `git diff --check` clean.

This is intentionally self-contained and does not depend on the blocked/stale OpenCode PR #101 or the stacked PR #223.

PR #244, opened 2026-07-26
fix(pi): extend forceAdaptiveThinking to Claude 5+ models

dgokeeffe:
## Problem

Claude Opus 5 (e.g. `databricks-claude-opus-5-0`) rejects the legacy `thinking.type.enabled` field and requires `thinking.type.adaptive`:

```
Error: 400 {"message":"\"thinking.type.enabled\" is not supported for this model. Use \"thinking.type.adaptive\" and \"output_config.effort\" to control thinking behavior."}
```

Pi's `forceAdaptiveThinking` compat flag tells Pi to send `thinking.type.adaptive` instead. The regex that gates this flag only matched Claude **4.6+**, so Opus 5 was missing it and 400ed at the gateway.

## Fix

Extend the regex in `_pi_claude_model_entry` to also match **any major version ≥ 5** (with any minor version), while keeping the existing 4.6+ behavior unchanged:

```python
r"claude-(?:opus|sonnet)-(?:[5-9]|[1-9]\d)-\d"
r"|claude-(?:opus|sonnet)-4-(?:[6-9]|[1-9]\d)"
```

This correctly:
- ✅ matches `claude-opus-5-0`, `claude-sonnet-5-0`, and future majors ≥ 5
- ✅ keeps matching `claude-opus-4-6` through `claude-opus-4-99` (4.6+)
- ✅ does **not** match `claude-sonnet-4-5` (predates the adaptive thinking API)

## Tests

- `test_claude_opus_5_gets_force_adaptive_thinking` — verifies opus-5 and sonnet-5 both get `forceAdaptiveThinking: True` with correct 1M/128k token limits
- `test_claude_sonnet_4_5_does_not_get_force_adaptive_thinking` — regression guard that sonnet 4.5 does **not** get the flag

All 53 tests in `test_agent_pi.py` pass. ruff format clean.

## Validation

```
.venv/bin/python -m pytest tests/test_agent_pi.py -q  # 53 passed
.venv/bin/ruff format --check src/ucode/agents/pi.py tests/test_agent_pi.py  # clean
```

> **Note:** This branch is based on `bugfix/consolidated-pi-opencode-upstream` (PR #238, closed) which carries the `forceAdaptiveThinking` infrastructure. The fix is a single regex change + two tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "ucode"
-version = "0.1.0"
+version = "0.2.0"
 description = "Bootstrap Codex against a Databricks workspace with minimal setup."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "ucode"
-version = "0.2.0"
+version = "0.3.0"
 description = "Bootstrap Codex against a Databricks workspace with minimal setup."
 readme = "README.md"
 requires-python = ">=3.12"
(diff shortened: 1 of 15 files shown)


dgokeeffe:
Closing in favor of #245 — the fix has been folded into the consolidated branch () which now includes the Claude 5+ adaptive thinking fix alongside the other Pi/OpenCode model routing changes.

PR #245, opened 2026-07-26
fix: consolidate Pi and OpenCode model routing (+ Claude 5 adaptive thinking)

dgokeeffe:
## Summary

Consolidates the Pi and OpenCode model fixes into one self-contained change, now including the Claude Opus 5 adaptive thinking fix.

## Changes

- **fix(pi): extend `forceAdaptiveThinking` to Claude 5+ models** — Opus 5 / Sonnet 5 reject `thinking.type.enabled` and require `thinking.type.adaptive`. The regex now matches any major version ≥ 5, not just 4.6+.
- Add Pi support for MLflow chat-completions-only OSS models.
- Add OSS reasoning, context-window, and output-token metadata.
- Harden the Pi MLflow SSE repair proxy.
- Prevent gpt-oss models from being routed through the Codex Responses API.
- Add long-context Claude and GPT metadata for Pi.
- Restore OpenCode access to GPT-5, GPT-5.6, and Codex through `databricks-openai`.
- Fall back to foundation-model serving endpoints when UC model-services does not expose OSS models.

## Validation

- `pytest tests/test_agent_pi.py -q` — 53 passed
- `ruff format --check` clean

This is a replacement for the closed PR #238 (GitHub would not allow reopening after the branch was updated).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "ucode"
-version = "0.1.0"
+version = "0.2.0"
 description = "Bootstrap Codex against a Databricks workspace with minimal setup."
 readme = "README.md"
 requires-python = ">=3.12"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "ucode"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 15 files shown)


dgokeeffe:
Closing — the consolidated branch is for local use only. The opus-5 fix will be applied to the individual PR #223 (fix/1m-context-claude-gpt-pi) where the force_adaptive_thinking logic actually lives.

PR #250, opened 2026-07-28
detect conflicting env keys set by enterprise settings

asujithan:
Adds warning message when default models are set in the managed/enterprise settings which would override what's in ucode-settings. Blocking error is thrown if ANTHROPIC_BASE_URL or ANTHROPIC_CUSTOM_HEADERS is set in the managed settings since that will prevent claude terminal from being started with Claude Max/Enterprise subscription.

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -129,17 +129,56 @@ def _managed_settings_path() -> Path | None:
     return None
 
 
-def _managed_api_key_helper_path() -> Path | None:
-    """Path to the enterprise managed-settings.json when it sets an apiKeyHelper,
-    else None. The managed scope always wins over --settings and subscription
-    OAuth, so a managed apiKeyHelper silently shadows relayed auth."""
+def _managed_relayed_conflicts() -> tuple[Path, list[str]] | None:
+    """Enterprise managed-settings keys that would break relayed auth, if any.
+    The managed scope always wins (per key) over the --settings file and
+    subscription OAuth, so a managed value here overrides what ucode writes:
+    'apiKeyHelper' shadows the subscription login, 'env.ANTHROPIC_BASE_URL'
+    clobbers our loopback proxy URL, and 'env.ANTHROPIC_CUSTOM_HEADERS' drops
+    the Databricks-Model-Provider-Service routing headers — each sends traffic
+    somewhere the relayed token swap can't reach or route correctly.
+    Returns (path, conflicting-key-labels) or None when there's no conflict."""
     path = _managed_settings_path()
     if path is None or not path.is_file():
         return None
     settings = read_json_safe(path)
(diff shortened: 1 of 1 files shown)


src/ucode/agents/claude.py
```diff
@@ -129,17 +129,56 @@ def _managed_settings_path() -> Path | None:
...
+# Default-model env keys the enterprise managed settings can pin. Unlike the
+# conflict keys above these don't break relayed auth, but a pinned (possibly
+# deprecated) model id silently overrides ucode's picker, so we surface them.
+_MANAGED_DEFAULT_MODEL_ENV_KEYS = (
+    "ANTHROPIC_MODEL",
+    "ANTHROPIC_DEFAULT_OPUS_MODEL",
+    "ANTHROPIC_DEFAULT_SONNET_MODEL",
+    "ANTHROPIC_DEFAULT_HAIKU_MODEL",
+    "ANTHROPIC_DEFAULT_FABLE_MODEL",
+)
```
AarushiShah-db: why do we need to include these? shouldn't ANTHROPIC_MODEL be sufficient, because that is the one that we launch with?

src/ucode/agents/claude.py
```diff
@@ -800,17 +839,25 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
...
+    conflict = _managed_relayed_conflicts()
+    if conflict is not None:
+        managed_path, keys = conflict
         print_err(
-            f"Enterprise managed settings ({managed_path}) set an 'apiKeyHelper', "
-            "which Claude Code always applies over relayed (Claude Max/Enterprise) "
-            "auth — you'd be billed for API usage instead of using your subscription. "
-            "Remove the 'apiKeyHelper' from that file (or ask your admin to) before "
-            "running relayed auth. Not starting Claude Code."
+            "Enterprise managed settings is present, which Claude Code always "
```
AarushiShah-db:
```suggestion
            "Enterprise managed settings are present, which Claude Code always "
```

src/ucode/agents/claude.py
```diff
@@ -800,17 +839,25 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
...
         raise SystemExit(1)
 
+    pinned_models = _managed_default_model_keys()
+    if pinned_models is not None:
+        managed_path, pairs = pinned_models
+        models = ", ".join(f"{key}: {value}" for key, value in pairs)
+        print_warning(
+            f"Default models {models} are set in your enterprise managed settings "
+            f"({managed_path}). Remove those keys if the default models are deprecated."
+        )
```
AarushiShah-db: what about 
```
f"Default models {models} are set in your enterprise-managed settings "
f"({managed_path}) and may override ucode settings. Remove these entries if "
f"you encounter issues."
```

AarushiShah-db (approved):

PR #252, opened 2026-07-30
skills: add --skill subset filter to the download path

xsh310:
## What

Adds a `--skill` filter to the **download** path of `ucode configure skills`, so a user can download a named subset of a schema's skills instead of every skill:

```bash
# Download only the named skill(s) from the schema (comma-separated for several).
ucode configure skills --location cat.sch --skill my_skill
ucode configure skills --location cat.sch --skill skill_a,skill_b
```

This mirrors the existing `--services` subset filter on `ucode configure mcp --location cat.sch --service github`.

## Behavior

- **Subset by leaf name.** Only the named skills are downloaded from each `--location` schema; every other skill is skipped.
- **Applies to every `--location` schema** when several are given.
- **Unknown names warn, don't fail** (`Ignoring requested skills not found in \`cat.sch\`: …`), so a partial match still downloads the rest — matching `--services` handling.
- **Omitted → whole schema** (unchanged default).
- **Download-only.** Rejected with `--mcp` (the skills MCP connection is schema-scoped, with no per-skill route) and rejected without `--location`.
- The schema-less MCP connection registration after download is unchanged; `--skill` never touches `skill_locations`.

The filter is applied purely at the orchestration layer (`download_skills`) — no changes to the download client, the on-disk writer, or `mcp.py`.

## How do you know it works?

- New unit tests in `tests/test_skills_download.py`: filter downloads only matching leaves; unknown names warn and the rest still download; empty-set filter downloads nothing; `None` filter downloads everything; `configure_skills_download_command` threads the set through to `download_skills`.
- New CLI tests in `tests/test_cli.py`: `--skill` dispatches with the parsed set (single + comma list); `--skill` + `--mcp` exits 1; `--skill` without `--location` exits 1. Updated the two existing download-dispatch assertions for the new `skills=` kwarg.
- Full non-e2e suite passes locally (`uv run pytest`, 1055 passed, including `test_lint.py`), and `uv run ruff check .` / `ruff format` are clean.
- Manually smoke-tested the CLI: `--help` shows the flag, and both rejection paths exit 1 with clear messages.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1545,6 +1545,15 @@ def configure_skills(
             help="(download) Existing absolute dir to download into; defaults to your home dir.",
         ),
     ] = None,
+    skill: Annotated[
+        str | None,
+        typer.Option(
+            "--skill",
+            help="(download) Download only this comma-separated subset of skills from each "
+            "--location schema (by leaf name, e.g. `my_skill`), instead of every skill. "
+            "Not valid with --mcp.",
+        ),
+    ] = None,
 ) -> None:
     """Configure Databricks Skills for your coding tools.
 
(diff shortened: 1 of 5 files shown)


src/ucode/cli.py
```diff
@@ -1554,18 +1563,35 @@ def configure_skills(
...
+        )
         if mcp and path is not None:
             raise RuntimeError("--path is not valid with --mcp.")
+        if mcp and selected_skills is not None:
+            raise RuntimeError("--skill is not valid with --mcp; it only applies when downloading.")
         if path is not None and not locations:
             raise RuntimeError("--path only applies when downloading with --location.")
+        if selected_skills is not None and not locations:
+            raise RuntimeError("--skill only applies when downloading with --location.")
+        # A single skill-name set can't be split across schemas, so require one
```
xsh310: nit: remove this comment, it's clear in the error message below

src/ucode/skills_download.py
```diff
@@ -262,6 +271,14 @@ def download_skills(workspace: str, token: str, locations: list[str], path: str
         if reason:
             print_warning(f"Skipping `{location}`: {reason}.")
             continue
+        if skills is not None:
+            unknown = skills - set(leaves)
+            if unknown:
+                print_warning(
+                    f"Ignoring requested skills not found in `{location}`: "
```
xsh310: nit: Skipping requested skill ...

src/ucode/cli.py
xsh310: Done — removed the comment in 5a5db32.

src/ucode/skills_download.py
xsh310: Done in 5a5db32 — reworded to "Skipping requested skill(s) not found in ...". Kept "skill(s)" since the set can list more than one name.

sunishsheth2009 (approved):

PR #256, opened 2026-08-02
Feat: Align Pi and OpenCode model family options

anthonyivn2:
## Summary

- expose OpenAI/Codex models in OpenCode alongside Claude, Gemini, and OSS
- expose OSS models in Pi alongside Claude, OpenAI/Codex, and Gemini
- route Pi OSS models through the MLflow Responses API so GLM and Kimi complete with valid stream termination
- exclude incompatible `gpt-oss` model services from Codex discovery instead of exposing options that fail in Pi and OpenCode
- extend focused and end-to-end coverage for discovery, provider config, selectors, defaults, and live OSS launches

## Testing

- `.venv/bin/ruff check .`
- `.venv/bin/pytest tests/test_agent_pi.py tests/test_agent_opencode.py tests/test_agents_init.py tests/test_databricks.py tests/test_cli.py tests/test_lint.py -q` (`471 passed`)
- live validation with a Databricks profile across representative Claude, OpenAI/Codex, Gemini, GLM, and Kimi models in both clients
- exhaustive live per-model E2E for OpenCode and Pi (`2 passed in 239.58s`)

### DIFF
--- a/src/ucode/agents/pi.py
+++ b/src/ucode/agents/pi.py
@@ -1,6 +1,6 @@
 """Pi coding agent: writes ~/.pi/agent/models.json with Databricks-backed providers.
 
-Pi (https://pi.dev) is a multi-provider coding agent. We register three
+Pi (https://pi.dev) is a multi-provider coding agent. We register four
 providers in its `models.json`, each speaking the API dialect best suited to
 that family's gateway path:
 
@@ -9,7 +9,7 @@ that family's gateway path:
 - `databricks-gemini`  (api: google-generative-ai)     → /ai-gateway/gemini/v1beta
 - `databricks-oss`     (api: openai-responses)         → /ai-gateway/mlflow/v1
 
-Per-provider `compat` flags work around fields the gateway translators reject:
+One provider-specific `compat` flag works around a field the gateway translator rejects:
 
 - claude: `supportsEagerToolInputStreaming: false` — the Anthropic translator
   rejects `tools[].eager_input_streaming` on the streaming + tools path that
@@ -77,7 +77,7 @@ PROVIDER_KEYS: list[list[str]] = [["providers", name] for name in PROVIDER_NAMES
 
 # Old provider names earlier ucode versions wrote; cleaned up on each write so
 # users don't end up with stale entries pointing at routes that 400.
-LEGACY_PROVIDER_NAMES = ("databricks-anthropic", "databricks-codex", "databricks-kimi")
+LEGACY_PROVIDER_NAMES = ("databricks-anthropic", "databricks-codex")
 
 
 def is_update_available() -> tuple[str, str] | None:
--- a/src/ucode/agents/pi.py
+++ b/src/ucode/agents/pi.py
@@ -7,6 +7,7 @@ that family's gateway path:
 - `databricks-claude`  (api: anthropic-messages)       → /ai-gateway/anthropic
 - `databricks-openai`  (api: openai-responses)         → /ai-gateway/codex/v1
 - `databricks-gemini`  (api: google-generative-ai)     → /ai-gateway/gemini/v1beta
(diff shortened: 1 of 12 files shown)


anthonyivn2:
@rohita5l can I get you or the team to OK this?

src/ucode/agents/pi.py
```diff
@@ -68,13 +70,14 @@
...
     "databricks-gemini",
+    "databricks-oss",
 )
 
 PROVIDER_KEYS: list[list[str]] = [["providers", name] for name in PROVIDER_NAMES]
 
 # Old provider names earlier ucode versions wrote; cleaned up on each write so
 # users don't end up with stale entries pointing at routes that 400.
-LEGACY_PROVIDER_NAMES = ("databricks-anthropic", "databricks-codex", "databricks-oss")
+LEGACY_PROVIDER_NAMES = ("databricks-anthropic", "databricks-codex", "databricks-kimi")
```
AarushiShah-db: where did kimi come from? I dont see a old ucode version that supported kimi?

src/ucode/databricks.py
```diff
@@ -1354,7 +1356,7 @@ def discover_model_services(
         if candidates:
             claude_models[family] = candidates[0]
 
-    codex_models = [m for m in ids if "gpt-" in m]
+    codex_models = [m for m in ids if "gpt-" in m and "gpt-oss-" not in m]
```
AarushiShah-db, line 1519: The gpt-oss exclusion only covers the UC path. The filter is in discover_model_services, but cli.py:426-427 falls back to discover_codex_models when the UC listing comes back empty for codex, and that path (discover_endpoints_with_api_type(...,"openai/v1/responses")) has no gpt-oss filter. So on a workspace without UC model-services, gpt-oss still reaches Pi/OpenCode and still  fails the same way. Probably wants to live one layer down, or in both places

tests/test_agent_pi.py
```diff
@@ -339,7 +399,7 @@ def test_legacy_providers_removed_on_upgrade(self, tmp_path, monkeypatch):
             pi_mod.write_tool_config(self._state(), "claude-sonnet", token="tok")
 
         written_providers = json.loads(config_file.read_text()).get("providers", {})
-        for legacy in ("databricks-anthropic", "databricks-codex", "databricks-oss"):
+        for legacy in ("databricks-anthropic", "databricks-codex", "databricks-kimi"):
```
AarushiShah-db: This swap means the databricks-oss cleanup is no longer tested. The old version of this test proved that a stale databricks-oss entry (on openai-completions) got stripped on upgrade; retargeting it to databricks-kimi drops that guard right as the PR removes databricks-oss from  LEGACY_PROVIDER_NAMES. Could we keep a case for it — a stale databricks-oss entry with oss_models empty, asserting it's gone after the
write? That's the scenario I'm unsure about in my other comment.

tests/test_databricks.py
```diff
@@ -217,18 +229,21 @@ def test_buckets_families_by_name(self, monkeypatch):
             "opus": "system.ai.claude-opus-4-8",
             "sonnet": "system.ai.claude-sonnet-4-6",
         }
+        # gpt-oss is not Responses-compatible with Pi/OpenCode and must not
```
AarushiShah-db, line 233: This covers the exclusion on the UC path, but discover_codex_models (the fallback at cli.py:426-427 when the UC listing is empty for codex) has no gpt-oss filter and isn't tested here. Worth a case on that path too — it's the one where gpt-oss can still reach Pi/OpenCode.

anthonyivn2:
@AarushiShah-db I addressed the feedbacks, thank you

AarushiShah-db (dismissed):

nerdrew:
This looks like it would enable the `databricks-oss` models for `pi`. Is that true?

PR #261, opened 2026-08-03
Show budget spend and threshold in `ucode usage`

andy-xu-db:
## Problem

`ucode usage` shows tokens only — no dollar spend or budget context.

## Change

Add the caller's coding-agent budget spend and alert threshold to the `ucode usage` summary, read off AI Gateway's `coding-agent-configs:recommendModel` (posted with an empty `available_models`, since we want the spend the recommendation was based on, not the recommendation — `recommended_models` is ignored):

```
Budget spend: $2.34 of $5,000.00 (0%)
[█░░░░░░░░░░░░░░░░░░░░░░░░░░░░░]
```

Amounts are parsed as `Decimal`. The endpoint is off by default per org, so absence returns a reason string rather than raising — the lines are simply omitted and the rest of the report is unchanged.

## Testing

29 new unit tests; 1167 passed, ruff clean.

Verified against the live endpoint on a workspace with the flag on: `current_spend` `2.3441188` / `effective_threshold` `5000` renders as `$2.34 of $5,000.00 (0%)`. A workspace with the flag off returns `FEATURE_DISABLED` and correctly renders no lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ skipped in a re-run is dropped.
 | Command | Description |
 |---------|-------------|
 | `ucode status` | Show current workspace, base URLs, managed config files, and selected models |
-| `ucode usage` | Show AI Gateway usage summary |
+| `ucode usage` | Show AI Gateway usage summary, plus your budget spend against its alert threshold when the workspace reports one |
 | `ucode revert` | Clear saved state and restore backed-up config files |
 | `ucode configure --dry-run` | Preview config files without writing them |
 | `ucode configure --agents claude,codex` | Configure specific agents without the interactive picker |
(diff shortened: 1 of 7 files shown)


AarushiShah-db (dismissed):

AarushiShah-db (commented):
one comment - could you add a spinner for the budget check 

>   Every other network call in usage() is wrapped — token retrieval, warehouse
>   discovery, the usage query. This one sits bare between the query and the
>   summary print, so ucode usage can appear hung for up to 10s with no
>   indication. AGENTS.md asks for the Rich UI conventions to be preserved;
>   wrapping it in spinner("Checking budget spend...") matches the surrounding
>   code.

AarushiShah-db (dismissed):

AarushiShah-db (dismissed):

AarushiShah-db (approved):

PR #262, opened 2026-08-03
add managed CodingAgentConfig read client, normalize, and local persist

asujithan:
## Changes

Developer-read side of [Coding Gateway Configuration Management]). Adds the data layer that lets a developer's ucode pull the org admin's CodingAgentConfig and stores it locally to ~/.ucode/managed-state.json, so later PRs can reconcile it against local state and apply it to agents.

This PR is the foundation only — no CLI wiring or agent application yet (those are follow-up PRs). Nothing here runs during a real `ucode`/`ucode configure` command; it's exercised by tests.

## Changes

- **`databricks.py`**: `fetch_managed_coding_agent_configs(workspace, token)` — reads the workspace's config via the non-admin List endpoint (`GET /api/ai-gateway/v2/coding-agent-configs`), returning raw manifest dicts or an actionable reason. Workspace is resolved server-side, so no id is needed.
- **`managed_config.py`** (new):
  - `normalize_managed_config` — proto-JSON → ucode-native shape (maps `CODING_AGENT_*`/`MCP_SERVER_TYPE_*` enums to ucode tool/type names; drops agents/types this build doesn't recognize).
  - `get_managed_config` — fetch + normalize, with a 3-way return (config / no-config / error).
  - `save/load/delete_managed_state` — persist to `~/.ucode/managed-state.json` (0600; load is workspace-scoped so a stale file is ignored).

## Test Plan
- **Tests**: 22 cases covering normalization, the fetch client's parsing/error paths, and persistence (round-trip, 0600, workspace-scoping).

## Not in this PR (follow-ups )
- Reconcile managed vs local state + apply manifest to agents/MCP/skills/tracing.
- `ucode configure` gating + bare-`ucode` Path A wiring.
- Sudo/root-ownership enforcement of `managed-state.json` (0600 is an interim best-effort; real enforcement needs a privileged path + elevation and is its own change).

### DIFF
--- a/src/ucode/managed_config.py
+++ b/src/ucode/managed_config.py
@@ -23,6 +23,10 @@ from ucode.databricks import fetch_managed_coding_agent_configs
 
 MANAGED_STATE_PATH = config_io.APP_DIR / "managed-state.json"
 
+# Shown to a developer when their workspace has no admin-defined managed config yet — the normal
+# case, not an error. Kept here so the CLI (which surfaces it) uses one consistent message.
+NO_MANAGED_CONFIG_MESSAGE = "No coding-agent config has been set up by your workspace admin yet."
+
 # CodingAgent proto enum -> ucode tool name. Anything unrecognized (e.g. a newer agent this ucode
 # build doesn't know) is dropped during normalization rather than guessed at.
 _AGENT_ENUM_TO_TOOL: dict[str, str] = {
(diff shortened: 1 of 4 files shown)


src/ucode/managed_config.py
```diff
@@ -0,0 +1,297 @@
...
+    return result
+
+
+def get_managed_config(workspace: str, token: str) -> tuple[dict | None, str | None]:
+    """Fetch and normalize the workspace's managed config.
+
+    Returns ``(config, reason)``:
+    - ``(config, None)`` — the normalized manifest for the workspace's single config;
+    - ``(None, None)`` — no managed config is defined for the workspace (not an error);
+    - ``(None, reason)`` — the read failed; ``reason`` says why.
```
AarushiShah-db, line 241: I think if no config exists it will throw NOT_FOUND error? https://sourcegraph.prod.databricks-corp.com/databricks-eng/universe/-/blob/ai-gateway-api/dbv2/main/CodingAgentConfigEntityStore.scala?L118-131 - we should catch that

AarushiShah-db (approved):

asujithan:
Closing, will instead merge https://github.com/databricks/ucode/pull/263 which will skip e2e tests (failing due to unrelated codex models list error)

PR #265, opened 2026-08-04
Resolve managed coding-agent config over local state at launch

asujithan:
### Changes

Applies the admin-published managed coding-agent config at launch, taking precedence over the developer's own ucode settings.

On every ucode <agent> launch, the workspace's managed config is re-read and persisted to ~/.ucode/managed-state.json, then resolved against ~/.ucode/state.json per key — the manifest wins, the developer's state fills any gaps. The resolved view renders the agent settings file (e.g. ~/.claude/ucode-settings.json); the two state files are never merged on disk, so state.json keeps recording what the developer configured for themselves.

— managed_resolve.py — pure precedence logic. Claude resolves per model family (a family the admin didn't pin keeps the developer's choice); flat-list agents take the manifest's list wholesale.
— managed_config.py — adds the per-launch refresh. A failed read falls back to the last persisted config rather than blocking the launch; a successful "no config" read clears the persisted copy so a removed policy can't reappear after an outage.
— state.py — save_state swaps managed values back for the developer's own before writing, so precedence never overwrites state.json.
— cli.py — wires this into _launch_tool. An explicit --provider that conflicts with the admin's is a hard error; a managed provider that can't be used names the managed config in the error.

**Followups**: 
- models and provider services, MCPs/skills, custom_headers, budget_policy/recommendModels, and sudo ownership of managed-state.json are follow-up PR
- ucode configure / first-run auto-configure still pull agent settings from local state in this PR; follow-up PR will gate configure when a managed config exists ("deprecated, run ucode" message + path)

### Test plan

uv run pytest — 1173 passed, 6 skipped. ruff check . and ruff format --check . clean.
**tests/test_managed_resolve.py**
— TestClaudeModels (5) — proto slots → families; manifest wins per family; unpinned family keeps developer's value; falls back to local; None when neither side has models
— TestListModels (3) — manifest list replaces local; local stands when manifest silent; blank entries dropped
— TestManagedProviderService (3) — manifest-only read, no fallback to local state
— TestResolveState (4) — input state not mutated; managed values layered onto the copy; unrelated keys preserved; other tools' providers untouched
— TestStateFileIsNotRewritten (8) — the core contract, run against a real temp state.json with save_state unstubbed: developer's model survives; settings file still gets the managed model; overlay bookkeeping never on disk; repeated saves stay idempotent (the relayed-proxy path saves twice); provider_services preserved; a developer with no prior value isn't given one; codex + gemini parametrized

**tests/test_managed_config.py**
— TestRefreshManagedConfig (9) — persists and returns the manifest; no-config returns None; no-config clears the persisted copy; read failure and auth failure each fall back to persisted, or to local settings when there is none; empty persisted marker isn't treated as a fallback; no workspace is a no-op
— TestManagedLaunchState (2) — layers managed models when a config exists; hands state back untouched when none
— TestPersistence — empty config overwrites a previous one

Ran e2e tests locally against eng-ml-inference-us-east-1 workspace and all tests pass except tests/test_e2e.py::TestCodexLaunch::test_launch_codex_per_model which is expected due to codex model list fix not yet being rolled out to prod.

**First command: with env var set uses managed config and launches with opus-4-8 as specified by admin**
**Second command: fallback path uses my user settings not managed config so claude launched with opus-5 when env var unset**

https://github.com/user-attachments/assets/ec4d6e25-54b0-41de-9c80-978691f6d1c8

Note: Clearer warning messages for when fallback path is being used and edits to ucode configure command will be added in followup pr

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -55,6 +55,8 @@ from ucode.databricks import (
     resolve_pat_token,
     run_databricks_login,
 )
+from ucode.managed_config import managed_launch_state
+from ucode.managed_resolve import managed_provider_service
 from ucode.mcp import (
     MCP_CLIENTS,
     SKILLS_MCP_KIND,
@@ -1170,6 +1172,9 @@ def _launch_tool(
         if needs_auto_configure:
             _auto_configure_tool(tool)
         state = ensure_provider_state(tool)
+        # Remembered before the fallback below collapses the two cases: a managed config may not
+        # silently override a provider the user typed on the command line (it errors instead).
+        explicit_provider = provider
         # An explicit --provider overrides the persisted choice; otherwise fall
         # back to whatever `ucode configure` saved for this tool.
         provider = provider or get_provider_service(state, tool)
@@ -1179,17 +1184,6 @@ def _launch_tool(
                 f"{TOOL_SPECS[tool]['display']} smart routing cannot be enabled with "
                 "--provider. Launch without a Model Provider Service and try again."
             )
-        # Validate the provider service before launching — it must exist, be a
-        # provider type this tool can route to (e.g. claude can't use an OpenAI
-        # or Foundry service), and, for Bedrock, expose Claude models to pin.
-        # Surfaces a clear error up front instead of a cryptic gateway failure
-        # mid-session. For a Bedrock service this also returns the model ids.
-        provider_models = None
-        relayed = False
-        if provider:
-            provider_models, error, relayed = resolve_provider_models(tool, state, provider)
-            if error:
-                raise RuntimeError(error)
         # Re-fetch model lists on every launch so newly-added Databricks
         # endpoints show up without a manual `ucode configure` (and so that
         # tools like pi which read multiple model bundles never run on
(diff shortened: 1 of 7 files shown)


src/ucode/cli.py
```diff
@@ -1202,6 +1196,38 @@ def _launch_tool(
             skip_model_discovery=bool(provider),
             skip_preflight=skip_preflight,
         )
+        # An admin-published managed config wins over the developer's own settings. Resolved before
+        # the provider and model are settled below, so each is decided once against the values that
+        # will actually be written — the two state files are never merged on disk.
+        state, managed = managed_launch_state(state, tool)
```
AarushiShah-db: omniagent uses the --skip-preflight command to skip any extra network calls (to not add latency) - can we make sure that if  --skip-preflight is passed in, we don't do the managed config pull and then instead fallback to whatver managed-state.json already exisits locally ( if any)?

src/ucode/managed_resolve.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Claude keys its models by family (``opus``/``sonnet``/``haiku``/``fable``) and resolves per
+    family, so a family the manifest omits keeps the developer's value. Every other agent stores a
+    flat list, which has no per-key identity — there the manifest's list replaces the local one
+    outright, or the local one stands when the manifest specifies none.
+    """
+    manifest_models = _agent_model_config(managed, tool).get("models")
+    if tool == "claude":
+        local = dict(_as_dict(state.get("claude_models")))
+        for slot, family in _CLAUDE_FAMILY_SLOTS.items():
```
AarushiShah-db, line 71: what about default model? shouldn't we always be launching with the default model specified in the config?

src/ucode/managed_config.py
```diff
@@ -307,11 +315,67 @@ def load_managed_state(workspace: str | None) -> dict | None:
...
+
+
+def managed_launch_state(state: dict, tool: str) -> tuple[dict, dict | None]:
+    """Return ``(state, managed)`` for launching ``tool`` under any managed config.
+
+    The returned state has the manifest's models and provider layered over the developer's own —
+    managed wins per key — so the settings file written from it reflects the admin's choices. When
+    the workspace has no managed config the state is handed back untouched.
+    """
+    managed = refresh_managed_config(state)
```
AarushiShah-db: We should print something out and have a spinner when we are trying to pull the managed config otherwise to the user it will just look like ucode is really slow

Maybe: 
"Checking for a managed coding agent config"

AarushiShah-db (commented):
is it possible to gate this whole change behind an env var for now? I don't want to release the new functionality to users until we've properly bug bashed it 

Im thinking we have a env var for now like ENABLE_MANAGED_AGENT_CONFIG and if it is set then only do we pull the config on launch

src/ucode/managed_config.py
```diff
@@ -307,11 +315,67 @@ def load_managed_state(workspace: str | None) -> dict | None:
...
+    persisted = load_managed_state(workspace)
+    if persisted:
+        print_warning(
+            f"Could not read your workspace's managed config ({reason}); "
+            "using the last one saved for this workspace."
+        )
+        return persisted
+    print_warning(
+        f"Could not read your workspace's managed config ({reason}); using your local settings."
+    )
```
AarushiShah-db: someone with an expired session and no managed config would get a warning about a feature they don't use - might be worth not printing a warning when there's no persisted config at all.

src/ucode/cli.py
```diff
@@ -1170,6 +1172,9 @@ def _launch_tool(
         if needs_auto_configure:
             _auto_configure_tool(tool)
         state = ensure_provider_state(tool)
+        # Remembered before the fallback below collapses the two cases: a managed config may not
+        # silently override a provider the user typed on the command line (it errors instead).
+        explicit_provider = provider
         # An explicit --provider overrides the persisted choice; otherwise fall
         # back to whatever `ucode configure` saved for this tool.
         provider = provider or get_provider_service(state, tool)
```
AarushiShah-db, line 1180: the smart routing guard runs before the managed config sets the provider, so you can end up with
routing persisted as "on" while a provider is set, which is the combination the guard exists to prevent.

I think we may have to move this check to after the maanged config sets the provider

AarushiShah-db (approved):

PR #267, opened 2026-08-05
Add managed-config serializer, validation, and per-agent model catalogs

tt-le:
### Changes

The admin-**write** foundation for the managed `CodingAgentConfig`, mirroring the developer-read side from #263. No CLI wiring and no network calls: pure functions plus tests. The interactive `ucode setup` wizard (#PR2) and `ucode apply` build on this.

**`managed_setup.py`** (new)

— `serialize_managed_config` — ucode-native manifest → proto-JSON `CodingAgentConfig`, the exact inverse of `normalize_managed_config`. Builds each agent's `AgentModelConfig` oneof variant, which the server rejects when it doesn't match its agent: claude gets `ClaudeDefaultModels` family slots, codex gets no model list at all, and opencode/pi/gemini/copilot get a flat list. Output-only fields (`workspace_id`, timestamps, user ids) are never emitted.

— `validate_manifest` — client-side pre-flight mirroring the handler's `validateStoredConfig`, so an admin sees problems before the round-trip rather than as an opaque `INVALID_PARAMETER_VALUE`. Also cross-checks configured models against the workspace's discovered inventory, skipped for agents routing through a Model Provider Service (those ids come from the provider's catalog).

— Per-agent model catalogs: claude → Claude only, gemini → Gemini only, codex → GPT + OSS, opencode/pi/copilot → everything discovered.

— `~/.ucode/managed-settings.json` persistence (0600, workspace-scoped), distinct from `managed-state.json`: this is the *authored* manifest, that is the *pulled* one.

**`managed_config.py`** — makes the two enum maps public so the write side inverts them instead of restating them, so a new agent or MCP type is declared once and both directions pick it up. Mechanical rename otherwise.

### Note on units

`spending_percentage` is a fraction in `[0, 1]`, not a percent — the server validates that range. `validate_manifest` rejects e.g. `80` with a message naming the expected form, since the spec doc's prose says "80%".

### Testing

`uv run pytest` — 1218 passed, 36 skipped on this commit alone. 77 new cases.

The load-bearing one is the **round-trip property** (`serialize` → #263's `normalize` == identity), asserted over the full manifest, the minimal manifest, every known agent, and every MCP type. It pins read and write together so they cannot drift as the proto grows — which matters now that #265 consumes this same manifest shape.

### DIFF
--- a/src/ucode/managed_setup.py
+++ b/src/ucode/managed_setup.py
@@ -31,6 +31,7 @@ from typing import cast
 import ucode.config_io as config_io
 from ucode.databricks import (
     ANTHROPIC_FAMILIES,
+    model_version_sort_key,
     tool_supports_provider_type,
 )
 from ucode.managed_config import (
@@ -61,8 +62,9 @@ _AGENT_MODEL_CONFIG_VARIANT: dict[str, str] = {
 # (`ClaudeDefaultModels`), and Codex has no model list at all — it selects exactly one model.
 _FLAT_MODEL_LIST_AGENTS = frozenset({"opencode", "pi", "gemini", "copilot"})
 
-# Claude family slot names in `ClaudeDefaultModels`, keyed by ucode's family name.
-_CLAUDE_SLOT_FOR_FAMILY: dict[str, str] = {
+# Claude family slot names in `ClaudeDefaultModels`, keyed by ucode's family name. Public because
+# the wizard prompts one slot at a time.
+CLAUDE_SLOT_FOR_FAMILY: dict[str, str] = {
     family: f"default_{family}_model" for family in ANTHROPIC_FAMILIES
 }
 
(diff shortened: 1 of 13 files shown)


src/ucode/managed_setup.py
```diff
@@ -0,0 +1,561 @@
...
+
+
+def _validate_agent_models(tool: str, agent_config: dict, known: set[str]) -> list[str]:
+    """Check one agent's configured models against the workspace inventory.
+
+    Skipped entirely when the agent routes through a Model Provider Service: those model ids come
+    from the provider's own catalog, not from UC model services, so the workspace inventory says
+    nothing about them.
+    """
+    model_config = agent_config.get("model_config")
```
AarushiShah-db, line 403: here can we also check against AGENT_MODEL_FAMILIES to make sure that the user isn't specifying a gpt model for claude code etc.

src/ucode/managed_setup.py
```diff
@@ -0,0 +1,561 @@
...
+# `AgentModelConfig` oneof variant key per agent. The server rejects a config whose variant doesn't
+# match its agent (`validateAgentModelConfig`), so this mapping is not cosmetic.
+_AGENT_MODEL_CONFIG_VARIANT: dict[str, str] = {
+    "claude": "claude",
+    "codex": "codex",
+    "opencode": "opencode",
+    "pi": "pi",
+    "gemini": "gemini",
+    "copilot": "copilot",
+}
```
AarushiShah-db, line 59: do we need this mapping if its anyways the same name for the key -> value? I think we have another list somewhere in the codebase with all the supported agents, can we use that? Just trying to reduce the number of places that we need to update everytime we want to support a new harness

AarushiShah-db (dismissed):
few comments but stamping to unblock

src/ucode/managed_setup.py
tt-le: I'll remove, its pretty redundant

src/ucode/managed_setup.py
tt-le: I'll fold this one into the TUI flow pr

AarushiShah-db (dismissed):

AarushiShah-db (approved):

PR #268, opened 2026-08-05
Add `ucode setup`: interactive managed-config authoring

tt-le:
### Demo

https://github.com/user-attachments/assets/767f90ba-aed6-40f2-9ad2-c74efa8be376

### Changes

> **Stacked on #267** — review that one first. This PR's diff includes it until #267 merges.

The admin-facing wizard on top of #267's serializer. Walks a workspace admin through agents, per-agent models, tracing, MCP servers, skills, and a spend-routing budget policy, then writes the manifest to `~/.ucode/managed-settings.json`. Publishing is a separate `ucode apply` (next PR) so an admin can review the file first — and `ucode setup show` prints both the summary and the exact proto-JSON that apply will POST.

`managed_wizard.py` is the interaction layer only; catalogs, validation, and serialization stay in `managed_setup.py`. Sub-flows an admin already knows are delegated to the existing commands (`configure_tracing_command`, `configure_mcp_command`, `configure_skills_mcp_command`) and their results read back out of `state.json`, so there is still exactly one picker per concern.

### Claude Code is prompted per model family

`ClaudeModelConfig.models` is a `ClaudeDefaultModels` message — four optional slots (`default_opus_model` / `default_sonnet_model` / `default_haiku_model` / `default_fable_model`), since Claude Code addresses models by family alias rather than from a list. A flat multi-select cannot express that.

```
? Default opus model:
  › system.ai.claude-opus-5
    system.ai.claude-opus-4-8
    system.ai.claude-opus-4-7
    ... (skip opus)
```

Each family is skippable (the slots are `optional`, and an unset one falls back to `default_model`); the overall `default_model` is then chosen from the slots that were filled, so it can never name a model the config doesn't carry. With one slot filled there is nothing to choose, so it reports what it inferred rather than asking a single-option question.

That required looking past `state["claude_models"]`, which is `{family: newest}` because the launch path pins one model per alias — the workspace tested against has 12 Claude models and the bucketed view shows 4. `discover_claude_models_unbucketed` returns the full listing, `claude_family_candidates` groups it by family, and `_known_models` reads the same set so validation cannot reject a version the prompts offered. `discover_model_services` is untouched — every agent's config writer depends on the shape of `claude_models`.

### Other model-prompt behaviour

— **Provider-service agents get a real picker.** `list_model_provider_services` already returns each service's `config.targets`; the wizard now selects the whole service dict and offers those, instead of an empty text prompt that required knowing a provider-side id by heart. `allow_all_targets` and relayed Anthropic subscription services have nothing to enumerate, so those fall back to free text — but say why and give an example.

— **Gemini and Copilot get one model**, not a multi-select: their writers take a single model and set one env var, so extra picks were silently discarded at apply time. Wizard-level only — both protos declare `repeated string models`, so `--from-file` still serializes a list. OpenCode and Pi keep their multi-select; they really do show a model picker.

— **Nothing is pre-checked.** `options[0]` is whatever discovery sorted first, not a recommendation — for Pi a Claude model, for Codex the oldest GPT. It only existed to make Enter work, which stopped mattering once every agent required a model: an empty selection is re-prompted. That also makes `_prompt_models_for_agent` total. Without it, picking a multi-model agent and hitting Enter produced an agent with no `model_config`, and if it was the `default_agent` the run died at the *end* — after tracing, MCP, skills, and budget had all been answered.

### Budget tiers offer only the agent's own models

The picker previously read the workspace catalog and called `load_state()`, so it could not have seen this run's choices even in principle — configuring Pi with one Kimi model still offered every family. Picking one of the extras was not harmless: `validateStoredConfig` checks only that a tier's `default_agent` is in `enabled_agents`, never that the agent *has* the model, so the tier would activate and hand a developer a model their agent was never given. `validate_manifest` gains that missing rule, which is the durable half — it guards `--from-file` and hand-written manifests too.

Tiers are prompted in percent and stored as fractions, keeping the spec-vs-API units mismatch in one place.

### Smaller notes

Model ids stay bare (`system.ai.claude-opus-4-8`), not provider-prefixed: each agent's writer adds whatever its config format needs, so the manifest stays agent-neutral.

Two readback details: tracing is stored as `tracing.uc_destination` but the managed config calls it `tracing.table`, and an enabled-but-not-UC-backed experiment has no table so it is omitted rather than published empty; `state.json` records each MCP server's URL but not its type, so `_mcp_type_for_url` maps the URL shape back and skips anything unrecognized instead of guessing.

`databricks.py` adds `is_workspace_admin` (SCIM `Me` group membership) and `list_workspace_budgets` (workspace-scoped — ucode never *creates* budgets, an admin picks an existing one). The admin gate errors on a definite non-admin and warns-then-continues when the check itself fails, since the API enforces the same rule server-side.

### Testing

`uv run pytest` — 1334 passed, 36 skipped. 116 new cases. (`tests/test_e2e_user_agent.py::test_user_agent_arrives_at_gateway` fails on my machine because `/etc/claude-code/managed-settings.json` overrides `ANTHROPIC_BASE_URL` past the test's capture server; it fails on `main` too and is unrelated.)

Several are mutation-checked: reverting the claude flat multi-select fails eight, reverting the budget-tier picker fails `test_offers_only_the_models_the_agent_was_configured_with`, and dropping the tier-model validation rule fails `test_tier_model_must_be_one_the_agent_has`.

Verified interactively against a staging workspace: six opus versions and four sonnet offered, a skipped family absent from `models`, and the result round-trips through `normalize_managed_config` to the proto shape.

### Known gap

`discover_model_services` and `discover_claude_models_unbucketed` both page `/api/2.1/unity-catalog/model-services`, so a run that configures Claude lists it twice (one extra spinner). `_claude_candidates` caches on `state["all_claude_models"]`, bounding it to once per run. Removing the second listing means widening `discover_model_services`' 5-tuple return, which is asserted in eight tests and mocked in sixteen — left for a follow-up rather than growing this PR.

### DIFF
--- a/src/ucode/managed_wizard.py
+++ b/src/ucode/managed_wizard.py
@@ -393,6 +393,15 @@ def _claude_candidates(state: dict) -> dict[str, list[str]]:
     Caches the full listing on ``state["all_claude_models"]`` so `validate_manifest` recognizes the
     older versions these prompts offer — ``claude_models`` alone holds just the newest per family,
     and would reject a legitimately-picked ``claude-opus-4-8``.
+
+    INVARIANT: whatever this returns must be recognizable by ``validate_manifest``, which reads
+    ``all_claude_models`` (falling back to ``claude_models``) via ``_known_models``. The two paths
+    below both satisfy it, for different reasons: the listing path widens the candidates *and* sets
+    the cache, while the fallback path sets nothing but also narrows the candidates to
+    ``claude_models``, which ``_known_models`` already covers. Widening the fallback without also
+    populating the cache breaks the invariant, and the symptom is a confusing rejection at the very
+    end of the flow ("claude: model 'system.ai.claude-opus-4-8' is not available on this
+    workspace") rather than an error at the prompt that offered it.
     """
     cached = state.get("all_claude_models")
     if isinstance(cached, list) and cached:
@@ -421,15 +430,16 @@ def _claude_candidates(state: dict) -> dict[str, list[str]]:
 def _require_selection(prompt: str, options: list[tuple[str, str]]) -> str:
     """Single-select that won't take "nothing" for an answer.
 
-    ``prompt_for_selection`` returns None both for Ctrl-C and for an empty submission. questionary
-    raises KeyboardInterrupt on Ctrl-C before returning, so a None here means the picker was
-    dismissed without a choice — re-ask instead of silently producing a model-less config.
+    ``prompt_for_selection`` returns None for both Ctrl-C and an empty submission, and the two are
+    genuinely indistinguishable here: questionary's ``Question.ask`` catches KeyboardInterrupt
+    internally and returns None (v2.1.1, question.py), so nothing propagates for a caller to see.
+    A None is therefore treated as an abort rather than re-asked — re-asking looped forever on
+    Ctrl-C, printing the error once per keypress and never exiting.
     """
-    while True:
-        answer = prompt_for_selection(prompt, options, searchable=True)
-        if answer:
-            return answer
-        print_err("Please choose one of the options.")
+    answer = prompt_for_selection(prompt, options, searchable=True)
+    if not answer:
+        raise KeyboardInterrupt
+    return answer
 
 
 def _require_multi_selection(
@@ -448,9 +458,15 @@ def _require_multi_selection(
 
 
 def _require_text(prompt: str) -> str:
-    """Free-text prompt that requires a non-empty answer."""
+    """Free-text prompt that requires a non-empty answer.
+
+    ``required=True`` makes closed stdin abort instead of returning None. Without it a
+    non-interactive run (piped stdin, CI) spun here forever: ``prompt_for_text`` returns its default
+    on EOF, the default is None, and the loop re-asked an empty stream. Reachable whenever model
+    discovery finds nothing, which is exactly when a run is most likely to be scripted.
+    """
     while True:
-        answer = prompt_for_text(prompt)
+        answer = prompt_for_text(prompt, required=True)
         if answer:
             return answer
         print_err("Please enter a model id.")
--- a/src/ucode/managed_wizard.py
+++ b/src/ucode/managed_wizard.py
@@ -1,0 +1,836 @@
+"""Interactive `ucode setup`: author the workspace's managed coding-agent config.
+
+Workspace admins run this to build the ``CodingAgentConfig`` their developers will pull. It walks
+the admin through agents, per-agent models, tracing, MCP servers, skills, and a spend-routing budget
+policy, then writes the manifest to ``~/.ucode/managed-settings.json``. Publishing it to the
+workspace is ``ucode apply`` (a separate command, so an admin can review the file first).
+
+Serialization, validation, and the per-agent model catalogs live in :mod:`ucode.managed_setup`; this
+module is the interaction layer on top of them. Sub-flows an admin already knows — tracing, MCP,
+skills — are delegated to the existing ``ucode configure <thing>`` commands and their results read
+back out of ``state.json``, so there is exactly one picker per concern in the codebase.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from ucode.agents import TOOL_SPECS, check_gateway_endpoint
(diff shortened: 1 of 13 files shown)


AarushiShah-db (changes requested):
Some high level comments based on the video 

- Configuration Summary - can we put a box around it 

Image

- Dry run: is there a reason why its ucode configure --dry-run? Can it just be ucode —dry-run?
- Warning message update: "A published Claude configuration already exists for this workspace. Publishing will replace it. Make sure this configuration includes everything you want to keep."
- For the model picker , can we add option to filter by search (you type and it will filter the models) Same with budget policy tracker
- What is the initial "Fetching available models" step doing and is that needed? because aren't we doing individual API calls anyways to get the models for the individual agents?

src/ucode/ui.py
```diff
@@ -302,11 +336,130 @@ def prompt_for_tools(available: list[tuple[str, str]]) -> list[str]:
...
+    answer = questionary.checkbox(
+        prompt,
+        choices=choices,
+        style=style,
+        pointer="›",
+        qmark="",
+        instruction=instruction,
+        use_search_filter=searchable,
+        use_jk_keys=not searchable,
+    ).ask()
```
AarushiShah-db, line 382: This loops forever on Ctrl-C. The docstring says "questionary raises KeyboardInterrupt on Ctrl-C before
returning, so a None here means the picker was dismissed" — but questionary's Question.ask() catches
KeyboardInterrupt internally and returns None (v2.1.1, question.py). So Ctrl-C → None → print_err →
re-prompt, forever. Repros:

ERROR Please choose one of the options.
ERROR Please choose one of the options.   # x N

_require_multi_selection right below it gets this right (if picked is None: raise KeyboardInterrupt) —
_require_selection needs the same.

src/ucode/ui.py
```diff
@@ -302,11 +336,130 @@ def prompt_for_tools(available: list[tuple[str, str]]) -> list[str]:
...
+        except EOFError as exc:
+            if default is not None:
+                return default
+            # Closed stdin with no default to fall back on is the admin abandoning the prompt, which
+            # is what Ctrl-C means here too. Raised as KeyboardInterrupt so the CLI's existing
+            # handler prints "Interrupted." and exits 130; a bare EOFError has no handler anywhere
+            # above this and reached the admin as a traceback.
+            raise KeyboardInterrupt from exc
+        if not raw_value and default is not None:
+            return default
```
AarushiShah-db, line 450: Same infinite loop on closed stdin. prompt_for_text returns default on EOFError, and here default is None, so the while True never exits — a non-interactive run (piped stdin, CI) spins printing "Please enter a model
  id." forever. Reachable whenever discovery finds nothing: _prompt_models_for_agent:281,
  _prompt_claude_models:342, and the provider-service free-text path at :272.

  prompt_for_percentage already solved exactly this — it raises KeyboardInterrupt on EOF with no default, with a good comment about why. Could _require_text do the same, or prompt_for_text grow a required=True?

src/ucode/managed_wizard.py
```diff
@@ -0,0 +1,884 @@
...
+        model_config["default_model"] = chosen[0]
+        print_success(f"Overall default for {display}: {chosen[0]} (the only model configured)")
+    else:
+        model_config["default_model"] = _require_selection(
+            f"Which of those is {display}'s overall default?", [(m, m) for m in chosen]
+        )
+    return model_config
+
+
+def _claude_candidates(state: dict) -> dict[str, list[str]]:
```
AarushiShah-db, line 390: The all_claude_models cache is what lets validate_manifest accept the older versions these prompts offer, and it's only set when the unbucketed listing succeeds (:412). On the fallback path (discover_claude_models_unbucketed raises, or returns empty) it isn't set — which is fine today, because the fallback also narrows the prompts to claude_models (newest per family) so nothing unknown can be picked. But the coupling is invisible and one-directional: anything that later widens the fallback candidates without setting the cache turns into a confusing end-of-flow rejection:

_claude: model 'system.ai.claude-opus-4-8' is not available on this workspace._

A comment tying the two together, or having claude_family_candidates be the single source both read, would keep it from drifting.

AarushiShah-db (approved):

PR #269, opened 2026-08-05
Add GLM to Codex model catalog on startup; remove subagent GLM skip

masonc08:
## What & why

Codex fetches its model catalog from the gateway at startup (`GET /ai-gateway/codex/v1/models`), but the Databricks AI Gateway returns a 400 for that endpoint — so Codex has no model catalog and rejects models it doesn't recognize. This is why GLM was blocked for subagent spawns (the `skip_arms` guard in PR #251): GLM isn't in Codex's catalog, so a `spawn_agent` with `model: system.ai.glm-5-2` would fail.

This PR writes a static model catalog JSON file (`model_catalog_json` config key) on `ucode codex` startup that includes GLM alongside the discovered GPT models. Codex reads the catalog from the file instead of the API, so GLM becomes a valid subagent model — no API call needed.

## The mechanism

- `_write_model_catalog(state)` builds a `ModelsResponse` JSON with one entry per discovered `codex_models` entry, plus a synthetic GLM entry (if not already present).
- The catalog is written to `~/.ucode/codex-model-catalog.json` and referenced via `model_catalog_json = "<path>"` in the codex config overlay.
- The catalog schema was reverse-engineered from the Codex 0.146 binary's serde structs (`ModelInfo` with 38 fields, `ReasoningEffortPreset`, `TruncationPolicyConfig`, etc.) and verified by launching codex with the catalog — it parsed and launched successfully.
- `model_catalog_json` is added to `MANAGED_KEYS` so it's tracked/reverted with the rest of ucode's config.

## The guard removal

With GLM in the catalog, the `skip_arms={GLM_ROUTE_ARM: ...}` guard in `codex_routing.route_pre_tool_use` is no longer needed. A GLM routing decision is now applied like any other arm — the subagent's model is rewritten to `system.ai.glm-5-2` and Codex accepts it.

The `GLM_SUBAGENT_SKIP_MESSAGE` constant is removed. The test `test_spawn_glm_decision_keeps_original_model` is replaced with `test_spawn_glm_decision_applies_glm_model` which asserts GLM is now applied.

## How do you know it works

- Full suite: 1139 passed, lint clean.
- The `model_catalog_json` format was verified empirically: codex `exec` with the catalog launched successfully (no parse errors, no "model not found" errors).
- The gateway's `GET /models` 400 failure was confirmed via trace logging (`failed to refresh available models: ... doesn't match any known API type`), proving the catalog fetch is the root cause.

This is a follow-up to #251 (codex smart routing) and #266 (route subagents on plaintext message). It decouples GLM subagent support from Tien's separate work on fetching the model list from the gateway.

### DIFF
--- a/src/ucode/agents/codex.py
+++ b/src/ucode/agents/codex.py
@@ -35,6 +35,7 @@ CODEX_BACKUP_PATH = APP_DIR / "codex-ucode-config.backup.toml"
 LEGACY_CODEX_CONFIG_PATH = CODEX_CONFIG_DIR / "config.toml"
 LEGACY_CODEX_BACKUP_PATH = APP_DIR / "codex-config.backup.toml"
 CODEX_MODEL_PROVIDER_NAME = "ucode-databricks"
+CODEX_MODEL_CATALOG_PATH = APP_DIR / "codex-model-catalog.json"
 MINIMUM_CODEX_VERSION = (0, 134, 0)
 MINIMUM_CODEX_VERSION_TEXT = "0.134.0"
 MINIMUM_ROUTING_CODEX_VERSION = (0, 145, 0)
@@ -55,6 +56,7 @@ SPEC: ToolSpec = {
 MANAGED_KEYS: list[list[str]] = [
     ["model_provider"],
     ["model"],
+    ["model_catalog_json"],
     ["model_providers", CODEX_MODEL_PROVIDER_NAME],
     ["model_providers", CODEX_MODEL_PROVIDER_NAME, "http_headers"],
 ]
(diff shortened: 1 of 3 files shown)


masonc08:
Closing in favor of #273. Tien's model-list change is deployed to staging, so Codex now fetches a real model catalog from the gateway (`GET /ai-gateway/codex/v1/models` no longer 400s) — GLM is present via the API and this static `model_catalog_json` file is redundant. #273 (just dropping ucode's GLM subagent guard) is the complete fix, and has been validated end-to-end: a GLM-routed subagent spawns successfully.

PR #274, opened 2026-08-05
resolve managed models from the config alone and skip redundant disco…

asujithan:
### Changes
Models specified in a workspace's managed config are now the whole allowlist for that agent, and the per-launch Databricks model discovery is skipped when the config already names them.

**Model resolution**: Each Claude family resolves to its own slot in the config, or to the config's default_model when that slot is unset. With neither set the family is left out entirely, so ucode writes no ANTHROPIC_DEFAULT_<FAMILY>_MODEL and the agent falls back to its own default rather than a model the admin never sanctioned — previously an unset family inherited whatever the developer had discovered locally, which let launches reach models outside the config. The developer's own list still stands when the config names no models for that agent at all.

**Skipping discovery**: managed_supplies_models reports whether the config names a provider, a default_model, or any model list entry, and the launch path folds that into the existing skip_model_discovery flag alongside the --provider case. That drops several seconds per launch — mostly discover_model_services walking Unity Catalog for models the config has already specified. The check reads the persisted managed-state.json rather than the fresh fetch, since the decision has to be made before configure_shared_state runs; it's a local file, and a first launch (no persisted copy yet) discovers exactly as before.

**Enterprise settings warning**: Claude Code's enterprise scope outranks the --settings file ucode writes and can't be excluded with --setting-sources, so a model pinned there quietly beats the admin's. A non-blocking warning now names the file (reusing the same platform discovery as the relayed-auth check) when it pins any of the five model keys ucode selects with.

### Test Plan
uv run pytest - 1209 passed
e2e tests passing

https://github.com/user-attachments/assets/c451857c-fa07-4bbf-858f-580f309473a4

### DIFF
--- a/src/ucode/managed_resolve.py
+++ b/src/ucode/managed_resolve.py
@@ -62,20 +62,20 @@ def effective_agent_models(managed: dict, state: dict, tool: str) -> dict | list
 
     Once the manifest says anything about ``tool``'s models it is the whole allowlist: the
     developer's discovered models drop out entirely, so a launch can only reach models the admin
-    named. Each claude family resolves to its own slot, or to ``default_model`` when that slot is
-    unset; with neither set the family is left out, so ucode writes no
-    ``ANTHROPIC_DEFAULT_<FAMILY>_MODEL`` and the agent uses its own default rather than a model the
-    admin never sanctioned. Every other agent stores a flat list, which has no per-key identity —
-    there the manifest's list replaces the local one outright. Only when the manifest names nothing
-    for ``tool`` does the developer's own list stand.
+    named. Each claude family resolves only to its own slot — a family the manifest leaves out stays
+    unset rather than inheriting ``default_model``, so ucode writes no
+    ``ANTHROPIC_DEFAULT_<FAMILY>_MODEL`` for it and the agent falls back to its own default. Omitting
+    a family is how an admin steers people off it, and filling it in with ``default_model`` would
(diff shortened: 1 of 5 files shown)


src/ucode/managed_resolve.py
```diff
@@ -60,19 +60,28 @@ def _agent_model_config(managed: dict, tool: str) -> dict[str, object]:
...
-    manifest_models = _agent_model_config(managed, tool).get("models")
+    model_config = _agent_model_config(managed, tool)
+    manifest_models = model_config.get("models")
+    default_model = _str(model_config.get("default_model"))
     if tool == "claude":
-        local = dict(_as_dict(state.get("claude_models")))
+        slots: dict[str, str] = {}
         for slot, family in _CLAUDE_FAMILY_SLOTS.items():
-            model = _str(_as_dict(manifest_models).get(slot))
+            model = _str(_as_dict(manifest_models).get(slot)) or default_model
```
AarushiShah-db: this seems like if a family model is not specified it will fallback to applying the default model for each family?

So if default_model = opus 4.5

and no default_sonnet or default_haiku is defined, then 

default_sonnet = opus 4.5
default_sonnet = opus 4.5

We don't want that? Instead it should just be that default_sonnet and default_haiku are left **unset**

asujithan: fixed in latest commit to not pin default_model

AarushiShah-db (approved):

PR #275, opened 2026-08-05
Add `--warehouse-id` and fall through unusable SQL warehouses

andy-xu-db:
## Problem

ucode usage picks the first RUNNING SQL warehouse and aborts if it fails. On workspaces where a warehouse reports RUNNING but refuses connections, the command is unusable. For example:

• Using SQL warehouse `⚡ Reyden Warehouse` (RUNNING).
ERROR Usage query failed: ENDPOINT_NOT_FOUND: SQL warehouse (…) does not exist at all in the database

There was no way to choose a different one. The error text told users to "pass --http-path", a flag that never existed.

## Change

- discover_sql_warehouses() returns all usable candidates, RUNNING first, instead of just one.
- usage tries each in turn, warns on failure, and continues — so one bad warehouse no longer kills the report.
- New --warehouse-id flag skips discovery entirely.

## Testing

11 new unit tests; 1204 passed, ruff clean.

Verified on eng-ml-inference.staging, which has the broken warehouse above: auto-discovery now warns and falls through to a working one (exit 0, where main exits 1), and --warehouse-id pins directly.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -2360,11 +2360,16 @@ def revert_cmd() -> None:
 
 
 @app.command("usage")
-def usage_cmd() -> None:
+def usage_cmd(
+    warehouse_id: Annotated[
+        str | None,
+        typer.Option("--warehouse-id", help="SQL warehouse to query, instead of discovering one."),
+    ] = None,
+) -> None:
     """Show Databricks AI Gateway usage summary (last 7 days)."""
     try:
         install_databricks_cli()
-        usage_report()
+        usage_report(warehouse_id=warehouse_id)
     except RuntimeError as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -2652,7 +2652,14 @@ def run_usage_query(
     http_path: str,
     token: str,
     query: str,
+    on_connected: Callable[[], None] | None = None,
 ) -> tuple[list[str], list[tuple]]:
+    """Run `query` on one warehouse.
+
+    `on_connected` fires once the connection opens — the point a stopped
+    warehouse has finished starting — so callers can update their progress
+    message.
+    """
     try:
         logging.getLogger("databricks.sql").setLevel(logging.ERROR)
         from databricks import sql
@@ -2668,6 +2675,8 @@ def run_usage_query(
             http_path=http_path,
             access_token=token,
         ) as connection:
+            if on_connected is not None:
+                on_connected()
             with connection.cursor() as cursor:
                 cursor.execute(query)
                 columns = [desc[0] for desc in (cursor.description or [])]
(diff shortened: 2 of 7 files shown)


src/ucode/usage.py
```diff
@@ -453,15 +479,11 @@ def usage() -> int:
         token = get_databricks_token(workspace, profile)
 
     with spinner("Discovering SQL warehouse..."):
-        resolved_http_path = discover_sql_warehouse_http_path(workspace, token, quiet=False)
-
-    with spinner("Querying system.ai_gateway.usage..."):
```
AarushiShah-db, line 481: why was this removed? We should still display this message, otherwise the command will just appear as if its hanging while we query

src/ucode/cli.py
```diff
@@ -1936,11 +1936,16 @@ def revert_cmd() -> None:
 
 
 @app.command("usage")
-def usage_cmd() -> None:
+def usage_cmd(
+    warehouse_id: Annotated[
```
AarushiShah-db, line 2392: Can you add this new field to the README.md

src/ucode/databricks.py
```diff
@@ -2259,12 +2259,27 @@ def _looks_like_auth_failure(reason: str) -> bool:
...
+
+def discover_sql_warehouses(
     workspace: str,
     token: str,
     *,
-    quiet: bool = False,
-) -> str:
+    warehouse_id: str | None = None,
+) -> list[SqlWarehouse]:
+    """Candidate warehouses to run the usage query against, RUNNING ones first.
```
AarushiShah-db, line 2725: can we skip the non-running candidate warehouses?

src/ucode/usage.py
```diff
@@ -437,7 +439,31 @@ def render_usage_summary(
...
+    last_error: RuntimeError | None = None
+    for warehouse in candidates:
+        print_note(f"Using SQL warehouse `{warehouse.label}` ({warehouse.state}).")
+        try:
+            columns, rows = run_usage_query(workspace, warehouse.http_path, token, query)
+        except RuntimeError as exc:
+            last_error = exc
+            print_warning(f"SQL warehouse `{warehouse.label}` is unusable: {exc}")
+            continue
+        return warehouse.http_path, columns, rows
```
AarushiShah-db, line 491: should we implement some sort of timeout here? how long does it usually take for it for return an error?

src/ucode/databricks.py
andy-xu-db: Currently if none of the warehouses are running, the code will start one up which works but can take minutes if not more. Should we allow that? If not then I'll remove the non-running candidate warehouses.

src/ucode/usage.py
andy-xu-db: Errors are usually returned within a second. The main concern is if there aren't any running sql warehouses, which will cause the fallback to start up a new one. If we don't bother with that then I can add a timeout of a minute or two minutes.

AarushiShah-db (dismissed):

AarushiShah-db (dismissed):

AarushiShah-db (dismissed):

AarushiShah-db (approved):

PR #276, opened 2026-08-06
Apply managed agent config per-agent and honor default model

asujithan:
### Changes
Makes an admin-published managed coding-agent config work for every agent, not just Claude Code, and fixes three bugs found while testing it. All behavior stays behind `ENABLE_MANAGED_AGENT_CONFIG`; with it unset the launch path is unchanged.

1. **Per-agent model shapes.** `managed_state_overrides` translates the manifest's model list into the shape each agent reads: provider-bucketed for opencode, per-agent keys for pi and copilot. Previously every non-claude agent got a flat list, which crashed opencode and was silently ignored by pi and copilot.
2. **pi and copilot get their own state keys** (`pi_models`, `copilot_models`) instead of writing the shared `claude_models`/`codex_models`/`gemini_models`. Those are workspace-wide discovery that claude, codex, and gemini also read, so one agent's managed policy could change what another launched.
3. **The admin's `default_model` now survives the whole session.** Non-claude agents applied it once at launch and then immediately overwrote it: pi/copilot's `launch()` and the 30-minute token refresh re-derive the model from state, which never held it, so they fell back to the allowlist's first entry. `resolve_state` now overlays `<tool>_default_model` and each agent prefers it. Claude was already correct – it pins the model as `ANTHROPIC_MODEL` and never re-derives. Family slots are unaffected: an unset family stays unset rather than inheriting `default_model`.
4. **Blocking error for agents the admin didn't enable.** `enabled_agents` is an allowlist; launching outside it ran fully unmanaged. Checked before model discovery so a rejected launch fails fast instead of after a ~19s sweep.
5. **Model discovery is skipped when the managed config already names models,** removing that sweep from most managed launches.
6. **The managed config is fetched before discovery instead of read from cache.** The allowlist check and the skip-discovery decision are now made from the authoritative config; previously a stale local copy could suppress discovery for a config that no longer named models. `--skip-preflight` still reads the cache by design. This replaces `managed_launch_state` with a fetch plus a separate `resolve_state`, since the manifest has to be layered over post-discovery state.
7. **`classify_model_family`** extracted so a model named in a config buckets exactly as discovery would.

### Test Plan
- uv run pytest --> 1257 passed, 36 skipped
- ruff check + ruff format clean
- added new unit tests all passing

https://github.com/user-attachments/assets/10fea9a2-cc44-4b0f-986b-1426ed8f7361

### DIFF
--- a/src/ucode/agents/pi.py
+++ b/src/ucode/agents/pi.py
@@ -42,8 +42,10 @@ from ucode.config_io import (
     write_json_file,
 )
 from ucode.databricks import (
+    ANTHROPIC_FAMILIES,
     TOKEN_REFRESH_INTERVAL_SECONDS,
     build_pi_base_urls,
+    classify_model_family,
     get_databricks_token,
 )
 from ucode.state import mark_tool_managed, save_state
@@ -171,13 +173,19 @@ def write_tool_config(
             state["workspace"], state.get("profile"), force_refresh=force_refresh
         )
     pi_base_urls = state.get("base_urls", {}).get("pi") or build_pi_base_urls(state["workspace"])
+    managed_families = _managed_model_families(state)
+    claude_models, codex_models, gemini_models = managed_families or (
+        state.get("claude_models") or {},
+        state.get("codex_models") or [],
+        state.get("gemini_models") or [],
+    )
     overlay, managed_keys = render_overlay(
         model,
         token,
         pi_base_urls,
-        state.get("claude_models") or {},
-        state.get("codex_models") or [],
-        state.get("gemini_models") or [],
+        claude_models,
+        codex_models,
+        gemini_models,
     )
     existing = read_json_safe(PI_CONFIG_PATH)
     providers = existing.get("providers")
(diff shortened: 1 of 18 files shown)


src/ucode/agents/pi.py
```diff
@@ -205,8 +213,42 @@ def _write_settings(model_selector: str) -> None:
     write_json_file(PI_SETTINGS_PATH, merged)
 
 
+def _managed_model_families(state: dict) -> tuple[dict[str, str], list[str], list[str]] | None:
+    """Split a managed config's ``pi_models`` into the per-family inputs Pi's providers need.
+
+    Pi builds one provider block per family, so a flat list has to be classified back out. Returns
+    None when no managed models are set, leaving the workspace-wide discovery lists in play.
+    """
+    managed = state.get("pi_models")
```
AarushiShah-db, line 223: see isaac comment:

pi._managed_model_families returns a truthy all-empty tuple, wiping every provider

The docstring says it returns None when no managed models are set, so managed_families or (…discovery…) falls back. But it also returns ({}, [], []) whenever pi_models is non-empty and nothing in it classifies — and that tuple is truthy, so it wins the or:

  pi._managed_model_families({"pi_models": ["system.ai.kimi-k2-7-code"]})
  → ({}, [], [])          # truthy

  classify_model_family returns "oss" for kimi/glm, and pi has no OSS provider, so it's dropped. Same for any family this build doesn't recognize. render_overlay then writes zero providers, while write_tool_config has already popped all the stale databricks-* provider blocks — so ~/.pi/agent/models.json ends up with a pinned model and no provider that can serve it. The developer's discovered models are discarded too, so there's no fallback. Fix: return None when all three buckets come back empty.

src/ucode/managed_resolve.py
```diff
@@ -57,37 +58,69 @@ def _agent_model_config(managed: dict, tool: str) -> dict[str, object]:
...
+
+
+def _bucket_by_provider(models: list[str]) -> dict[str, list[str]]:
+    """Group model FQNs into OpenCode's provider buckets, mirroring how discovery builds them.
+
+    Discovery derives these from the per-family lists (claude -> anthropic, and gemini/oss as-is), so
+    the same family classification recovers them from a flat manifest list. Models whose family
+    can't be identified are dropped.
+    """
+    buckets: dict[str, list[str]] = {}
```
AarushiShah-db, line 137: Same all-empty-bucket issue as pi._managed_model_families, but here it's worse because {} gets written into state. opencode has no OpenAI provider block (PROVIDER_KEYS is anthropic/google/oss), so models:["system.ai.gpt-5"] classifies as codex, gets dropped, and overrides["opencode_models"] = {} — which replaces the developer's discovered buckets, leaving opencode with no models at all from a config that named a real one.

Can we skip the override when every bucket comes back empty (and ideally warn that the models aren't servable by opencode)? test_unclassifiable_models_are_dropped_from_buckets currently asserts the {}, so it'd need updating.

AarushiShah-db (dismissed):

AarushiShah-db (approved):

PR #280, opened 2026-08-06
Route bare ucode by managed config and warn on ucode configure

asujithan:
### Changes
Makes bare ucode the developer entry point under a managed config, and deprecates ucode configure. All behavior stays behind ENABLE_MANAGED_AGENT_CONFIG; with it unset the CLI is unchanged (bare ucode prints help).

- Bare ucode launches the managed default agent. Fetches the workspace's config, shows a summary of what the admin set, and launches default_agent (falling back to the first enabled agent). Reuses refresh_managed_config + resolve_state — no new reconciliation logic.
- When no managed config exists, ucode routes by role. A workspace admin is pointed at ucode setup; everyone else is told to ask an admin.
- Admin detection via SCIM Me group membership (is_workspace_admin). Returns bool | None so an unreachable SCIM reads as "unknown" rather than misreporting a real admin as a non-admin. Matches the shape in #270/#271 so the two don't conflict.
- ucode configure is deprecated. With a managed config it errors and points at ucode; without one it warns and runs the existing flow unchanged.
- Config summary box shows the workspace, launching agent, enabled agents, provider, model, MCPs, skills, and the budget policy's tiers. MCPs/skills are marked (pending) — they are listed as configured but not yet registered locally.
- Flag semantics for the managed path. --dry-run uses the last saved managed-state.json and prints rather than writes; --skip-preflight launches fully unmanaged on your own settings. Help text keeps the existing non-managed meaning first.
- One config fetch per launch. Bare ucode hands the config it already read to _launch_tool instead of letting it refetch — previously 2 network reads per launch, which also printed any fallback warning twice.  

### Not in this PR
MCP/skill registration into state.json, and launching on the RecommendModel agent/model (bare ucode uses default_agent, so budget tiers are shown but not yet acted on) --> will be added in follow up pr.

### Test plan
uv run pytest — 1293 passed, 36 skipped. ruff check and ruff format --check clean; ty 117 vs. 118 on main. 19 new tests:

- ucode configure gating — blocks with a managed config, warns without one, silent when the env var is off.
- Config summary — tier rows, MCP/skill rows, "none configured" when empty, no policy rows without a budget policy.
- Bare ucode — launches default_agent; falls back to the first enabled agent; admin vs. non-admin guidance; --dry-run uses the cache and does not fetch; --skip-preflight explains it has no config to pick an agent from; prints help when the env var is off; subcommands still work (the callback must not intercept ucode status).
- Admin detection — in/not-in the admins group, None when SCIM is unreachable, False when the payload names no groups.
**New `ucode` command displays config summary + launches**
note if you run ucode configure in workspace that has a managed config you get a blocking error.

https://github.com/user-attachments/assets/07d53a50-84ff-4a97-ad09-e24cdc32ffeb

**Existing `ucode configure` command runs as normal when you're in a workspace without a managed config**
note there's a warning that this command is being deprecated.

https://github.com/user-attachments/assets/8c1471f8-2706-4943-a970-45787d6f0859

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -189,7 +189,7 @@ def _reject_configure_under_managed_config() -> None:
     """Refuse ``ucode configure`` when the workspace publishes a managed config.
 
     Configuring locally would be overridden at launch anyway, so it is an error rather than a
-    silently-ignored run. Without a managed config the command still works, but is deprecated.
+    silently-ignored run. Without a managed config the command still runs unchanged.
     """
     if not managed_agent_config_enabled():
         return
@@ -198,7 +198,6 @@ def _reject_configure_under_managed_config() -> None:
             "The ucode configure command is being deprecated. Please run `ucode` to launch "
             "with your admin's managed config applied"
         )
-    print_warning("The ucode configure command is being deprecated")
 
 
 def _print_discovery_diagnostics(state: dict) -> None:
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -49,6 +49,7 @@ from ucode.databricks import (
     get_databricks_token,
     install_databricks_cli,
     is_model_provider_feature_unavailable,
+    is_workspace_admin,
     list_profile_entries,
     list_tool_provider_services,
     normalize_workspace_url,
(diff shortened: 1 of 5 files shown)


AarushiShah-db (dismissed):
lgtm can we just remove the "deprecated" warning for now - because technically we still need the user to run ucode configure whenever they want to change/update the workspace right?

AarushiShah-db (approved):

PR #308, opened 2026-08-10
configure: add --skip-unavailable so --agents can configure the available subset

Edwinhe03:
## Problem

`--agents` treats an explicit list as all-or-nothing. If any named agent isn't available on the workspace, `configure_workspace_command` raises before configuring *any* of them:

```python
if unavailable_tools:
    _print_discovery_diagnostics(state)
    raise RuntimeError(f"Requested agent(s) not available on this workspace: {displays}.")
```

So on a workspace whose AI Gateway exposes no OpenAI models (`codex discovery: no endpoint exposes api_type openai/v1/responses`):

```bash
ucode configure --profiles DEFAULT --agents claude,codex,pi --use-pat --skip-validate --skip-upgrade
```

configures **nothing** and exits 1 — even though `claude` and `pi` are both usable there (`pi` needs only one of claude/codex/gemini). For CI running one command across heterogeneous workspaces, one unavailable model family takes down the whole setup.

## Why not just make `--agents` fail open

The default stays strict, because the two directions aren't symmetric:

- **Strict default:** a caller who wants tolerance opts in with a flag.
- **Fail-open default:** a caller who wants a *guarantee* has no way to get it back. `--agents claude,codex` exiting 0 having configured only claude leaves the pipeline believing codex works — the failure resurfaces later at `ucode codex` launch time, far from the cause.

Naming agents explicitly is a request for a specific outcome, so it should keep meaning "all of these, or tell me why not."

## Change

Add an opt-in `--skip-unavailable`:

```bash
ucode configure --agents claude,codex,pi --skip-unavailable
```

- Configures the available subset, preserving the order given in `--agents`.
- Skips the rest with a warning (`Skipping agent(s) not available on this workspace: Codex.`), on top of the existing per-source discovery diagnostics.
- Still exits non-zero when *none* of the requested agents are available — the pre-existing `if not available_on_workspace: return 1` covers this, since that list is built only from `selected_tools`. The flag can't turn a completely unusable workspace into a silent success.
- Requires `--agents`: the interactive picker already offers only available agents, and `--agent` names a single agent whose absence is the whole answer.
- The strict error now points at the flag: `... Pass --skip-unavailable to configure the available ones instead.`

Behavior with the flag absent is unchanged.

## Tests

Added to `tests/test_cli.py` (the existing `test_unavailable_selected_tool_errors_before_configure` still guards the strict default):

- `test_skip_unavailable_configures_available_subset` — the azure/no-GPT case: requesting `claude,codex,pi` configures `["claude", "pi"]`, installs only those binaries, and warns about Codex.
- `test_skip_unavailable_still_fails_when_none_available` — returns 1, never calls `configure_selected_tools`.
- `test_strict_error_mentions_skip_unavailable` — discoverability of the flag from the error.
- `test_skip_unavailable_requires_agents` / `_forwarded_with_agents` / `_absent_by_default` — CLI wiring.

`uv run pytest`: 1690 passed, 1 failed. The one failure (`test_managed_wizard.py::TestCliWiring::test_successful_apply_exits_zero`) reproduces on pristine `origin/main` in a clean worktree with none of these changes — pre-existing and unrelated. `uv run ruff check .` and `ruff format --check` are clean.

## Docs

README gets the strict-vs-tolerant distinction next to `--agents`, plus a quick-reference row.

### DIFF
--- a/src/ucode/agents/codex.py
+++ b/src/ucode/agents/codex.py
@@ -355,6 +355,14 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
     return state
 
 
+def _is_gpt_family(model: str) -> bool:
+    """Return True if this id is in the GPT family (versioned or OSS variants)."""
+    tail = model.split("/")[-1]
+    if tail.startswith("system.ai."):
+        tail = tail[len("system.ai.") :]
+    return tail.startswith("gpt-")
+
+
 def default_model(state: dict) -> str | None:
     """Pick the best available codex model.
 
@@ -380,10 +388,12 @@ def default_model(state: dict) -> str | None:
 
         return max(parsed, key=_gpt_version_key)[0]
 
-    # No versioned GPT found — fall back to the first available model. Any id in
-    # codex_models comes from UC model-services' codex bucket and exposes the
-    # responses API, so it's routable (e.g. system.ai.gpt-oss-120b).
-    return codex_models[0] if codex_models else None
+    # No versioned GPT found. Fall back to the first GPT-family id (gpt-*
+    # after stripping the system.ai. prefix). gpt-oss-* models are confirmed
+    # routable through the responses API; non-GPT ids (e.g. moonshotai/kimi-k2.5)
+    # would be rejected by the gateway, so they stay excluded.
+    gpt_family = [m for m in codex_models if _is_gpt_family(m)]
+    return gpt_family[0] if gpt_family else None
 
 
 def launch(state: dict, tool_args: list[str]) -> None:
--- a/src/ucode/agents/codex.py
+++ b/src/ucode/agents/codex.py
@@ -356,16 +356,14 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
 
 
 def default_model(state: dict) -> str | None:
-    """Pick the newest GPT model when multiple are available.
+    """Pick the best available codex model.
 
-    A managed config's ``codex_default_model`` takes priority. The discovery list
-    is alphabetically sorted, which can put "databricks-gpt-5" ahead of
-    "databricks-gpt-5-5". Prefer the highest semantic version instead.
-
-    Only GPT-parseable ids are considered. Codex routes the chosen ``model``
-    through the gateway as-is, so a non-GPT entry (e.g. ``moonshotai/kimi-k2.5``)
-    would be rejected with a Unity Catalog endpoint-name error. When no
-    candidate parses as GPT we return None rather than pinning an unroutable id.
+    A managed config's ``codex_default_model`` takes priority. Among versioned
+    GPT ids (e.g. ``system.ai.gpt-5``, ``system.ai.gpt-5-6-luna``) the highest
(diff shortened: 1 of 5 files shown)


Edwinhe03:
## Live-tested against a real workspace

Verified end-to-end (not just the mocked unit tests) against `ai-oss-ecosystem-integration-testing` with `--dry-run`: real OAuth, real AI Gateway verification, real model discovery. To reproduce the "GPT not allowed" workspace, GPT ids were filtered out of the model-services response and `discover_codex_models` was stubbed to return the real-world empty result `([], "no endpoint exposes api_type \`openai/v1/responses\`")` — everything else (auth, gateway, Claude/Gemini/OSS discovery from 51 live model-services ids) ran unmodified.

| Case | Result | Exit |
|---|---|---|
| `--agents claude,codex,pi --skip-unavailable`, no GPT | `Skipping agent(s) not available on this workspace: Codex.` → **Claude Code + Pi configured** | 0 |
| `--agents claude,codex,pi`, no GPT (strict default) | `ERROR Requested agent(s) not available on this workspace: Codex. Pass --skip-unavailable…`, nothing configured | 1 |
| `--agents codex --skip-unavailable`, no GPT (none available) | `ERROR No coding agents are available on this workspace.`, nothing configured | see note |
| `--agents claude,codex,pi --skip-unavailable`, all available (control) | all three configured, no spurious skip | 0 |

Also confirms Pi configures correctly *without* codex — the `if codex_models` guard in `agents/pi.py` drops the `databricks-openai` provider rather than erroring.

## Pre-existing bug found while testing (not fixed here)

The "none available" case prints `ERROR No coding agents are available on this workspace.` but **exits 0**. Cause: `configure_workspace_command` returns `1` there, but all seven call sites in `configure()` discard the return value — nothing converts it into `typer.Exit(1)`. Reproduced on pristine `origin/main` with none of this branch's changes, so it predates this PR and affects the existing strict path too.

I've left it out to keep this PR scoped. Worth noting for CI users: with `--skip-unavailable`, a workspace where *no* requested agent is available currently exits 0 with an error on stderr. Happy to fix it here or in a follow-up — say which you'd prefer.

tt-le (dismissed):

src/ucode/agents/codex.py
```diff
@@ -356,32 +356,34 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
...
-        major, minor, patch, suffix = entry[1]
-        base_bonus = 1 if not suffix else 0
-        return (major, minor or 0, patch or 0, base_bonus)
+        return max(parsed, key=_gpt_version_key)[0]
 
-    return max(parsed, key=_gpt_version_key)[0]
+    # No versioned GPT found — fall back to the first available model. Any id in
+    # codex_models comes from UC model-services' codex bucket and exposes the
+    # responses API, so it's routable (e.g. system.ai.gpt-oss-120b).
+    return codex_models[0] if codex_models else None
```
Edwinhe03: @tt-le added this fallback for oss models for codex on azure

tt-le (dismissed):

AarushiShah-db (approved):

PR #313, opened 2026-08-11
launch: don't let a local provider service override a managed config

tt-le:
## The bug

You run `ucode` under a managed config whose manifest pins Codex to `system.ai.glm-5-2` with **no** `model_provider_service`. On a machine where you'd separately run `ucode configure` and set a provider service for Codex (`main.dev.openai_all`), the launch shows:

```
Provider: main.tien_le.openai_all
```

…and Codex starts on the provider's own model (`gpt-5.6-sol`) — **not** the admin's `glm-5-2`.

## Why

`_launch_tool` resolved the provider as `provider or get_provider_service(state, tool)` and only *overrode* it with the manifest's provider **when the manifest named one** (`cli.py`). When the manifest named a model but no provider, the developer's locally-configured provider service leaked in. Because a provider service suppresses Databricks model pinning (`if provider: resolved_model = None`), the admin's model was silently dropped and the agent was rerouted through the local provider — the managed config did **not** win.

## The fix

Settle the provider **once, before model discovery**, from the managed config alone:

- the manifest's provider, or
- an explicit `--provider` that agrees with it (a disagreement is still a hard error, now raised earlier / fail-fast), or
- otherwise **none** — so the managed model gets pinned.

Only when there is **no** managed config do we fall back to the developer's configured provider, exactly as before.

## Behavior-change note

An agent enabled by a managed config that names **neither** a model nor a provider will now launch on discovered Databricks models rather than the developer's local provider service. This is intentional — the managed config is authoritative — but calling it out in case a narrower rule (only suppress the local provider when the manifest names a model) is preferred.

## Testing

- New `TestManagedProviderResolution` in `tests/test_cli.py`: local provider is ignored and the managed model is pinned; a managed provider is used; an explicit `--provider` conflicting with the manifest is rejected. Mutation-verified (the key test fails without the fix — `provider` came through as the leaked local service).
- Full non-e2e suite green; `ruff check` / `ruff format` clean.

Independent of #309 (that only consolidated the managed-config file); this is launch-time provider resolution.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1486,12 +1486,9 @@ def _launch_tool(
         if needs_auto_configure:
             _auto_configure_tool(tool)
         state = ensure_provider_state(tool)
-        # Remembered before the fallback below collapses the two cases: a managed config may not
-        # silently override a provider the user typed on the command line (it errors instead).
+        # Remembered so a managed config may not silently override a provider the user typed on the
+        # command line (it errors instead), and so the fallback below can distinguish the two.
         explicit_provider = provider
-        # An explicit --provider overrides the persisted choice; otherwise fall
-        # back to whatever `ucode configure` saved for this tool.
-        provider = provider or get_provider_service(state, tool)
         routing_agent = _ROUTING_AGENTS.get(tool)
         # Fetched before `configure_shared_state` because it decides whether this agent may launch
         # at all and whether the model discovery below can be skipped.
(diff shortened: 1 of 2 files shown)


tt-le:
Closing on reflection: a managed config that enables an agent without naming a provider may legitimately intend the developer's own configured provider to apply, so the local-provider fallback isn't necessarily a bug. Revisiting whether/how to scope this before changing launch behavior.

PR #314, opened 2026-08-11
cli: remove the `ucode` (launch) and `ucode setup` --dry-run flags

tt-le:
## What

Removes two `--dry-run` flags:

1. **`ucode --dry-run`** (the launch flag). It launched the managed agent from the last saved `managed-state.json` without fetching and wrote no config files — a third, confusing meaning of `--dry-run` that could run a launch off a stale local copy. Bare `ucode` now always refreshes the managed config from the workspace (the source of truth) before launching.
2. **`ucode setup --dry-run`**. It "walked the flow without writing any files", but the authoring flow's only writes are the local manifest (already covered by `ucode setup show` / `--from-file`) and a confirmed delete of an existing published config — so a dedicated dry-run added little.

`ucode configure --dry-run` (preview config files) is **unchanged** and still the one dry-run that remains.

## Changes

- **Launch:** drop the flag + `set_dry_run` + `dry_run` threading from the top-level callback and `_launch_managed_default` (now always `refresh_managed_config`); remove the now-dead launch-path `is_dry_run()` guards and the unused `is_dry_run` import in `cli.py`.
- **Setup:** drop the flag + `set_dry_run` from the `setup` callback; remove the now-dead `is_dry_run()` branches in `setup_command`, `setup_from_file`, and the delete-confirmation path in `managed_wizard.py`, plus that module's now-unused `is_dry_run` import.
- Revert the #309 docs that pointed at `ucode --dry-run` as the local try-out step (setup Next steps, README, `managed_config`/`managed_wizard` docstrings), and drop the `ucode setup --dry-run` README example.
- Update/remove the affected tests. `set_dry_run` stays (used by `configure`), and `save_managed_state` keeps respecting the global dry-run flag like the other config writers (still exercised by its unit test and `configure --dry-run`).

## Behavior

- `ucode --dry-run` and `ucode setup --dry-run` → unknown option; not listed in their `--help`.
- `ucode configure --dry-run` still works.

## Testing

- Full non-e2e suite green (1732); `ruff check` / `ruff format` clean.

Off current `main` (includes #309).

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,10 +186,9 @@ skipped. It then offers tracing, managed MCP servers, skills, and a spend-based 
 switches the default agent and model as the workspace burns through a budget.
 
 The result is written to `~/.ucode/managed-state.json` — the one local managed-config file — which
-`ucode apply` publishes to the workspace. Because a launch reads that same file, `ucode --dry-run`
-tries the authored config on this machine before it is published, without fetching or overwriting
-it. Your own agent configs are left alone, with one exception: answering yes to tracing, MCP
-servers, or skills runs the matching `ucode configure` step, which does configure this machine.
+`ucode apply` publishes to the workspace. Your own agent configs are left alone, with one exception:
+answering yes to tracing, MCP servers, or skills runs the matching `ucode configure` step, which
+does configure this machine.
 
 ```bash
 # Review the manifest and the exact payload `ucode apply` would publish.
@@ -197,9 +196,6 @@ ucode setup show
 
 # Walk the flow without writing anything.
 ucode setup --dry-run
-
-# Try the authored config locally before publishing (no fetch, no overwrite).
-ucode --dry-run
 
 # Skip the prompts and load a hand-written config instead (validated before saving).
 ucode setup --from-file ./managed-config.json
--- a/README.md
+++ b/README.md
@@ -194,9 +194,6 @@ does configure this machine.
 # Review the manifest and the exact payload `ucode apply` would publish.
 ucode setup show
 
-# Walk the flow without writing anything.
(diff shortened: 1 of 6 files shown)


src/ucode/cli.py
```diff
@@ -33,7 +33,7 @@
 )
 from ucode.agents.codex import revert_legacy_shared_config
 from ucode.agents.pi import PI_SETTINGS_BACKUP_PATH, PI_SETTINGS_PATH
-from ucode.config_io import is_dry_run, restore_file, set_dry_run
+from ucode.config_io import restore_file, set_dry_run
```
lilly-luo, line 36: does set_dry_run still need to be imported?

README.md
```diff
@@ -198,9 +197,6 @@ ucode setup show
 # Walk the flow without writing anything.
 ucode setup --dry-run
```
lilly-luo: remove this?

src/ucode/cli.py
tt-le: yeah still used by `configure` and `setup`

README.md
tt-le: ucode setup still has the `--dry-run` flag, it walks through the setup flow without writing to the `managed-state.json`

src/ucode/cli.py
AarushiShah-db: I think we should just remove the dry run feature altogether? it doesn't seem to provide much help tbh

AarushiShah-db (approved):

src/ucode/cli.py
tt-le: done

PR #315, opened 2026-08-12
Filter budget picker to per-user hard blocks

AarushiShah-db:
The workspace-metrics/budgets listing now exposes each alert's action_configurations, so list_workspace_budgets checks for a per-user threshold that carries a BLOCK_USAGE action rather than just any per-user alert. An email-only per-user alert doesn't enforce spend routing, so the managed-config picker now hides those too instead of leaning on the server to reject them at config-create time.

  ▎ Also: the admin config summary now shows both the tracked budget's name and the policy name,
  ▎ instead of collapsing to just the policy name. The budget name is remembered locally at
  ▎ authoring time and isn't serialized to the API.

Co-authored-by: Isaac

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -409,29 +409,39 @@ def is_workspace_admin(workspace: str, token: str) -> bool | None:
 _WORKSPACE_BUDGETS_API_PATH = "/api/ai-gateway/v2/workspace-metrics/budgets"
 
 
-# Alert-config scope that carries a per-user threshold. A budget's coding-agent spend routing only
-# works when it has one: the gateway's `recommendModel` measures the caller's spend against a
-# per-user threshold, so a budget with only a shared (workspace-wide) alert reports no spend and
-# leaves every tier inert. The listing exposes `scope_type` but not the alert's action, so ucode can
-# only check for the scope's presence; the server enforces the (block) action on config create.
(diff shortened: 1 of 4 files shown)


src/ucode/databricks.py
```diff
@@ -409,29 +409,39 @@ def is_workspace_admin(workspace: str, token: str) -> bool | None:
...
-# per-user threshold, so a budget with only a shared (workspace-wide) alert reports no spend and
-# leaves every tier inert. The listing exposes `scope_type` but not the alert's action, so ucode can
-# only check for the scope's presence; the server enforces the (block) action on config create.
+# A budget's coding-agent spend routing only works when it has a per-user alert threshold that also
+# hard-blocks: the gateway's `recommendModel` measures the caller's spend against a per-user
+# threshold (a shared, workspace-wide alert reports no per-user spend and leaves every tier inert),
+# and the routing policy is only enforced when that threshold carries a `BLOCK_USAGE` action rather
+# than an email notification alone. The listing exposes each alert's `scope_type` and its
+# `action_configurations`, so ucode can check both directly instead of leaning on the server to
+# reject unusable budgets at config-create time.
```
lilly-luo: can you rm all these commends

lilly-luo:
can you paste the output of the cli to in PR description to confirm it works?

lilly-luo (dismissed):

PR #319, opened 2026-08-12
Global settings claude codex

AarushiShah-db:
When an admin marks an agent machine-wide in ucode setup (use_as_global_settings), ucode now also writes the agent's
native config file so a bare claude/codex hits the gateway — not just ucode <agent>. It keeps writing the existing
ucode-private file too, so ucode <agent> is unchanged.

Only claude and codex qualify: their gateway auth self-refreshes, so the native file keeps working indefinitely. Other
agents bake a short-lived token with no bare-launch refresher (copilot has no native config file at all), so the
machine-wide prompt in ucode setup is now only asked for claude and codex.

ucode revert surgically prunes only ucode's keys from the native file — never deletes it, never touches the user's own
settings. Also fixes a latent hydrate_state bug that dropped the native tracking.
  
31 new tests; full suite green except one pre-existing environmental e2e failure. ruff + ty clean.

https://github.com/user-attachments/assets/bcd277b3-552e-4e23-943d-15f004fce20d

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -12,6 +12,7 @@ import socket
 import subprocess
 import sys
 import threading
+from collections.abc import Callable
 from pathlib import Path
 from typing import cast
 
@@ -21,6 +22,7 @@ from ucode.config_io import (
     ToolSpec,
     backup_existing_file,
     deep_merge_dict,
+    prune_key_paths,
     read_json_safe,
     write_json_file,
 )
@@ -42,6 +44,10 @@ from ucode.ui import print_err, print_note, print_success, print_warning
 CLAUDE_CONFIG_DIR = Path.home() / ".claude"
 CLAUDE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "ucode-settings.json"
 CLAUDE_BACKUP_PATH = APP_DIR / "claude-ucode-settings.backup.json"
+# Claude Code's own user-scope settings file, read by a bare `claude` with no flags. ucode writes
+# here (in addition to the private file above) only when the managed config sets
+# use_as_global_settings, so a developer who launches `claude` directly still hits the gateway.
+CLAUDE_NATIVE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "settings.json"
 
 SPEC: ToolSpec = {
     "binary": "claude",
@@ -516,43 +522,50 @@ def write_tool_config(
                 "to install the Claude Stop hook — traces won't be emitted. Re-run "
                 "`ucode configure tracing`."
             )
-
-    existing = read_json_safe(CLAUDE_SETTINGS_PATH)
-    merged = deep_merge_dict(existing, overlay)
-    # Drop any apiKeyHelper a prior non-relayed launch left in the file; relayed
-    # must not carry one (it would outrank the subscription OAuth).
-    if relayed:
-        merged.pop("apiKeyHelper", None)
-    if tracing_env_vars and stop_hook_command:
-        _upsert_tracing_stop_hook(merged, stop_hook_command)
-    if not tracing_env_vars:
-        env_block = merged.get("env")
-        if isinstance(env_block, dict):
-            for key in CLAUDE_TRACING_ENV_KEYS:
-                env_block.pop(key, None)
-        # Strip only ucode's tracing Stop hook so user hooks stay intact.
-        _remove_tracing_stop_hook(merged)
-    # Prune ucode-managed model env keys we deliberately don't write this run
-    # (e.g. ANTHROPIC_MODEL — see render_overlay).
-    overlay_env = overlay.get("env", {})
-    merged_env = merged.get("env")
-    if isinstance(merged_env, dict):
-        for key in CLAUDE_MANAGED_MODEL_ENV_KEYS:
-            if key not in overlay_env:
-                merged_env.pop(key, None)
-    # deep_merge_dict keeps keys already in the file, so drop the ones ucode no
-    # longer writes.
-    if isinstance(merged_env, dict):
-        for key in CLAUDE_REMOVED_ENV_KEYS:
-            merged_env.pop(key, None)
-    # Smart-routing hooks: install ucode's PreToolUse/SessionStart/
-    # SubagentStart hooks when routing is enabled (and not under a provider,
-    # which pins no Databricks model), else surgically strip only ucode's own.
+    # Smart-routing hooks: install ucode's PreToolUse/SessionStart/SubagentStart hooks when routing
+    # is enabled (and not under a provider, which pins no Databricks model), else surgically strip
+    # only ucode's own. Applied per file inside _compose_claude_settings.
     routing_enabled = smart_routing_enabled(state) and provider is None
-    sync_smart_routing_hooks(merged, state, enabled=routing_enabled)
     if routing_enabled:
         managed_keys = managed_keys + [["hooks", event] for event in CLAUDE_ROUTING_HOOK_EVENTS]
-    write_json_file(CLAUDE_SETTINGS_PATH, merged)
+
+    def _compose(base: dict) -> dict:
+        # deepcopy the overlay per file so merging into one base can't alias nested dicts into
+        # the other (deep_merge_dict grafts overlay's own dict objects onto a base missing the key).
+        merged = deep_merge_dict(base, copy.deepcopy(overlay))
+        # Drop any apiKeyHelper a prior non-relayed launch left in the file; relayed
+        # must not carry one (it would outrank the subscription OAuth).
+        if relayed:
+            merged.pop("apiKeyHelper", None)
+        if tracing_env_vars and stop_hook_command:
+            _upsert_tracing_stop_hook(merged, stop_hook_command)
+        if not tracing_env_vars:
+            env_block = merged.get("env")
+            if isinstance(env_block, dict):
+                for key in CLAUDE_TRACING_ENV_KEYS:
+                    env_block.pop(key, None)
+            # Strip only ucode's tracing Stop hook so user hooks stay intact.
+            _remove_tracing_stop_hook(merged)
+        # Prune ucode-managed model env keys we deliberately don't write this run
+        # (e.g. ANTHROPIC_MODEL — see render_overlay).
+        overlay_env = overlay.get("env", {})
+        merged_env = merged.get("env")
+        if isinstance(merged_env, dict):
+            for key in CLAUDE_MANAGED_MODEL_ENV_KEYS:
+                if key not in overlay_env:
+                    merged_env.pop(key, None)
+            # deep_merge_dict keeps keys already in the file, so drop the ones ucode no
+            # longer writes.
+            for key in CLAUDE_REMOVED_ENV_KEYS:
+                merged_env.pop(key, None)
+        sync_smart_routing_hooks(merged, state, enabled=routing_enabled)
+        return merged
+
+    write_json_file(CLAUDE_SETTINGS_PATH, _compose(read_json_safe(CLAUDE_SETTINGS_PATH)))
+
+    native_configs = None
+    if state.get("write_native_config"):
+        native_configs = _write_native_settings(_compose, managed_keys, relayed)
 
     if web_search_model:
         _register_web_search_mcp(state["workspace"], web_search_model, state.get("profile"))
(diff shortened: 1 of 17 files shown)


AarushiShah-db:
btw @lilly-luo we can only support claude/codex for global settings right now because the other agents don't have auth self-refresh capabilities ( the token would expire within 1 hr) so the user needs to continue using ucode for the other ones

src/ucode/agents/claude.py
```diff
@@ -42,6 +44,10 @@
 CLAUDE_CONFIG_DIR = Path.home() / ".claude"
 CLAUDE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "ucode-settings.json"
 CLAUDE_BACKUP_PATH = APP_DIR / "claude-ucode-settings.backup.json"
+# Claude Code's own user-scope settings file, read by a bare `claude` with no flags. ucode writes
+# here (in addition to the private file above) only when the managed config sets
+# use_as_global_settings, so a developer who launches `claude` directly still hits the gateway.
+CLAUDE_NATIVE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "settings.json"
```
lilly-luo: `/etc/claude-code/managed-settings.json` (which is what isaac sets) will override this ?

src/ucode/databricks.py
```diff
@@ -409,29 +409,39 @@ def is_workspace_admin(workspace: str, token: str) -> bool | None:
...
 
 
-def _has_per_user_alert(entry: dict) -> bool:
-    """Whether a raw budget entry carries a per-user alert threshold."""
+def _has_per_user_block(entry: dict) -> bool:
+    """Whether a raw budget entry has a per-user alert threshold that hard-blocks usage.
+
+    True only when some alert is per-user scoped *and* carries a ``BLOCK_USAGE`` action; a per-user
+    alert with only an email notification does not enforce spend routing.
+    """
```
lilly-luo, line 421: i think this is already in master, can you rebase?

src/ucode/agents/codex.py
```diff
@@ -354,11 +354,89 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
         enabled=smart_routing_enabled(state) and provider is None,
     )
     write_toml_file(CODEX_CONFIG_PATH, doc)
-    state = mark_tool_managed(state, "codex", MANAGED_KEYS)
+    # use_as_global_settings: also write the modern overlay to the native ~/.codex/config.toml so a
+    # bare `codex` (no `--profile ucode`) defaults to the gateway. Written last, after
```
lilly-luo: instead of writing to ~/.codex/config.toml, should we instead write to `/etc/codex/managed_config.toml`?  https://learn.chatgpt.com/docs/enterprise/managed-configuration 

Image

src/ucode/agents/claude.py
lilly-luo: see https://code.claude.com/docs/en/settings#settings-files

src/ucode/agents/claude.py
```diff
@@ -564,11 +574,97 @@ def write_tool_config(
...
 
+def _write_managed_settings(
+    compose: Callable[[dict], dict], managed_keys: list[list[str]], relayed: bool
+) -> list[dict] | None:
+    """Write ucode's config into Claude Code's OS managed-settings.json so a bare `claude` works.
+
+    Runs only under use_as_global_settings. The managed file is root-owned and the highest-precedence
+    scope, so it applies whether or not `ucode` launches `claude`. The same compose (merge overlay +
+    prune stale keys) that produced the private file is applied to the existing managed file, so any
+    real IT-authored keys already there survive. The write goes through the isaac-style sudo path
```
lilly-luo: can u rm isaac from the comments

src/ucode/managed_files.py
```diff
@@ -0,0 +1,173 @@
...
+            f"{display}: machine-wide managed settings aren't supported on this platform; "
+            f"skipped {path}."
+        )
+        return "skipped"
+    # Drift check first — reading is unprivileged, so an unchanged file never triggers a sudo prompt.
+    if _read_existing(path) == desired_text:
+        return "unchanged"
+    if is_dry_run():
+        console.print(f"\n[bold]\\[dry run] {path} (via sudo)[/bold]\n{desired_text}")
+        return "written"
```
lilly-luo, line 91: i thought we're removing dry run for now?

src/ucode/agents/claude.py
```diff
@@ -516,43 +519,50 @@ def write_tool_config(
...
+            for key in CLAUDE_REMOVED_ENV_KEYS:
+                merged_env.pop(key, None)
+        sync_smart_routing_hooks(merged, state, enabled=routing_enabled)
+        return merged
+
+    write_json_file(CLAUDE_SETTINGS_PATH, _compose(read_json_safe(CLAUDE_SETTINGS_PATH)))
+
+    managed_descriptors = None
+    if state.get("write_managed_config"):
+        managed_descriptors = _write_managed_settings(_compose, managed_keys, relayed)
```
lilly-luo: so does this mean we write to both ~/.claude/settings.json AND /etc/claude-code/managed-settings.json for managed config? i think we should only write to the /etc/claude-code/managed-settings.json if it's managed config

src/ucode/cli.py
```diff
@@ -999,6 +1000,13 @@ def revert() -> int:
     # Older Codex (< 0.134.0) had ucode edit the shared ~/.codex/config.toml in
     # place; restoring the per-profile file above does not undo that.
     legacy_codex_stripped = revert_legacy_shared_config()
+    # OS managed settings files written under use_as_global_settings (the highest-precedence config a
+    # bare `claude` / `codex` reads): surgically strip ucode's keys via sudo, never touching other
+    # keys. Runs before clear_state so the tracked descriptors are still available.
+    managed_reverts = {
+        "claude": claude_agent.revert_managed_config(state),
+        "codex": codex_agent.revert_managed_config(state),
+    }
```
lilly-luo: do we have to support reverts?

src/ucode/agents/claude.py
```diff
@@ -564,11 +574,97 @@ def write_tool_config(
...
+        # `sync_smart_routing_hooks` / `_upsert_tracing_stop_hook` in `write_tool_config`). Only plain,
+        # ucode-owned key paths go to `prune_key_paths`.
+        plain_keys: list[list[str]] = []
+        touches_routing_hooks = False
+        touches_tracing_stop_hook = False
+        for key in keys:
+            if len(key) == 2 and key[0] == "hooks" and key[1] in CLAUDE_ROUTING_HOOK_EVENTS:
+                touches_routing_hooks = True
+            elif len(key) == 2 and key[0] == "hooks" and key[1] == "Stop":
+                touches_tracing_stop_hook = True
```
lilly-luo: im pretty worried about maintaining these...it feels pretty brittle to rely on the order of hook events. can we just save a copy of the managed config beforehand and upsert it instead of selectively pruning?

src/ucode/agents/codex.py
```diff
@@ -354,11 +357,110 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
...
+    # use_as_global_settings: also write the modern overlay to Codex's OS managed config
+    # (/etc/codex/managed_config.toml), the highest-precedence scope a bare `codex` reads — so it
+    # defaults to the gateway without `--profile ucode`. codex auth self-refreshes via
+    # `ucode auth-token`, so the file keeps working. The write goes through the isaac-style sudo path.
+    managed_descriptors = None
+    if state.get("write_managed_config"):
+        managed_descriptors = _write_managed_config(
+            workspace, chosen_model, databricks_profile, bool(state.get("use_pat")), provider
+        )
+    state = mark_tool_managed(state, "codex", MANAGED_KEYS, native=managed_descriptors)
```
lilly-luo: same comment about only updating the managed config vs the local settings instead of both

src/ucode/agents/codex.py
```diff
@@ -354,11 +357,110 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
...
+def _managed_config_path() -> Path | None:
+    """OS-level Codex managed config file, or None on unsupported platforms.
+
+    Linux and macOS both use ``/etc/codex/managed_config.toml`` (root-owned, highest precedence);
+    Windows uses ``~/.codex/managed_config.toml``. See
+    https://learn.chatgpt.com/docs/enterprise/managed-configuration.
+    """
+    if sys.platform == "darwin" or sys.platform.startswith("linux"):
+        return Path("/etc/codex/managed_config.toml")
+    if sys.platform.startswith("win"):
```
lilly-luo: can we just match on windows. also can we lowercase sys.platform?

src/ucode/agents/codex.py
```diff
@@ -354,11 +357,110 @@ def write_tool_config(state: dict, model: str | None = None, provider: str | Non
...
+    providers = doc.get("model_providers")
+    if isinstance(providers, dict) and CODEX_MODEL_PROVIDER_NAME in providers:
+        providers.pop(CODEX_MODEL_PROVIDER_NAME, None)
+        if not providers:
+            doc.pop("model_providers", None)
+        changed = True
+    return changed
+
+
+def revert_managed_config(state: dict) -> str | None:
```
lilly-luo: same comment on just saving what it was before and then upserting that, instead of stripping + pruning

src/ucode/managed_files.py
```diff
@@ -0,0 +1,173 @@
...
+        _report_sudo_failure(path, display, exc)
+        return "skipped"
+    return "written"
+
+
+def _sudo_replace(path: Path, desired_text: str) -> None:
+    """Replace ``path`` with ``desired_text`` via sudo (temp file → ``sudo cp``), handling immutability.
+
+    Writes the payload to a user-owned temp file first (no sudo), then copies it into place with
+    ``sudo`` and makes it world-readable — the same sequence isaac uses so the file it lays down is
```
lilly-luo: rm isaac from comments

src/ucode/managed_files.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        return "skipped"
+    except subprocess.CalledProcessError as exc:
+        _report_sudo_failure(path, display, exc)
+        return "skipped"
+    return "written"
+
+
+def _sudo_replace(path: Path, desired_text: str) -> None:
+    """Replace ``path`` with ``desired_text`` via sudo (temp file → ``sudo cp``), handling immutability.
```
lilly-luo, line 107: you're saying sudo cp doesn't need me to enter my password? 😂

src/ucode/managed_files.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    try:
+        # `path.exists()` stats the file; under a root-locked parent dir (e.g. a 750 /etc/codex we
+        # haven't opened yet) that raises PermissionError. There's nothing to unlock we can see, and
+        # the subsequent `sudo cp` (as root) overwrites regardless, so treat it as "nothing to clear".
+        if not path.exists():
+            return False
+    except OSError:
+        return False
+    if sys.platform == "darwin":
```
lilly-luo: can we have a set of enums for darwin, linux, etc?

src/ucode/managed_resolve.py
```diff
@@ -176,6 +176,23 @@ def managed_provider_service(managed: dict, tool: str) -> str | None:
...
+    """True when the admin marked ``tool`` machine-wide AND ``tool`` can support it.
+
+    ``use_as_global_settings`` means: also write the agent's OS-level managed settings file
+    (``/etc/claude-code/managed-settings.json``, ``/etc/codex/managed_config.toml``) so a bare
+    ``claude`` / ``codex`` picks up the gateway config. Only agents in
+    :data:`~ucode.agents.GLOBAL_SETTINGS_AGENTS` have such a file, so the flag is ignored for any
+    other agent — a hand-written ``--from-file`` config can't turn it on for an agent that has no
+    managed settings path.
+    """
+    from ucode.agents import GLOBAL_SETTINGS_AGENTS
```
lilly-luo: move import up

src/ucode/state.py
```diff
@@ -236,9 +243,31 @@ def clear_state() -> None:
...
+    ``use_as_global_settings`` — each ``{"path": str, "format": "json"|"toml", "keys": [...]}`` —
+    so ``ucode revert`` can surgically prune only ucode's keys from the user's shared file.
+
+    ``native=None`` means "this launch wrote no native file", not "clear the tracking": a later
+    launch that skips the native write (the admin unset ``use_as_global_settings``, a relayed Claude
+    launch, or a legacy-layout Codex launch) leaves ucode's keys sitting in the user's shared file, so
+    the descriptor from the launch that *did* write them must be preserved or ``ucode revert`` can no
+    longer find and prune them. ``ucode revert`` wipes state wholesale (``clear_state``), so a stale
+    descriptor never lingers past a revert.
+    """
```
lilly-luo: can you make this comment shorter

src/ucode/agents/claude.py
AarushiShah-db: no not settings.json, we write to ucode-settings.json and the managed-settings.json. We need to still write to the ucode file otherwise commands like "ucode claude" will fail

src/ucode/cli.py
AarushiShah-db: removed revert support for now to make it simpler

src/ucode/managed_files.py
AarushiShah-db: ucode configure --dry-run still exists

lilly-luo (approved):

PR #321, opened 2026-08-12
Add --skip-managed-config and decouple managed config from --skip-pre…

AarushiShah-db:
…flight

--skip-managed-config (on `ucode configure` and every launcher) ignores the workspace's managed coding-agent config for that run, behaving as though ENABLE_MANAGED_AGENT_CONFIG were unset — it clears the env var for the process, so the single managed_agent_config_enabled() gate short-circuits every managed path (fetch/apply, budget read, MCP registration, the configure reject flow, and the bare-`ucode` agent picker). Per-invocation only.

--skip-preflight no longer touches managed config: it now means only "skip the per-launch auth + AI Gateway re-validation." Dropped its cache-instead-of-refresh branch, its budget-read and MCP-registration gating, and the bare-`ucode` "no config to pick an agent" error.

Co-authored-by: Isaac

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1765,6 +1765,7 @@ SkipManagedConfigOption = Annotated[
         "--skip-managed-config",
         help="Ignore your workspace's managed coding-agent config for this run, as if managed "
         "configs were switched off — use your own local settings instead.",
+        hidden=True,
     ),
 ]
 
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1384,9 +1384,6 @@ def _fetch_budget_recommendation(state: dict, managed: dict | None) -> dict | No
     Enforcement is server-side, so a failed read only costs the recommendation: the config's own
     ``default_model`` still applies and the launch proceeds.
     """
-    # No managed config (feature off, --skip-managed-config, or none published) means no budget to
-    # read. --dry-run resolves the agent from the last saved config alone, so it must not reach the
-    # control plane — mirroring the managed-config read, which is likewise skipped under --dry-run.
     if managed is None or is_dry_run():
         return None
     reason: str | None = None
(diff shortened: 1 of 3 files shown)


src/ucode/cli.py
```diff
@@ -1379,15 +1378,16 @@ def _note_recommended_agent(recommendation: dict | None, tool: str) -> None:
...
+def _fetch_budget_recommendation(state: dict, managed: dict | None) -> dict | None:
     """The agent and model the caller's budget tier allows, or None when there is no budget to read.
 
     Enforcement is server-side, so a failed read only costs the recommendation: the config's own
     ``default_model`` still applies and the launch proceeds.
     """
-    if managed is None or skip_preflight:
+    # No managed config (feature off, --skip-managed-config, or none published) means no budget to
+    # read. --dry-run resolves the agent from the last saved config alone, so it must not reach the
+    # control plane — mirroring the managed-config read, which is likewise skipped under --dry-run.
```
lilly-luo: can you rm this comment

lilly-luo:
i don't understand why we need this - doesn't this undercut the purpose of the managed config by giving userse a way around it?

AarushiShah-db:
> don't understand why we need this - doesn't this undercut the purpose of the managed config by giving userse a way around it?
@lilly-luo we need a way to disable the feature for omniagent. omniagent requires `ucode configure` to be a non-interactive command, but once we roll this feature out 'ucode configure' will start to ask admins to run "ucode setup" instead (with an interactive prompt of yes or no)

lilly-luo:
> @lilly-luo we need a way to disable the feature for omniagent. omniagent requires ucode configure to be a non-interactive command, but once we roll this feature out 'ucode configure' will start to ask admins to run "ucode setup" instead (with an interactive prompt of yes or no)

i see. can we hide this command from the --help section then? ideally a normal ucode user never knows about it.

AarushiShah-db:
@lilly-luo updated so it doesnt show up in --help

lilly-luo (approved):

tt-le (approved):

PR #329, opened 2026-08-13
setup: let admins enter a custom model in the hosted-model pickers

AarushiShah-db:
Discovery only lists system.ai.* models, so an admin whose workspace serves a coding model somewhere else
  (main.default.claude-opus-4-5) had no way to pick it in ucode setup.

  Each Databricks-hosted model picker now shows the top few discovered models plus an "✎ Enter a custom model…" row. A
  typed catalog.schema.model is verified against the workspace's model services and re-prompted on a miss (a 404 means the
  catalog/schema doesn't exist → re-prompt; only transient errors like 5xx/network are accepted as typed, with a short
  warning). Multi-select agents (opencode, pi) can add several custom models; single-model agents take one.

  Custom ids are marked in model_config.custom_models so validation doesn't reject a model the discovered inventory
  legitimately doesn't contain. The marker is local to the manifest — it's dropped from the published proto, so the server
  just sees default_model and validates it itself.

  Tests cover the existence check (found / absent / 404 / transient / pagination), the picker's custom row + truncation,
  multi-add, and validation acceptance.

https://github.com/user-attachments/assets/9f9560fe-85fa-4c9d-9f4a-7401ed9d1bf4

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -1424,6 +1424,10 @@ def _model_service_id(service: dict) -> str | None:
 _MODEL_SERVICES_PAGE_SIZE = 100
 _MODEL_SERVICES_PAGE_RETRIES = 4
 
+# Substrings that mark a failure reason (`HTTP <code> <phrase>: <body>`) as a 404 / NOT_FOUND:
+# the HTTP status line and the Databricks `error_code` carried in the response body.
+_NOT_FOUND_REASON_MARKERS = ("http 404", "not_found")
+
 
 def _get_model_services_page(
     url: str, token: str, *, retries: int = _MODEL_SERVICES_PAGE_RETRIES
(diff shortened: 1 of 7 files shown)


src/ucode/databricks.py
```diff
@@ -1542,6 +1542,74 @@ def list_model_services(
...
+    normalized = f"{catalog}.{schema}.{_model}"
+    parent = f"schemas/{catalog}.{schema}"
+    hostname = workspace_hostname(workspace)
+    page_token: str | None = None
+    seen_tokens: set[str] = set()
+    for _ in range(max_pages):
+        params: dict[str, str] = {"parent": parent, "page_size": str(_MODEL_SERVICES_PAGE_SIZE)}
+        if page_token:
+            params["page_token"] = page_token
+        url = f"https://{hostname}/api/2.1/unity-catalog/model-services?{urlencode(params)}"
```
lilly-luo, line 1591: extract into const or util?

lilly-luo (dismissed):

PR #337, opened 2026-08-14
opencode: refresh Databricks token via loopback proxy (fixes ~1h mid-session 401s)

kishor-rkrishnan:
## Problem

Long-running `ucode opencode` sessions fail roughly every hour with `401 Invalid Token` and cannot recover without a restart. This is the OpenCode side of #190.

## Root cause

`ucode opencode` writes the Databricks OAuth token into `opencode.json` (the provider `Authorization` header + `apiKey`). OpenCode resolves that config **once at process start** and never re-reads it, so when the token reaches its ~1h OAuth lifetime every subsequent request 401s. The existing background thread (`_refresh_forever`) rewrites `opencode.json` every 30 min, but the running OpenCode process ignores the file after launch, so the rewrite has no effect.

(`ucode claude` is unaffected — it uses Claude Code's `apiKeyHelper`, which re-fetches the token on a 401.)

## Fix

Reuse the loopback refresh proxy already used for relayed Claude (`gateway_proxy`): point each OpenCode provider's `baseURL` at the proxy, which overwrites the `Authorization` header with a freshly-minted token on every request via its own background `_TokenCache` refresher. The token written into `opencode.json` becomes a bootstrap value the proxy replaces per request, so it never goes stale mid-session.

- **One proxy serves all providers** — the anthropic, gemini and mlflow paths all live under the same workspace host, so the proxy only swaps host + `Authorization` and forwards the path unchanged.
- `start_proxy` gains `upstream_base` and `swap_header` keyword arguments; their defaults reproduce the existing relayed-Claude behavior exactly (the unchanged header tests still pass), so this is **purely additive for the Claude path**.
- The obsolete per-launch file-rewrite refresher (`_refresh_forever`) is removed.

Scoped to this single behavior (the static-token mid-session death). The separate concurrent-refresh token-rotation race (#190 / #197) is intentionally out of scope here.

## Testing

- `uv run pytest` — unit suite passes (the only failures are pre-existing environment/e2e ones unrelated to this change).
- `uv run ruff check .` and `ruff format --check .` — clean.
- New unit tests: swap-header selection + upstream-base override (`test_gateway_proxy.py`); loopback URL rewrite (`test_agent_opencode.py`).
- Verified end-to-end against a real AI Gateway workspace with the real `opencode` binary: with an expired token in the config, requests sent straight to the gateway return **401** (bug reproduced); the same expired-token config pointed at the proxy returns **200** (fixed). Also confirmed the gateway authenticates on `Authorization` and tolerates the stale `x-api-key` OpenCode still sends, so swapping `Authorization` alone is sufficient.

### DIFF
--- a/src/ucode/agents/opencode.py
+++ b/src/ucode/agents/opencode.py
@@ -6,6 +6,7 @@ import os
 import signal
 import subprocess
 import threading
+from urllib.parse import urlsplit, urlunsplit
 
 from ucode.agent_updates import available_npm_package_update
 from ucode.config_io import (
@@ -17,7 +18,6 @@ from ucode.config_io import (
     write_json_file,
 )
 from ucode.databricks import (
-    TOKEN_REFRESH_INTERVAL_SECONDS,
     build_opencode_base_urls,
     get_databricks_token,
     model_token_limits,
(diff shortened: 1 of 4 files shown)


src/ucode/gateway_proxy.py
rohita5l, line 1: Instead of reusing this proxy which is meant for a different provider, can you please create an opencode specific proxy

rohita5l (changes requested):

src/ucode/gateway_proxy.py
kishor-rkrishnan: Done — moved this into a dedicated `opencode_proxy` module, so `gateway_proxy` is now untouched (identical to `main`). The new proxy is purpose-built for OpenCode (swaps the `Authorization` header, forwards to the workspace host). Thanks for the review!

PR #338, opened 2026-08-14
mcp-proxy: support both mcp 1.x and 2.x (drop the <2 cap) (#307)

### LINKED ISSUE #307: [Bug] Fresh ucode install pulls mcp 2.0 and mcp-proxy crashes on a removed API
## Summary

A fresh `ucode` installation resolved `mcp 2.0.0` because the package declares `mcp>=1.28.0`. `ucode mcp-proxy` then crashed during import because it uses `streamablehttp_client`, which is no longer available under that name in MCP 2.0.0. Every MCP server configured through the proxy consequently failed to connect.

The same incompatible combination remains possible on current `main` (`5f0d5ecdb8106d20f0d218bdd635b4bf707f4973`):

- [`pyproject.toml` permits `mcp>=1.28.0`](https://github.com/databricks/ucode/blob/5f0d5ecdb8106d20f0d218bdd635b4bf707f4973/pyproject.toml)
- [`mcp_proxy.py` imports `streamablehttp_client`](https://github.com/databricks/ucode/blob/5f0d5ecdb8106d20f0d218bdd635b4bf707f4973/src/ucode/mcp_proxy.py)

## Environment

- `ucode 0.0.0+174.g62c5536`
- `mcp 2.0.0` from the unconstrained dependency resolution
- Python 3.14
- Linux container
- Databricks CLI v1.11.0

## Steps to reproduce

```bash
uv tool install git+https://github.com/databricks/ucode

ucode mcp-proxy \
  --url https://<workspace>/ai-gateway/mcp-services/<service> \
  --host https://<workspace> \
  --profile <profile>
```

The failure occurs while importing the proxy, before the upstream MCP connection is established.

## Actual result

```text
File "ucode/cli.py", line 1015, in mcp_proxy_cmd
  from ucode.mcp_proxy import serve

File "ucode/mcp_proxy.py", line 33, in <module>
  from mcp.client.streamable_http import streamablehttp_client

ImportError: cannot import name streamablehttp_client from mcp.client.streamable_http
```

The coding agent reports that the MCP server failed to connect.

## Expected result

A clean installation should resolve a supported MCP SDK version and start `ucode mcp-proxy` successfully.

## Workaround

```bash
uv tool install --force --with "mcp<2" git+https://github.com/databricks/ucode
```

This resolved MCP `1.29.0`, after which the same proxy and governed GitHub MCP connection started successfully.

## Acceptance criteria

- Either constrain the dependency to `mcp<2` or support the MCP 2.x client API.
- Add a clean-install integration test that resolves the newest dependency version allowed by package metadata and starts `ucode mcp-proxy`.
- Surface an actionable compatibility error if an unsupported MCP SDK is installed.

## Additional context

This was reproduced during an autonomous end-to-end coding-agent governance test on August 9, 2026. The failure blocked the documented clean-install developer path until the dependency was manually downgraded.

sunishsheth2009:
## Summary

Fixes #307. A fresh `ucode` install resolved **mcp 2.0.0** off `mcp>=1.28.0`, and `ucode mcp-proxy` crashed on import — 2.x renamed `streamablehttp_client` → `streamable_http_client` and moved from **httpx** to **httpx2**. Every MCP server through the proxy then failed to connect.

PR #311 capped `mcp<2` to stop the bad resolution. **This PR instead makes the proxy work against both mcp 1.x and 2.x and removes the cap**, so a clean install on either major just works.

## How one code path covers both majors

The key realization: both mcp 1.28+ and 2.x export the **2.x-native** call shape `streamable_http_client(url, http_client=<AsyncClient>)` (1.x's `streamablehttp_client` is a deprecated shim over it). Only two things differ across the majors, and both are handled:

| | mcp 1.x | mcp 2.x | handling |
|---|---|---|---|
| HTTP lib | `httpx` | `httpx2` | `_httpx()` resolves whichever the SDK uses |
| client yield | `(read, write, get_session_id)` | `(read, write)` | `_run` takes `streams[0], streams[1]` |
| auth | `httpx.Auth` | `httpx2.Auth` | `_build_token_auth()` subclasses the resolved flavor |

- `_httpx()` imports `httpx2` (mcp 2.x) else `httpx` (mcp 1.x), via `importlib` so the type checker / CI — pinned to mcp 1.x, no httpx2 present — don't choke on a static `import httpx2`.
- `_build_token_auth()` builds the per-request bearer `Auth` from that flavor; `auth_flow`'s generator contract is identical in httpx and httpx2.
- **pyproject:** `mcp>=1.28.0,<2` → `mcp>=1.28.0`. `httpx` stays a direct dep (the 1.x path); `httpx2` arrives transitively with mcp 2.x, so no direct httpx2 dep is declared.

The fail-fast auth behavior from #201's follow-up is preserved unchanged.

## How do you know it works?

Verified in a real venv against **both** majors (not just the pinned one):

- **mcp 1.28.1** (pinned): 21 proxy tests pass; `ty check src/` clean; full suite 1788 passed.
- **mcp 2.0.0 + httpx2 2.9.1** (installed into the venv, `--no-sync`): the proxy **imports** (the #307 crash point is gone), `_httpx()` selects `httpx2`, and `httpx2.AsyncClient(auth=…)` + `streamable_http_client(http_client=…)` compose without error; 21 proxy tests pass; `ty check src/` clean.
- Confirmed the yield-arity difference empirically (1.28.1 yields a 3-tuple, 2.0.0 a 2-tuple) and that `streams[0], streams[1]` handles both.
- New/updated tests in `tests/test_mcp_proxy.py`: mcp dep is **uncapped**; `_httpx()` matches the installed SDK; the shared `streamable_http_client` symbol imports; the token `Auth` is an instance of the selected flavor's `Auth`.
- The only 2 suite failures are the pre-existing `test_e2e_user_agent` live-gateway network tests (fail identically on `main`). `ruff format` + `ruff check` clean.

## Not done / caveats

- **No live-gateway handshake under mcp 2.x**: all my OAuth tokens expired this session (no browser), so I proved transport *construction* + import + preflight under 2.x, not a real end-to-end tool call. A reviewer with a live workspace and mcp 2.x installed can confirm the round trip.
- The internal pypi mirror currently resolves mcp 1.28.1 (2.x may not be mirrored yet); the uncapped constraint takes whatever is newest available.

This supersedes the earlier "keep the cap + actionable error" revision of this branch.


sunishsheth2009:
Reworked per maintainer direction: instead of keeping the `mcp<2` cap + surfacing an error, the proxy now **supports both mcp 1.x and 2.x** and the cap is removed — a clean install on either major works.

One code path handles both because mcp 1.28+ and 2.x both export `streamable_http_client(url, http_client=…)`. The only differences — httpx vs httpx2, and a 3- vs 2-tuple yield — are resolved at runtime (`_httpx()` picks the flavor the installed SDK uses; `_run` unpacks positionally).

Verified against **both** mcp 1.28.1 and mcp 2.0.0 in a real venv: proxy imports, transport composes, 21 proxy tests + `ty check` clean on each. Details in the updated description.

src/ucode/mcp_proxy.py
```diff
@@ -99,13 +127,22 @@ async def _pump(
 
 
 async def _run(url: str, workspace: str, profile: str | None, use_pat: bool) -> None:
```
xsh310: `use_pat` is in the signature but is never forwarded to `_build_token_auth` or used anywhere in the body. The original code had the same gap — `_DatabricksTokenAuth` stored `_use_pat` but never passed it to `get_databricks_token`. If PAT mode is handled entirely via env (as the comment says), the parameter should be removed from `_run` and its callers so readers do not wonder whether PAT auth is silently broken.

xsh310 (approved):

src/ucode/mcp_proxy.py
sunishsheth2009: Settled: `use_pat` stays on `serve()` (restored) and the CLI keeps its one-liner `serve(url, workspace, profile, use_pat=use_pat or bool(state.get("use_pat")))`. `serve` now actually resolves the PAT — it calls `ensure_pat_bearer(profile)` when `use_pat` is set, exporting the profile's static token as DATABRICKS_BEARER so the per-request mint returns it, and fails fast if none is found. `_run` still takes no `use_pat` (once the bearer's in the env the transport path is identical to OAuth). Squashed the intermediate churn into one commit; PAT behavior is covered in test_mcp_proxy.

PR #340, opened 2026-08-14
mcp: add `ucode mcp add` for additive MCP registration

sunishsheth2009:
## What

Adds a new `ucode mcp add` command that registers Databricks MCP servers **additively** — it never removes servers that are already configured.

`ucode configure mcp` *replaces* the registered MCP servers with the current selection: anything outside the `--location`/`--services` scope (or left unchecked in the interactive picker) is removed. That makes it awkward to tack on a new server without disturbing existing config, and was the source of the "clobbering" reports.

Rather than change `configure mcp`'s replace semantics (which some flows rely on), this PR leaves `configure mcp` exactly as-is and introduces a dedicated additive command, matching the maintainer direction to grow `ucode mcp <verb>` subcommands.

## Behavior

`ucode mcp add` takes the **same** options as `ucode configure mcp` — `--location`, `--services`, and the no-argument interactive picker — but the resolved selection is unioned with the already-configured servers before applying, so nothing outside it is removed.

```bash
# Register a whole schema, keeping any servers already configured.
ucode mcp add --location system.ai

# Register a specific subset.
ucode mcp add --services system.ai.slack,system.ai.github

# Interactive picker, additive (never removes).
ucode mcp add
```

## Changes

- `src/ucode/mcp.py`: `add_mcp_command()` + an internal `append` path on `configure_mcp_command()` (`configure mcp` behavior unchanged) and a `_union_missing()` helper.
- `src/ucode/cli.py`: `@mcp_app.command("add")`.
- `README.md`: documents `ucode mcp add` (additive) vs `configure mcp` (replace).
- `tests/test_mcp.py`: `TestAddMcpCommand` covering the location and services-subset append paths.

## Testing

`uv run ruff format`/`ruff check`, `uv run ty check src/`, and `uv run pytest` all pass (the only failures are the two pre-existing network-dependent `test_e2e_user_agent` tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,26 @@ ucode configure --agents claude --mcp system.ai.slack
 `--mcp` also works without `--agents` for MCP-only clients (it configures just the workspace,
 then registers the servers); pass a comma-separated list to register several at once.
 
+#### Add servers without replacing existing ones
+
+`ucode configure mcp` **replaces** the registered MCP servers with your selection — anything
+outside a `--location`/`--services` scope (or left unchecked in the picker) is removed. To
+**add** servers while leaving everything already configured in place, use `ucode mcp add`:
+
+```bash
+# Register a whole schema's services, keeping any servers already configured.
+ucode mcp add --location system.ai
+
+# Register just a subset (same name rules as `configure mcp --services`).
+ucode mcp add --services system.ai.slack,system.ai.github
+
+# No arguments launches the same interactive picker, but never removes servers.
+ucode mcp add
+```
+
+`ucode mcp add` takes the same `--location` and `--services` options as `ucode configure mcp`;
+the only difference is that it never removes servers outside the selection.
+
 ### Skills (optional)
 
 Configure Unity Catalog Skills for your coding tools with `ucode configure skills`:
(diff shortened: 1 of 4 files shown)


sunishsheth2009:
Added README docs for the append-vs-replace behavior, per the thread consensus (models replace, MCP servers append):

- Adding an MCP server (`ucode configure mcp` / `--mcp`) appends — existing servers are preserved.
- Scoped replace is limited to `configure mcp --location <cat>.<schema>` (replaces that schema's services, leaves other schemas/sources untouched).
- Contrasted with model config, which replaces.

Commit 026192e. No code change — documents the behavior this PR fixes.

README.md
```diff
@@ -142,6 +142,26 @@ ucode configure --agents claude --mcp system.ai.slack
...
+
+```bash
+# Register a whole schema's services, keeping any servers already configured.
+ucode mcp add --location system.ai
+
+# Register just a subset (same name rules as `configure mcp --services`).
+ucode mcp add --services system.ai.slack,system.ai.github
+
+# No arguments launches the same interactive picker, but never removes servers.
+ucode mcp add
```
xsh310, line 159: Do you think we should enable interactive picker with `ucode mcp` instead of `ucode mcp add`? Claude flagged that the interactive picker in add mode silently ignores unchecks in the interactive mode which could be misleading.

src/ucode/cli.py
```diff
@@ -1056,6 +1057,43 @@ def _version_callback(value: bool) -> None:
...
+            "Servers already configured outside this location are kept.",
+        ),
+    ] = None,
+    services: Annotated[
+        str | None,
+        typer.Option(
+            "--services",
+            help="Register exactly this comma-separated subset of MCP services. Full names like "
+            "`system.ai.github` work on their own; bare short names like `github` need --location "
+            "to locate them. Omit --services to register the whole --location schema.",
```
xsh310: `--services ""` behaves differently here than in `configure mcp`, and the help text doesn't flag it:

- `configure mcp --services "" --location <schema>` is documented as "remove all".
- `mcp add --services "" --location <schema>` is a harmless no-op (nothing to add once the union re-keeps everything).
- `mcp add --services ""` without `--location` raises the shared error `--services without --location must all share one <catalog>.<schema> (got: none); pass --location instead.`, which reads oddly for an add-only command.

Non-blocking. Worth a one-line note in the `--services` help, or just confirm the no-op/error behavior is intended.

xsh310 (approved):
Overall LGTM! Left 2 quick comments, ptal!

README.md
sunishsheth2009: Good call — done in 0034115. In `ucode mcp add` (additive) mode the picker now shows already-configured servers as non-toggleable `(already configured)` notes rather than pre-checked boxes, so unchecking is no longer silently ignored; you only select new servers to add. `configure mcp` keeps its pre-checked/removable toggles (replace semantics). Kept it on `mcp add` rather than bare `ucode mcp` so the group stays a pure command namespace.

src/ucode/cli.py
sunishsheth2009: Fixed in 0034115. `mcp add --services ""` is now a clean no-op (nothing to add) instead of the shared "pass --location instead" error, and no longer requires `--location`. Updated the `--services` help to note it's additive and that an empty value is a no-op.

PR #347, opened 2026-08-17
claude: enable Claude Code native gateway model discovery

andy-xu-db:
## Summary

Adds opt-in support for Claude Code's native gateway model discovery. Set `ENABLE_CLAUDE_CODE_GATEWAY_MODEL_DISCOVERY=1` before launching ucode to have Claude Code's `/model` picker list every gateway Messages-API endpoint, rather than only the `ANTHROPIC_DEFAULT_*` family aliases ucode injects today.

## Why

Claude Code has native gateway model discovery: when enabled, it fetches `GET {ANTHROPIC_BASE_URL}/v1/models` and adds a picker row for each returned id (filtered to ids containing `claude` or `anthropic`). Keeping this behind an opt-in ucode environment flag prevents behavior changes for existing users while allowing custom Model Serving endpoints whose ids carry `claude`/`anthropic` (for example, `main.default.my_anthropic_ms` or `system.ai.claude-opus-4-8`) to appear and route natively as 3-part FQNs.

## What changed

- `render_overlay` sets `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1` only when `ENABLE_CLAUDE_CODE_GATEWAY_MODEL_DISCOVERY` is enabled (`1`, `true`, or `yes`).
- Discovery remains disabled under `provider`: a Model Provider Service routes every request to the external provider via header, so a discovered gateway endpoint id would reach a provider that cannot resolve it.
- Relayed launches work because the loopback refresh proxy (`gateway_proxy.py`) forwards `/v1/models` to the gateway.

## Testing

- Verified disabled/default and false-value behavior.
- Verified accepted opt-in values.
- Verified the provider-service exclusion.
- Full `test_agent_claude.py` suite passes (110 tests).
- Ruff passes.

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -40,6 +40,8 @@ from ucode.telemetry import agent_version, ucode_version
 from ucode.tracing import tracing_env
 from ucode.ui import print_err, print_note, print_success, print_warning
 
+GATEWAY_MODEL_DISCOVERY_ENV_VAR = "ENABLE_CLAUDE_CODE_GATEWAY_MODEL_DISCOVERY"
+
 CLAUDE_CONFIG_DIR = Path.home() / ".claude"
 CLAUDE_SETTINGS_PATH = CLAUDE_CONFIG_DIR / "ucode-settings.json"
 CLAUDE_BACKUP_PATH = APP_DIR / "claude-ucode-settings.backup.json"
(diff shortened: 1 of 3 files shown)


lilly-luo (commented):
can you pls gate this behind an environment feature flag (see `ENABLE_MANAGED_AGENT_CONFIG`) so that users won't break unless they explicitly use

andy-xu-db:
Addressed the opt-in request in a5eb902. Gateway model discovery is now disabled by default and only enabled when `ENABLE_CLAUDE_CODE_GATEWAY_MODEL_DISCOVERY` is set to `1`, `true`, or `yes`; the provider-service exclusion remains in place. Added coverage for disabled, enabled, and provider cases (110 Claude-agent tests passing).

src/ucode/agents/claude.py
```diff
@@ -302,6 +304,16 @@ def render_overlay(
         "ENABLE_TOOL_SEARCH": "1",
         "CLAUDE_CODE_USE_GATEWAY": "1",
     }
+    # Native /model discovery: picker lists every gateway Messages-API endpoint,
+    # not just the family aliases. Skipped under a provider (its routing header
+    # would send a discovered gateway id to a provider that can't resolve it).
+    discovery_enabled = os.environ.get(GATEWAY_MODEL_DISCOVERY_ENV_VAR, "").strip().lower() in (
+        "1",
```
lilly-luo: you can just make it 1 for simplicity, this is just temp until the server change is live

src/ucode/agents/claude.py
```diff
@@ -40,6 +40,8 @@
 from ucode.tracing import tracing_env
 from ucode.ui import print_err, print_note, print_success, print_warning
 
+GATEWAY_MODEL_DISCOVERY_ENV_VAR = "ENABLE_CLAUDE_CODE_GATEWAY_MODEL_DISCOVERY"
+
```
lilly-luo: nit rm newline

lilly-luo (approved):

PR #350, opened 2026-08-17
Don't install Databricks AI Tools on agent launch (configure-only)

rohita5l:
## Problem

Running a launch command like:

```
ucode claude --provider main.jasmine.rohit-test --workspace https://<ws>
```

prints **"Installing Databricks AI Tools for claude-code..."** unprompted. Launch commands auto-configure an unseen agent through `configure_single_tool`, which also installed Databricks AI Tools (skills + plugins). That put skill installation on the **launch** path, not just the **configure** path.

## Fix

Move the install out of the shared `configure_single_tool` and into the `ucode configure` callers only:

- `configure_single_tool` no longer installs — the launch path (`_auto_configure_tool`) auto-configures through it and must never install skills.
- `configure_workspace_command` installs explicitly after `configure_single_tool` on the single-agent path; the multi-agent path already installs via `configure_selected_tools`.

## Result

- **Launch (`ucode claude`, etc.) never installs skills.** `ucode claude --provider ... --workspace ...` now just configures the agent for the provider and launches — no install step.
- **`ucode configure` still installs**, on by default, with `--disable-databricks-ai-tools` to opt out (unchanged).

## Testing

- `TestConfigureWiresAiToolsInstall`: `configure_single_tool` is now asserted to **not** install; `configure_selected_tools` still does.
- The configure single-agent cli test now asserts the configure path installs (and mocks it, so it no longer makes a real `databricks aitools` subprocess call).
- `tests/test_agents_init.py` + `tests/test_cli.py`: 288 passed. Lint + format clean.

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -413,11 +413,7 @@ def _availability_failure_detail(tool: str, state: dict) -> str:
 
 
 def configure_single_tool(tool: str, state: dict) -> dict:
-    """Check availability, configure, and persist state for one tool only.
-
-    Does NOT install Databricks AI Tools — that is a `ucode configure`-only step
-    (see `install_databricks_ai_tools_for_agents` callers). The launch path auto-configures
-    through here, and launching must never install skills."""
+    """Check availability, configure, and persist state for one tool only."""
     provider = get_provider_service(state, tool)
     # A Model Provider Service routes through the same gateway and pins no
     # Databricks model, so the per-tool model availability check doesn't apply.
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -413,7 +413,11 @@ def _availability_failure_detail(tool: str, state: dict) -> str:
 
 
 def configure_single_tool(tool: str, state: dict) -> dict:
-    """Check availability, configure, and persist state for one tool only."""
+    """Check availability, configure, and persist state for one tool only.
+
+    Does NOT install Databricks AI Tools — that is a `ucode configure`-only step
+    (see `install_ai_tools_for_agents` callers). The launch path auto-configures
+    through here, and launching must never install skills."""
     provider = get_provider_service(state, tool)
     # A Model Provider Service routes through the same gateway and pins no
     # Databricks model, so the per-tool model availability check doesn't apply.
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -769,6 +770,10 @@ def configure_workspace_command(
         )
         state = states[0]
         state = configure_single_tool(tool, state)
+        # Install Databricks AI Tools here (a `ucode configure`-only step), not in
+        # configure_single_tool — the launch path auto-configures through that and
+        # must never install skills.
+        install_ai_tools_for_agents([tool], state)
```
lilly-luo: nit can we call this install_databricks_ai_tools_for_agents to make it clear that it's our tools (vs the customer's tools)

src/ucode/agents/__init__.py
```diff
@@ -413,7 +413,11 @@ def _availability_failure_detail(tool: str, state: dict) -> str:
 
 
 def configure_single_tool(tool: str, state: dict) -> dict:
-    """Check availability, configure, and persist state for one tool only."""
+    """Check availability, configure, and persist state for one tool only.
+
+    Does NOT install Databricks AI Tools — that is a `ucode configure`-only step
+    (see `install_ai_tools_for_agents` callers). The launch path auto-configures
+    through here, and launching must never install skills."""
```
lilly-luo: can rm this comment

src/ucode/cli.py
```diff
@@ -769,6 +770,10 @@ def configure_workspace_command(
         )
         state = states[0]
         state = configure_single_tool(tool, state)
+        # Install Databricks AI Tools here (a `ucode configure`-only step), not in
+        # configure_single_tool — the launch path auto-configures through that and
+        # must never install skills.
```
lilly-luo: maybe rm this

lilly-luo (approved):

src/ucode/cli.py
rohita5l: Done — renamed to install_databricks_ai_tools_for_agents.

src/ucode/agents/__init__.py
rohita5l: Done — removed the added docstring paragraph.

src/ucode/cli.py
rohita5l: Done — removed the comment.

PR #351, opened 2026-08-18
mcp: add `ucode mcp remove` to unregister configured MCP servers

sunishsheth2009:
## What

Adds `ucode mcp remove`, completing the `ucode mcp <verb>` set alongside the recently-merged `ucode mcp add` (#340).

`ucode mcp remove` shows the MCP servers you currently have configured — each annotated with the coding tools it's registered on — and unregisters the ones you select from those tools.

```bash
$ ucode mcp remove
? Remove MCP: (space to toggle, ctrl-a all, enter to remove, type to filter)
  [ ] system-ai-github (Claude Code, Codex)
  [ ] databricks-sql (Claude Code)
```

## Design

- **Interactive only** — no `--location`/`--services`/`--all` flags. Removal is a small, destructive operation where seeing the concrete list (and which tools each server is on) before selecting is the safer UX.
- **Skills connections are never shown** — they're owned by `ucode configure skills`, matching how the `configure mcp` / `add` pickers already treat them.
- **No Databricks login required** — unregistering a server is purely local (it shells out to `claude mcp remove` and friends), so it works even when the workspace token has expired. `setup_mcp_clients` grows a `require_auth` toggle for this; the add/configure paths keep forcing auth as before.

Removal targets each server's *recorded* clients (via the existing `apply_mcp_server_changes` remove path), so a server registered on a tool that isn't currently configured is still cleaned up.

## Changes

- `src/ucode/mcp.py`: `remove_mcp_command()`, `_prompt_for_mcp_removal()`, and a `require_auth`/`action_note` toggle on `setup_mcp_clients`.
- `src/ucode/cli.py`: `@mcp_app.command("remove")`.
- `README.md`: documents `ucode mcp remove`.
- `tests/test_mcp.py`: `TestRemoveMcpCommand` (removes selection, cancel = no-op, skills excluded, empty config skips the picker) + a picker-rendering test.

## Testing

`uv run ruff format`/`ruff check`, `uv run ty check src/`, and `uv run pytest` all pass (the only failures are the two pre-existing network-dependent `test_e2e_user_agent` tests).

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -91,6 +91,7 @@ from ucode.mcp import (
     configure_mcp_command,
     configure_skills_mcp_command,
     purge_cross_workspace_mcp_residue,
+    remove_mcp_command,
     revert_mcp_configs,
 )
 from ucode.skills_download import (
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -1097,6 +1098,23 @@ def mcp_add(
         raise typer.Exit(130) from None
 
 
+@mcp_app.command("remove")
+def mcp_remove() -> None:
```
xsh310, line 1102: Should we also add support `--location` and `--service` flags for remove command? I think that will make it more actionable for agents to run and also better parity with the add cmd. Feel free to add it in a future PR tho.

xsh310 (approved):
LGTM, left a quick comment

src/ucode/cli.py
sunishsheth2009: Yes since location and services mean we need to add validation that we are removing something that we have added.

PR #353, opened 2026-08-18
mcp add/remove: add `--agents` to target (and set up) specific agents

sunishsheth2009:
## What

Add an `--agents` option to `ucode mcp add` and `ucode mcp remove`.

Today both commands operate on **every** configured MCP client, and `mcp add` requires the agent to already be set up (via `ucode configure`) — so connecting a fresh agent to an MCP server takes two commands.

**`ucode mcp add --agents <list>`** — register the server(s) for exactly those agents. Any that aren't configured yet are set up first (workspace + models), reusing the `configure --agents` path, so it doubles as one-command setup:

```bash
ucode mcp add --agents claude --services system.ai.slack   # sets up claude if needed, then registers
ucode mcp add --agents claude,codex --location system.ai
```

**`ucode mcp remove --agents <list>`** — scope removal to those agents. A server registered on several agents is unregistered only from the named ones and kept on the rest; only servers on a named agent are offered in the picker.

Without `--agents`, both commands behave exactly as before.

## Why

This makes `ucode mcp add` a one-stop command: it can set the agent up *and* register the MCP server, so callers (and the Catalog Explorer / docs snippets) no longer need a separate `ucode configure --agents` step. It's the natural home for the agent+MCP bundling that `configure --mcp` used to provide, now under the `mcp` verb namespace.

## How

- `setup_mcp_clients` grows an `agents` filter that scopes the resolved clients (and errors clearly if a named agent isn't configured).
- `add_mcp_command` threads `agents` through to that filter; the CLI's `mcp add` reuses `configure_workspace_command` to set up any un-configured named agents before registering.
- `remove_mcp_command` applies agent-scoped removal: it presents each selected server to the existing removal machinery with its client list narrowed to the named agents, then keeps the entry (with those agents stripped) or drops it when no agents remain.

## Testing

`uv run ruff format`/`ruff check`, `uv run ty check src/`, and `uv run pytest` all pass. Added `tests/test_mcp.py` coverage: add scoped to a subset of configured agents, add erroring on an unconfigured agent, remove scoped to one agent (kept on the rest), and remove offering only servers on the named agent. The only suite failures are the two pre-existing network-dependent `test_e2e_user_agent` tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,12 +164,29 @@ the only difference is that it never removes servers outside the selection. In t
 picker, servers you already have configured are shown as `(already configured)` and can't be
 toggled off — you only pick new ones to add.
 
+Pass `--agents` to target specific coding agents. Any named agent that isn't set up yet is
+configured first (workspace + models), so this doubles as one-command setup:
+
+```bash
+# Set up Claude Code (if needed) and register the server for it, in one command.
+ucode mcp add --agents claude --services system.ai.slack
+
+# Target several agents at once.
+ucode mcp add --agents claude,codex --location system.ai
+```
+
+Without `--agents`, the server is registered for every already-configured agent.
+
 #### Remove configured servers
 
 To unregister servers you've already configured, use `ucode mcp remove`:
 
 ```bash
 ucode mcp remove
+
+# Remove only from specific agents. A server registered on several agents is
+# unregistered from the named ones and kept on the rest.
+ucode mcp remove --agents codex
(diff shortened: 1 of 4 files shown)


lilly-luo:
i'm worried this creates multiple entrypoints for agent configuration which will be confusing. could you link the issue/complaint that prompted this change?

xsh310 (approved):

PR #356, opened 2026-08-18
Model picker: show all models in a fixed scrolling viewport

rohita5l:
## Problem

In the Coding Agent config setup (`ucode setup` / `ucode configure`), the per-agent model picker only offered the **newest 5 models per family** (`_MODEL_PICKER_LIMIT = 5`). On a workspace with more (a typical one has ~6 opus versions), the rest were dropped from the list entirely — so it looked like the picker "only shows the top 5 and can't scroll." The only way to reach a dropped model was the "type a custom model" escape hatch.

## Changes

1. **Stop truncating.** Both the Claude per-family loop and `_custom_option_rows` now offer *every* discovered model, with the "✎ Enter a custom model…" row still last for ids outside `system.ai` that discovery never lists.

2. **Fixed 10-row scrolling viewport.** A full catalog can run long, so the choice list is pinned to a 10-row window that scrolls through the rest. questionary 2.1.1 has no `page_size`, so `_cap_choice_viewport` (in `ui.py`) bounds the built question's `InquirerControl` window to `Dimension(preferred=10, max=10)` when there are >10 options; lists at or below the threshold keep their natural height. The highlighted row stays in view via the pointer's cursor token. This reuses questionary's built-in select/checkbox keybindings and search rather than the heavier custom widget in `mcp.py`.

## Tests

- Updated the truncation test in `test_managed_wizard.py` to assert all discovered ids are offered (no cap).
- Added `TestChoiceViewportCap` in `test_ui.py`: short-list no-op, long-list cap for both select and checkbox, and missing-application no-op.
- Full suite passes except one pre-existing e2e failure (`test_user_agent_arrives_at_gateway`), confirmed to fail identically on a clean checkout — unrelated to this change.

### DIFF
--- a/src/ucode/managed_wizard.py
+++ b/src/ucode/managed_wizard.py
@@ -473,7 +473,7 @@ def _prompt_claude_models(state: dict) -> dict:
         family_models = candidates.get(family)
         if not family_models:
             continue
-        rows = [(model, model) for model in family_models[:_MODEL_PICKER_LIMIT]] + [
+        rows = [(model, model) for model in family_models] + [
             (_CUSTOM_MODEL, _CUSTOM_MODEL_LABEL),
             (_SKIP_FAMILY, f"(skip {family})"),
         ]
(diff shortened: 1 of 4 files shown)


lilly-luo:
Screenshot 2026-08-18 at 9 53 25 PM

lilly-luo (approved):

PR #362, opened 2026-08-20
claude: authenticate gateway model discovery

andy-xu-db:
## Summary

Set the gateway URL, mode, and auth token in the process environment before Claude starts.

Claude Code checks these values before settings-based environment variables and `apiKeyHelper` resolve. Without them, it skips `GET /anthropic/v1/models`.

Relayed launches are unchanged.

This PR is independent from #347 and based directly on `main`.

## Testing

- `pytest tests/test_agent_claude.py` (100 passed)
- Ruff

## 🥞 Stacked PR

Use this [link](https://github.com/databricks/ucode/pull/362/files) to review incremental changes.

- [**claude/gateway-model-discovery-auth**](https://github.com/databricks/ucode/pull/362) [[Files changed](https://github.com/databricks/ucode/pull/362/files)] ← _this PR_
- [andy.xu/claude-gateway-model-aliases](https://github.com/databricks/ucode/pull/375) [[Files changed](https://github.com/databricks/ucode/pull/375/files)]
- [andy.xu/claude-gateway-model-cache](https://github.com/databricks/ucode/pull/390) [[Files changed](https://github.com/databricks/ucode/pull/390/files)]

---------


src/ucode/agents/claude.py
```diff
@@ -1010,14 +1016,48 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
     raise SystemExit(returncode)
 
 
+def _launch_gateway(state: dict, binary: str, tool_args: list[str]) -> None:
+    from ucode.gateway_proxy import AUTHORIZATION_HEADER, start_proxy
```
lilly-luo: can this be at the top of the file instead

src/ucode/agents/claude.py
```diff
@@ -1010,14 +1016,48 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
...
+
 def launch(state: dict, tool_args: list[str]) -> None:
     binary = SPEC["binary"]
     workspace = state.get("workspace")
     if state.get("claude_relayed"):
         _launch_relayed(state, binary, tool_args)
         return
     if workspace:
-        os.environ["OAUTH_TOKEN"] = get_databricks_token(workspace, state.get("profile"))
+        _launch_gateway(state, binary, tool_args)
```
lilly-luo, line 1074: this behavior is changing? is that expected?

src/ucode/gateway_proxy.py
```diff
@@ -128,13 +126,11 @@ def __init__(self, workspace: str, profile: str | None) -> None:
         # Force on start so we begin on a full-TTL token rather than inheriting a
         # near-expiry one cached from an earlier CLI call. Raises if auth is dead
         # (surfaced by the caller at launch, before Claude Code starts).
-        self._refresh(force=True)
+        self._refresh()
```
lilly-luo: why is this changing

src/ucode/gateway_proxy.py
lilly-luo, line 1: did you confirm this works with relayed auth? https://docs.google.com/spreadsheets/d/19M0DShshcYS1aiLg49dPidIsujRPyWhC_ZQRcpeT3Kw/edit?gid=0#gid=0

lilly-luo, line 1: could you gate the new behavior here behind the model discovery env feature flag? we want to make sure relayed auth continues to work as expected and users have to opt into this behavior

src/ucode/gateway_proxy.py
```diff
@@ -128,13 +126,11 @@ def __init__(self, workspace: str, profile: str | None) -> None:
...
 
-    def _refresh(self, *, force: bool) -> None:
-        """Mint a token and record its expiry. Caller holds `_refresh_lock` (or is
-        __init__). Non-force lets a token another process just refreshed satisfy
-        this call from the shared cache with no write — shrinking lock contention."""
-        token = get_databricks_token(self._workspace, self._profile, force_refresh=force)
+    def _refresh(self) -> None:
+        """Force-mint a token and record its expiry."""
+        token = get_databricks_token(self._workspace, self._profile, force_refresh=True)
```
lilly-luo: why are we changing to always force refresh? this is altering the existing behavior, for ex line 154 in the original code did not force a refresh

src/ucode/gateway_proxy.py
andy-xu-db: I tested with relayed auth and it works without any issues.

src/ucode/gateway_proxy.py
andy-xu-db: We force refresh because a normal token request may return the same nearly expired cached token. Otherwise the refresh would trigger once the current token errors which could cause some queries to fail.

src/ucode/gateway_proxy.py
```diff
@@ -357,7 +362,11 @@ def __getattr__(self, name: str):
 
 
 def start_proxy(
-    workspace: str, profile: str | None, port: int
+    workspace: str,
+    profile: str | None,
+    port: int,
+    token_header: str = _SWAP_HEADER,
+    force_refresh_near_expiry: bool = False,
```
lilly-luo: nit don't set default, force call site to specify

lilly-luo (approved):

PR #363, opened 2026-08-20
smart routing: switch model in codex after first prompt 

lilly-luo:
this demo shows i can: 
1/ launch codex with ucode codex (gpt-5.6-luna)
2/ type a prompt 
3/ after my first prompt, switch model (glm-5-2)
4/ interact w the new model. 

for now, (3) is hardcoded. next step is for (3) to be triggered by smart router api call. this is only hit with feature flag `ENABLE_SMART_ROUTING_V2=1` so no current behavior will be impacted 

https://github.com/user-attachments/assets/6cb9e257-cde2-497b-9f48-1476664a1b4b

See https://docs.google.com/document/d/1lO1l3sOH0V9_YIx2SQQIvDOBlBJfIwpITLFl1rYTZ7s/edit?tab=t.whcgscnu6a6r "Codex - Smart Router flow on first prompt (“hi”)" on "Technical Implementation" for details on how this works


src/ucode/smart_routing/codex_interposer.py
```diff
@@ -0,0 +1,242 @@
...
+            return raw
+        if not isinstance(msg, dict):
+            return raw
+        params = msg.get("params")
+        if msg.get("method") == TURN_START and isinstance(params, dict):
+            if isinstance(params.get("threadId"), str):
+                self.thread_id = params["threadId"]
+            old = params.get("model")
+            if old != self.target:
+                params["model"] = self.target
```
andy-xu-db, line 55: Just wondering would this override every turn? I'm not sure if we want to allow users to select other models if they want after the first routed turn

src/ucode/smart_routing/v2.py
```diff
@@ -0,0 +1,151 @@
...
+        "Using Unity Gateway Smart Router.",
+        f"Selected Model : {model}",
+        f"Reason : {reason}",
+    ]
+    width = max(len(line) for line in lines)
+    border = "─" * (width + 2)
+    return "\n".join([f"┌{border}┐", *(f"│ {line:<{width}} │" for line in lines), f"└{border}┘"])
+
+
+def _generate_codex_app_server_home(
```
andy-xu-db: Would this miss out on any user-specific settings for tools in ~/.codex/config.toml?

andy-xu-db (approved):

src/ucode/smart_routing/codex_interposer.py
lilly-luo: great catch - yes, we need users to be able to change the model after their first routed turn.

src/ucode/smart_routing/v2.py
lilly-luo: great pt, yea - fixed to pass in the specified config

lilly-luo: confirming my mcps still load in 
Screenshot 2026-08-25 at 10 04 48 PM

src/ucode/smart_routing/codex_interposer.py
lilly-luo: confirmed models can now be changed 
Screenshot 2026-08-25 at 10 04 31 PM

PR #370, opened 2026-08-24
Separate managed config drafts from workspace cache

rohita5l:
## Why

`~/.ucode/managed-state.json` currently serves two different owners: `ucode setup` writes an admins unpublished draft, while every normal launch overwrites the same file with the workspace-published config. That means testing or even launching can destroy unpublished edits. A single fallback file also cannot safely represent multiple workspaces.

This PR fixes the storage model only. It does not add new CLI workflow.

## New invariants

- `~/.ucode/managed-state.json` is only the editable draft written by setup.
- `~/.ucode/managed-cache/<workspace-hash>.json` is only the last config fetched from that workspace.
- Normal refreshes and outage fallback never read or write the draft.
- A successful workspace response saying no config exists records an empty cache entry, so an older policy cannot be resurrected after an outage.

## Implementation

- add per-workspace cache paths derived from SHA-256 of the workspace URL
- add versioned envelopes and cache provenance metadata (`source`, `fetched_at`, canonical config digest)
- verify cache digests before using fallback data
- write drafts and caches atomically with same-directory temp files, `fsync`, mode `0600`, and `os.replace`
- continue reading the legacy single-workspace cache for migration, while all new writes use the new layout
- make bare `ucode --dry-run` read the published cache instead of the draft

## Review guide

1. Start with `managed_config.py`: `_save_managed_payload`, `save/load_managed_cache`, then `refresh_managed_config`.
2. Check the two-line CLI source change for dry-run.
3. Review `TestPersistence` and `TestRefreshManagedConfig` for the invariants and failure cases.

## Testing

- `pytest tests/test_managed_config.py tests/test_cli.py -q`: 275 passed

## Stack

1. This PR: storage isolation and persistence safety
2. Follow-up: admin-only local draft testing and removal of the config bypass
3. Follow-up: setup wizard presentation improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,8 @@ their next ucode run.
 | `~/.copilot/.env` | GitHub Copilot CLI |
 | `~/.pi/agent/models.json` | Pi |
 | `~/.cursor/mcp.json` | Cursor Agent (MCP servers only) |
-| `~/.ucode/managed-state.json` | The managed config — authored by `ucode setup` (admins) and refreshed from the workspace on launch |
+| `~/.ucode/managed-state.json` | The editable managed-config draft authored by `ucode setup` |
+| `~/.ucode/managed-cache/<workspace-hash>.json` | Per-workspace published-config cache with fetch metadata |
 
 Existing files are backed up before being overwritten. `ucode revert` restores backups.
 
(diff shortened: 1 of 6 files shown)


rohita5l:
Stack navigation: **#370 (storage)** → **#371 (local testing/publish policy)** → **#372 (setup UX)**. Each PR is based on the previous branch so the Files changed tab shows only that layer.

PR #371, opened 2026-08-24
Add admin-only local managed config testing

rohita5l:
## Depends on

- #370 separates the editable draft from the workspace-published cache.

Review this PR with **base = `managed-config-cache-safety`**. Its diff is only the CLI workflow layered on that storage model.

## Problem

Admins can author a managed config with `ucode setup`, but they need a safe way to launch an agent with unpublished edits before publishing. At the same time, ordinary developer launches must keep using the workspace-published config, and publishing should not depend on whether or when the admin tested.

## User-visible behavior

| Command | Config source | Who can use it |
| --- | --- | --- |
| `ucode` or `ucode <agent>` | latest workspace-published config | any developer |
| `ucode --local` | local drafts configured default agent | verified workspace admin |
| `ucode <agent> --local` | local draft for that agent | verified workspace admin |
| `ucode apply` | publishes the latest local draft | workspace admin |

Testing is optional. An admin can edit the draft repeatedly, test any revision, and `ucode apply` always publishes the latest saved revision.

## Authorization and failure behavior

A local launch has no server-side write on which the workspace can enforce admin permissions. Therefore `--local` checks the current workspace identity through SCIM and fails closed:

- admin = true: continue
- admin = false: block before agent launch
- admin status unknown or authentication fails: block and ask the user to retry
- no draft: show `ucode setup` guidance before making the admin check

Bare `ucode --local` verifies once and passes that result into the selected agent launch.

## Managed-config bypass removal

`--skip-managed-config` previously cleared the feature environment variable for one process and let a developer ignore the published config. That conflicts with the new source model, so the bypass is removed.

The old spelling is retained only as a hidden fail-closed tombstone. This matters because agent commands forward unknown flags to Claude/Codex; deleting the parser outright would silently pass an old ucode flag to the underlying agent. `--skip-preflight` remains supported and only skips redundant auth/gateway validation; it still applies the workspace config.

## Review guide

1. In `cli.py`, review `_require_local_config_admin` and the local branch in `_launch_tool`.
2. Review `_launch_managed_default` for bare `ucode --local` and its single-verification handoff.
3. Review the removed bypass option and confirm `--skip-preflight` remains independent.
4. Read `test_managed_workflow.py` for the setup -> local launch -> apply -> edit again -> normal published launch sequence.

## Testing

- `pytest tests/test_cli.py tests/test_managed_config.py tests/test_managed_wizard.py tests/test_managed_workflow.py -q`: 481 passed

## Stack

1. #370: storage isolation and persistence safety
2. This PR: local testing and publish policy
3. Follow-up: setup wizard presentation improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,9 @@ ucode setup                 # agents and models (start here)
 ucode setup mcps            # managed MCP servers
 ucode setup skills          # managed skills
 ucode setup spend-tiers     # spend-based routing
-ucode apply                 # publish it to the workspace
+ucode claude --local        # optionally test the local draft
+ucode codex --local         # edit and test as often as needed
+ucode apply                 # publish the latest draft to the workspace
 ```
 
 `ucode setup` walks through the agents to enable and which one bare `ucode` launches, then per agent:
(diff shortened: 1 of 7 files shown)


rohita5l:
Stack navigation: **#370 (storage)** → **#371 (local testing/publish policy)** → **#372 (setup UX)**. Review this PR against `managed-config-cache-safety`; the Files changed tab excludes the storage implementation.

PR #372, opened 2026-08-24
Streamline the managed config setup wizard

rohita5l:
## Depends on

- #370: draft/cache storage isolation
- #371: admin-only local testing and publish workflow

Review this PR with **base = `managed-config-local-draft`**. Its diff is only the interactive setup experience.

## Problem

The managed-config setup flow works, but it reads as a series of disconnected prompts and panels. Reviewers and admins cannot easily see where they are in the process, long model lists are hard to navigate, and setup repeats authentication/gateway checks that it already completed before discovery.

This PR changes presentation and prompt orchestration; it does not change the draft/published source rules introduced by the lower PRs.

## New setup flow

`ucode setup` is presented as four explicit phases on one vertical progress rail:

1. Select the workspace
2. Select coding agents
3. Select models for each agent
4. Select the default agent

The rail stays active across prompts, status messages, validation, and completion so the output reads as one wizard rather than unrelated cards.

## Prompt behavior

- use InquirerPy only inside `ucode setup`; the rest of ucode remains on the existing Questionary prompts
- add compact selected-value summaries after each answer
- use fixed-height choice windows and searchable fuzzy selection only for lists long enough to need filtering
- keep choice rows aligned under the wizard rail, including highlighted and separator rows
- make checkbox and confirmation controls state their keyboard behavior and defaults explicitly
- keep all newly authored agent configuration ucode-scoped; imported legacy manifests still display an explicit global-settings field when present

## Connection-check cleanup

Setup needs authentication and admin status before it can inspect an existing published config, and model discovery needs the gateway immediately afterward. Previously the shared configure path repeated those checks and success messages.

`configure_shared_state(..., connection_checks_verified=True)` skips only the duplicate auth and gateway probes. It still obtains the bearer token and performs model discovery. Normal configure and launch behavior is unchanged.

## Implementation notes

- `inquirerpy_wizard` uses context-local state, so nested prompt helpers automatically choose the wizard backend and cleanup does not leak into later commands
- existing Rich helpers become rail-aware only while the wizard context is active
- InquirerPy is a direct dependency; Questionary remains for the specialized MCP pickers and all non-wizard flows

## Review guide

1. Start with `setup_command` in `managed_wizard.py` to see the four phases and check ordering.
2. Review the small `connection_checks_verified` path in `configure_shared_state`.
3. In `ui.py`, review the wizard context/header/step/outro helpers, then the InquirerPy branches in each prompt helper.
4. Use `tests/test_ui.py` for layout, cleanup, fuzzy-list thresholds, keyboard hints, and choice-value behavior.

## Testing

- `pytest tests/test_ui.py tests/test_managed_wizard.py tests/test_cli.py -q`: 499 passed

## Stack

1. #370: storage isolation and persistence safety
2. #371: local testing and publish policy
3. This PR: setup wizard presentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,9 @@ dependencies = [
   # transitively by mcp itself); `ucode.mcp_proxy` selects whichever the
   # installed SDK uses at runtime, so no direct httpx2 dependency is declared.
   "httpx>=0.27.1",
+  # Polished searchable/selectable prompts for the managed-config setup wizard. Questionary remains
+  # in use for the specialized MCP pickers, whose custom viewport and toolbar depend on its API.
+  "InquirerPy>=0.3.4",
   # `ucode mcp-proxy` bridges a client's stdio MCP transport to a Databricks
   # streamable-HTTP MCP endpoint, injecting a freshly-minted OAuth bearer per
   # request, via the SDK's stdio server + streamable-HTTP client. Works against
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -279,9 +279,9 @@ def _maybe_offer_admin_setup(workspace: str, profile: str | None) -> None:
     if not is_admin:
         return
     print_note(
-        "✨ New: run `ucode setup` to publish a managed config to a workspace — set agents, models, mcps "
-        "and skills once, and every developer inherits them when running `ucode`. This scales "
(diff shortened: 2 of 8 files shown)


rohita5l:
Stack navigation: **#370 (storage)** → **#371 (local testing/publish policy)** → **#372 (setup UX)**. Review this PR against `managed-config-local-draft`; the Files changed tab contains only the presentation layer.

PR #375, opened 2026-08-24
Claude: support custom models in gateway discovery

andy-xu-db:
## Summary

Claude Code filters gateway-discovered models by ID, which hides custom model-service FQNs that do not contain `claude` or `anthropic`.

This change updates the relayed gateway proxy to:
- advertise those models with an `anthropic-aigw-` ID while preserving the original `display_name`
- rewrite pagination cursors to the upstream IDs
- rewrite selected model IDs back before forwarding Messages API requests
- reverse only aliases observed in discovery responses, leaving unrelated prefixed IDs untouched

## Testing

- `uv run ruff check .`
- `uv run pytest tests/test_gateway_proxy.py` (28 passed)
- `uv run pytest` (1944 passed, 37 skipped, 1 environment-dependent failure: the Claude user-agent e2e timed out before reaching its local capture server)

## 🥞 Stacked PR

Use this [link](https://github.com/databricks/ucode/pull/375/files) to review incremental changes.

- [claude/gateway-model-discovery-auth](https://github.com/databricks/ucode/pull/362) [[Files changed](https://github.com/databricks/ucode/pull/362/files)]
- [**andy.xu/claude-gateway-model-aliases**](https://github.com/databricks/ucode/pull/375) [[Files changed](https://github.com/databricks/ucode/pull/375/files)] ← _this PR_
- [andy.xu/claude-gateway-model-cache](https://github.com/databricks/ucode/pull/390) [[Files changed](https://github.com/databricks/ucode/pull/390/files)]

---------


src/ucode/gateway_proxy.py
```diff
@@ -193,11 +195,82 @@ def _forwarded_request_headers(
     return headers
 
 
+class _ModelAliases:
```
lilly-luo: _ModelAlias -> _AnthropicModelAliases

src/ucode/gateway_proxy.py
```diff
@@ -193,11 +195,82 @@ def _forwarded_request_headers(
...
+                return body
+        except (UnicodeDecodeError, json.JSONDecodeError, KeyError, TypeError):
+            return body
+
+        aliases: dict[str, str] = {}
+        for model in models:
+            if not isinstance(model, dict) or not isinstance(model.get("id"), str):
+                continue
+            model_id = model["id"]
+            if "claude" in model_id.lower() or "anthropic" in model_id.lower():
```
lilly-luo: is in the right check or is it startsWith?

src/ucode/gateway_proxy.py
```diff
@@ -193,11 +195,82 @@ def _forwarded_request_headers(
...
+                payload[cursor] = alias
+        return json.dumps(payload, separators=(",", ":")).encode()
+
+    def original_id(self, model_id: str) -> str:
+        with self._lock:
+            return self._original_by_alias.get(model_id, model_id)
+
+    def rewrite_path(self, path: str) -> str:
+        parsed = urlsplit(path)
+        if parsed.path != "/v1/models":
```
lilly-luo: can you make this anthropic/v1/models? we shouldnt be doing this for codex

src/ucode/gateway_proxy.py
```diff
@@ -234,7 +308,13 @@ def _handle(self) -> None:
                     elapsed_ms=round((time.monotonic() - started) * 1000),
                 )
                 if resp.status_code not in (401, 403):
-                    self._relay_response(resp, diagnostic_id=diagnostic_id, started=started)
+                    self._relay_response(
+                        resp,
+                        transform_models=self.command == "GET"
+                        and urlsplit(self.path).path == "/v1/models",
```
lilly-luo: can you split this into a variable instead? 

```
should_transform_model_names = self.command == "GET" and urlsplit(self.path).path == "/v1/models"
```

src/ucode/gateway_proxy.py
```diff
@@ -258,7 +338,13 @@ def _handle(self) -> None:
                     status=resp.status_code,
                     elapsed_ms=round((time.monotonic() - started) * 1000),
                 )
-                self._relay_response(resp, diagnostic_id=diagnostic_id, started=started)
+                self._relay_response(
+                    resp,
+                    transform_models=self.command == "GET"
+                    and urlsplit(self.path).path == "/v1/models",
```
lilly-luo: same here

src/ucode/gateway_proxy.py
```diff
@@ -289,6 +375,7 @@ def _relay_response(
         self,
         resp: httpx.Response,
         *,
+        transform_models: bool = False,
```
lilly-luo: no default value, force call site to specify

src/ucode/gateway_proxy.py
```diff
@@ -299,7 +386,9 @@ def _relay_response(
         try:
             self.send_response(resp.status_code)
             for key, value in resp.headers.items():
-                if key.lower() not in _HOP_BY_HOP:
+                if key.lower() not in _HOP_BY_HOP and not (
+                    transform_models and key.lower() == "content-encoding"
+                ):
```
lilly-luo: can you rename this into a clearer variable? like what does the below signify? 

```
not (
                    transform_models and key.lower() == "content-encoding"
                )
```

src/ucode/gateway_proxy.py
```diff
@@ -308,7 +397,12 @@ def _relay_response(
             # With ``chunk_size=None`` (the default), raw upstream chunks are
             # yielded as they arrive and pings keep the downstream connection
             # alive even before the model produces a large content block.
-            for chunk in resp.iter_raw():
+            response_chunks = (
+                [self.model_aliases.advertise_models(resp.read())]
+                if transform_models and 200 <= resp.status_code < 300
```
lilly-luo: make 200 and 300 constants

src/ucode/gateway_proxy.py
andy-xu-db: Yes, in is correct since claude code checks if "claude" or "anthropic" exist anywhere in the name.

src/ucode/agents/claude.py
lilly-luo: nit could you make 127.0.0.1 a const and put it in a generic file? i'll need to use it as well

src/ucode/gateway_proxy.py
```diff
@@ -195,11 +200,83 @@ def _forwarded_request_headers(
...
 
 
+class _AnthropicModelAliases:
+    """Maps Claude-compatible discovery IDs back to their gateway model IDs."""
+
+    def __init__(self) -> None:
+        self._original_by_alias: dict[str, str] = {}
+        self._lock = threading.Lock()
+
+    def advertise_models(self, body: bytes) -> bytes:
```
lilly-luo: advertise_models are like the models the user sees right? maybe UserDisplayedModels

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -306,7 +399,11 @@ def _relay_response(
         try:
             self.send_response(resp.status_code)
             for key, value in resp.headers.items():
-                if key.lower() not in _HOP_BY_HOP:
+                header_name = key.lower()
+                drop_stale_content_encoding = (
+                    transform_model_names and header_name == "content-encoding"
+                )
```
lilly-luo, line 409: what does this mean? why can't we have content-encoding?

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -395,11 +498,17 @@ def start_proxy(
     # to the gateway instead of a fresh handshake per request. Don't follow
     # redirects — a proxy relays 3xx verbatim.
     client = httpx.Client(base_url=upstream_base, timeout=_UPSTREAM_TIMEOUT, follow_redirects=False)
```
lilly-luo, line 504: hmmm is upstream_base then changing? because it used to be /anthropic. will this break relayed auth?

lilly-luo: ic, is that why you had /v1/models as the orig path to match one? maybe we should leave it as it was before then and just rename to anthropic_gateway_proxy since it looks like this thing is pretty anthropic specific

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -217,7 +295,11 @@ def _handle(self) -> None:
         started = time.monotonic()
         length = int(self.headers.get("Content-Length", 0) or 0)
         body = self.rfile.read(length) if length else None
-        url = self.path.lstrip("/")
+        body = self.anthropic_model_aliases.rewrite_body(self.path, body)
+        url = self.anthropic_model_aliases.rewrite_path(self.path).lstrip("/")
```
lilly-luo, line 299: will url here be diff btw relayed auth and the models?

PR #380, opened 2026-08-24
MPS: launch on a servable default model, and honor --model under --provider

masonc08:
## Problem

`ucode claude --provider <mps>` always started on Claude Code's built-in "family default" (opus). When a Model Provider Service declares no opus target, the gateway 403s that default — so a bare launch failed for any service that only offers sonnet/haiku (e.g. an API-key Anthropic MPS scoped to sonnet+haiku). `--model` couldn't rescue it either: `--model` + `--provider` was rejected outright with *"Use either --model or --provider, not both."*

Stacked on #376 (which pins the declared per-family targets); this makes the *launch* model land on one of them.

## Change

Under a provider:
- **Auto-pick the best servable tier.** With no `--model`, choose the most capable tier the service actually declares (opus > sonnet > haiku) and pin it via `ANTHROPIC_MODEL`, so the session starts on a servable model instead of dead-ending on opus. When the service *does* offer opus, pin nothing — Claude Code's own default already works, and this avoids the duplicate `/model` picker row `ANTHROPIC_MODEL` produces.
- **Honor `--model`** (guard removed). A family alias (`opus`/`sonnet`/`haiku`) resolves to that tier's declared target (clear error if the service doesn't offer it); any other value is trusted as a raw target id. Routed via `route_root_model`, not `custom_model`, so it doesn't clobber the per-family pins from #376.
- **Relayed services**: model selection is server-side, so `--model` is ignored with a warning.

The launch banner now also prints the chosen `Model:` under a provider.

## Known gap (deferred)

For an **opus-less Bedrock** service the region-prefixed slug can be rejected by `ANTHROPIC_MODEL`'s client-side check. That case was already broken (bare launch 403s on opus), and opus-having Bedrock services are unaffected (they return `None` and behave as today). The follow-up fix is to pin the servable target into the opus **family slot** instead — that channel is passed through unchecked.

## Testing

- Unit: `resolve_provider_launch_model` (7 cases: opus-present no-op, sonnet/haiku fallback, alias→target, unavailable-tier error, raw-id passthrough); CLI provider-launch (alias pin, auto-default, opus-present no-op, unavailable-tier error, relayed-ignored).
- Suites green: `test_databricks.py` (248), `test_cli.py` (218), `test_agent_claude.py` + `test_agents_init.py` (188). `ruff` clean.
- End-to-end against an API-key Anthropic MPS (canonical target names): recommended sanity check — bare launch starts on sonnet with no 403; `--model haiku` starts on haiku; `--model opus` errors clearly.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -2015,18 +2015,6 @@ def _launch_tool(
             # provider). Skip model resolution, which would otherwise fail when
             # the workspace has no matching Databricks models.
             resolved_model = None
-            # Claude Code starts on its built-in "family default" (opus), which the gateway 403s when
-            # the service declares no opus target. Pick the launch model explicitly and pin it via
-            # ANTHROPIC_MODEL (route_root_model): the user's --model when given, else the most capable
-            # tier the service actually offers. A relayed service selects the model server-side, so
-            # there's nothing to pin — and --model can't be honored, so say so rather than ignore it.
-            #
-            # KNOWN GAP (deferred): ANTHROPIC_MODEL is checked client-side, so this covers services
-            # whose targets are canonical Anthropic names (an API-key Anthropic service); a Bedrock
-            # service's region-prefixed slug can be rejected there. Only an opus-less Bedrock service
-            # hits this — an opus-having one returns None above and is unaffected — and that case was
-            # already broken (bare launch 403s on opus). The follow-up fix is to pin the servable
-            # target into the opus family slot instead (that channel is passed through unchecked).
             if tool == "claude" and relayed:
                 if model:
                     print_warning(
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -55,6 +55,7 @@ from ucode.databricks import (
     list_tool_provider_services,
     normalize_workspace_url,
     resolve_pat_token,
+    resolve_provider_launch_model,
     run_databricks_login,
 )
 from ucode.managed_budget import (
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -1875,6 +1875,26 @@ def _launch_tool(
...
+            # ANTHROPIC_MODEL (route_root_model): the user's --model when given, else the most capable
+            # tier the service actually offers. A relayed service selects the model server-side, so
+            # there's nothing to pin — and --model can't be honored, so say so rather than ignore it.
+            #
+            # KNOWN GAP (deferred): ANTHROPIC_MODEL is checked client-side, so this covers services
+            # whose targets are canonical Anthropic names (an API-key Anthropic service); a Bedrock
+            # service's region-prefixed slug can be rejected there. Only an opus-less Bedrock service
+            # hits this — an opus-having one returns None above and is unaffected — and that case was
+            # already broken (bare launch 403s on opus). The follow-up fix is to pin the servable
+            # target into the opus family slot instead (that channel is passed through unchecked).
```
lilly-luo: can u rm all this

lilly-luo (approved):

src/ucode/databricks.py
```diff
@@ -2345,6 +2345,46 @@ def map_claude_family_models(targets: list[str]) -> dict[str, str]:
     return result
 
 
+# Claude Code starts every session on its opus tier, which the gateway 403s when a Model Provider
+# Service declares no opus target. When opus is missing, fall back to the most capable tier the
+# service does offer. opus > sonnet > haiku.
+_CLAUDE_LAUNCH_TIER_PREFERENCE = ("opus", "sonnet", "haiku")
```
lilly-luo, line 2406: no fable?

PR #388, opened 2026-08-25
Let users choose a usage SQL warehouse

andy-xu-db:
## Summary

- list the SQL warehouses visible to the user before running the detailed usage query
- let the user choose which warehouse runs the query
- if the selected warehouse fails, reopen the picker with the remaining warehouses
- let the user cancel cleanly after a failed selection
- preserve the existing `--warehouse-id` override without prompting

## Testing

- `uv run pytest tests/test_usage.py` (89 passed)
- `uv run ruff check src/ucode/usage.py tests/test_usage.py`

https://github.com/user-attachments/assets/e8239382-cc69-4962-bb21-35d2a8990cc1


lilly-luo:
can you merge main and include a video of it working in the PR desc please?

lilly-luo (approved):

PR #392, opened 2026-08-25
[AIGTWY-4290] Reconcile OS managed settings on workspace switch

david-siqi-liu:
Fixes the OS-managed-file lifecycle gap from #319: when a workspace publishes a managed coding-agent config with use_as_global_settings=true, ucode writes the agent's root-owned OS managed file (Claude: /etc/claude-code/managed-settings.json, Codex: /etc/codex/managed_config.toml). There was no inverse path, so switching to a workspace with no managed config (or one that drops global settings) left the prior workspace's config applying to a bare claude/codex.

What changed:
- managed_files.py: write_managed_file snapshots each OS file's pre-ucode state into a machine-global ledger (~/.ucode/managed-os-backups.json) and commits that record only after a successful sudo write. New restore_managed_file puts the original back, or removes a file ucode created when there was no original. Drift-suppressed (no sudo when the file already matches) and nonfatal (permission or sudo failures warn and never block the private-config launch).
- agents/__init__.py: reconcile_global_settings undoes the write for every global-settings agent the current config no longer marks global, so both claude and codex are reconciled regardless of which one is launched. An agent the config still wants global is left alone; its writer rewrites it.
- cli.py: reconcile runs on launch (_launch_tool), on configure (_resolve_workspace_then_maybe_reject, including its early-return managed branch), and on bare ucode (_launch_managed_default, before its no-config early return). It is gated on the managed-config feature so --skip-managed-config does not tear down a config the workspace still publishes.

Design: this mirrors the reviewer-preferred approach on the #319 threads (save the original and restore it, rather than selectively pruning ucode's keys), which preserves unrelated IT-authored settings.

Known limitations (deliberately out of scope for this ticket):
- Files written by a pre-ledger ucode version cannot be enrolled retroactively, since the true pre-ucode original was never recorded; a machine already carrying an old global write keeps it until the config next changes.
- The ledger is per-user under ~/.ucode, matching state.json. Multi-user machines and concurrent launches share the single machine-global OS file without cross-user coordination, the same as the #319 write path.
- An admin editing the OS file while ucode's config is active will have that edit replaced by the captured original on cleanup, which is the intended restore-original behavior.

Test plan:
- New unit and integration tests cover: the capture/restore lifecycle, Claude and Codex workspace A to B cleanup, original values restored, a ucode-created file removed, unrelated IT keys preserved, cross-agent reconcile (launching one agent reconciles the other), same-workspace reconcile making no privileged call, missing files, sudo failure staying nonfatal, an unreadable original left untouched, a failed write recording no ownership, the feature-disabled path skipping reconcile, unsupported platforms staying safe, and reconcile wired at all three entry points.
- Regression proof: disabling the reconcile call makes the launch wiring test fail.
- Focused suites and the broad non-e2e suite pass locally via the checkout venv; CI runs the full unit gate including ruff check, ruff format, and ty.

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -25,6 +25,7 @@ from ucode.databricks import (
     map_claude_family_models,
     resolve_provider_service,
 )
+from ucode.managed_files import restore_managed_file
 from ucode.state import get_provider_service, load_state, save_state
 from ucode.telemetry import agent_version
 from ucode.ui import (
@@ -87,6 +88,32 @@ AITOOLS_AGENT_TOKENS = {
     "opencode": "opencode",
     "copilot": "copilot",
 }
+
+
+def _os_managed_file_path(tool: str):
+    if tool == "claude":
+        return claude._managed_settings_path()
+    if tool == "codex":
+        return codex._managed_config_path()
+    return None
+
+
+def reconcile_global_settings(managed: dict | None) -> None:
+    """Undo ucode's OS-managed-file write for each global-settings agent the current config no longer
+    marks global, so switching workspaces stops a prior one's settings applying to a bare agent.
+
+    Reconciles both agents, not just the launched one; an agent still marked global is left for its
+    writer to (re)write. Idempotent, drift-suppressed, and nonfatal. ``managed_use_as_global_settings``
+    is imported locally to avoid a cycle (managed_resolve imports ``GLOBAL_SETTINGS_AGENTS`` from here).
+    """
+    from ucode.managed_resolve import managed_use_as_global_settings
+
+    for tool in GLOBAL_SETTINGS_AGENTS:
+        if managed_use_as_global_settings(managed or {}, tool):
+            continue
+        path = _os_managed_file_path(tool)
+        if path is not None:
+            restore_managed_file(path, display=TOOL_SPECS[tool]["display"])
 
 
 def install_databricks_ai_tools_for_agents(tools: list[str], state: dict) -> None:
--- a/src/ucode/managed_files.py
+++ b/src/ucode/managed_files.py
@@ -17,14 +17,17 @@ first write (or one after the config changes) prompts for the developer's sudo p
 
 from __future__ import annotations
 
+import json
 import os
 import shlex
 import subprocess
 import sys
 import tempfile
+from collections.abc import Callable
 from enum import Enum
 from pathlib import Path
 
+from ucode import config_io
 from ucode.config_io import is_dry_run
 from ucode.ui import console, print_err, print_warning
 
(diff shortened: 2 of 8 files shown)


src/ucode/agents/__init__.py
```diff
@@ -89,6 +90,32 @@
...
+
+def reconcile_global_settings(managed: dict | None) -> None:
+    """Undo ucode's OS-managed-file write for each global-settings agent the current config no longer
+    marks global, so switching workspaces stops a prior one's settings applying to a bare agent.
+
+    Reconciles both agents, not just the launched one; an agent still marked global is left for its
+    writer to (re)write. Idempotent, drift-suppressed, and nonfatal. ``managed_use_as_global_settings``
+    is imported locally to avoid a cycle (managed_resolve imports ``GLOBAL_SETTINGS_AGENTS`` from here).
+    """
+    from ucode.managed_resolve import managed_use_as_global_settings
```
lilly-luo, line 109: import top of file

src/ucode/managed_files.py
```diff
@@ -186,3 +191,156 @@ def _report_sudo_failure(path: Path, display: str, exc: subprocess.CalledProcess
...
+def _forget_original(key: str) -> None:
+    if is_dry_run():
+        return
+    ledger = _read_ledger()
+    if key in ledger:
+        del ledger[key]
+        _write_ledger(ledger)
+
+
+def restore_managed_file(path: Path, *, display: str) -> str:
```
lilly-luo, line 279: this method is kind of complicated. can you include steps of what's happening? 

1/ xxx
2/ if x, then y. ...
3/ ...

src/ucode/managed_files.py
```diff
@@ -186,3 +191,156 @@ def _report_sudo_failure(path: Path, display: str, exc: subprocess.CalledProcess
...
+    _write_ledger(ledger)
+
+
+def _forget_original(key: str) -> None:
+    if is_dry_run():
+        return
+    ledger = _read_ledger()
+    if key in ledger:
+        del ledger[key]
+        _write_ledger(ledger)
```
lilly-luo, line 276: should we ensure that _write_ledger succeeded before deleting?

src/ucode/managed_files.py
```diff
@@ -186,3 +191,156 @@ def _report_sudo_failure(path: Path, display: str, exc: subprocess.CalledProcess
...
+def _sudo_snapshot(path: Path) -> dict:
+    """Determine ``path``'s pre-ucode state with sudo, for a root-locked parent or root-only file.
+
+    A file sudo confirms absent is recorded ``existed=False`` so cleanup removes the file ucode is
+    about to create; one readable only as root is captured verbatim; one unreadable even as root stays
+    ``existed=True/contents=None`` (left untouched on cleanup). A sudo failure is harmless: the write
+    that follows fails the same way, so this snapshot is never recorded.
+    """
+    present = subprocess.run([_SUDO, "test", "-e", str(path)], capture_output=True, check=False)
+    if present.returncode != 0:
```
lilly-luo, line 251: currently this is used to identify absence of a file. will this also trigger if i don't have access to the file? if yes, does that break anything?

PR #393, opened 2026-08-25
smart routing: route cc with /model

lilly-luo:
## Summary
- route the first Claude prompt by typing `/model <model>` directly instead of navigating the model picker
- snapshot and surgically restore only `~/.claude/settings.json`'s `model` field, with recovery journals for interrupted launches
- launch `ucode claude` with the saved default via `--model`, while preserving explicit and routed model precedence

architecture: https://docs.google.com/document/d/1lO1l3sOH0V9_YIx2SQQIvDOBlBJfIwpITLFl1rYTZ7s/edit?tab=t.whcgscnu6a6r, "Technical Implementation", "Smart Router flow on first prompt" 

https://github.com/user-attachments/assets/bca4a3a5-b614-4bfb-9359-1efd991609a4

## Testing
- [x] my first prompt routes to the hardcoded sonnet
- [x] i start out as haiku. after my first prompt routes to the hardcoded sonnet, if i open a new session w/ smart routing, it's haiku. 
- [x] after my first prompt routes to sonnet, i can change the model via /model. 
- [x] after my first prompt routes to sonnet, i can change the model via /model and it stick for the next session 
- [x] normal ucode claude works. /model changes my default model there. 
- [x] model discovery works 
- [x] running /model before the first prompt
- [x] starting out w/ sonnet and routing to sonnet


lilly-luo:
LL: make sure that /model works with smart routing

src/ucode/agents/claude.py
```diff
@@ -1077,6 +1165,31 @@ def launch(state: dict, tool_args: list[str]) -> None:
...
+        and _uses_interactive_tui(tool_args)
+    )
+    # Smart routing v2 needs Unix PTY support, which Windows does not provide.
+    if first_prompt_routing and os.name == "nt":
+        raise RuntimeError(
+            "Smart routing in Claude Code is currently not supported on Windows. "
+            "Please use Codex or disable smart routing."
+        )
+    if first_prompt_routing:
+        smart_routing_v2.launch_claude(
```
andy-xu-db: Would this mean that claude is launched separate from the gateway proxy, so model discovery wouldn't work?

lilly-luo: yes great catch. i didn't see this in testing because my tokens were 1hr and because all the models in the picker have claude in them (so no translation)

andy-xu-db (approved):

PR #395, opened 2026-08-26
Claude: separate model discovery proxy

andy-xu-db:
## Summary
- keep token refresh, retry, forwarding, and streaming in the shared gateway proxy
- move Anthropic model prefixing and stripping into a discovery handler subclass
- use the discovery subclass for both standard and relayed Claude launches

## Testing
- `uv run --frozen ruff check src/ tests/`
- `uv run --frozen ty check src/`
- `uv run --frozen pytest tests/test_gateway_proxy.py tests/test_anthropic_model_discovery_proxy.py tests/test_agent_claude.py tests/test_lint.py -q`
- `uv run --frozen pytest -q --ignore=tests/test_e2e_user_agent.py` (1951 passed, 37 skipped)

e2e demo

https://github.com/user-attachments/assets/aa74af67-537e-493f-894c-796fd0675d7f

Stacked on #375.


src/ucode/anthropic_model_discovery_proxy.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def start_proxy(
+    workspace: str,
+    profile: str | None,
+    port: int,
+    token_header: str,
+    force_refresh_near_expiry: bool,
+):
+    return gateway_proxy._start_proxy(
```
lilly-luo: generally _ prefix is saved for private methods that aren't called outside of the file. can you use the public start_proxy instead? then you can override _start_proxy in the original gateway_proxy one and invoke it in the public start_proxy

src/ucode/gateway_proxy.py
```diff
@@ -304,9 +312,15 @@ def _relay_response(
         bytes_relayed = 0
         first_byte_ms: int | None = None
         try:
+            transformed_body = self._transform_response(resp)
```
lilly-luo: how can you transform the response if the req hasn't been sent yet?

src/ucode/gateway_proxy.py
```diff
@@ -315,7 +329,10 @@ def _relay_response(
             # With ``chunk_size=None`` (the default), raw upstream chunks are
             # yielded as they arrive and pings keep the downstream connection
             # alive even before the model produces a large content block.
-            for chunk in resp.iter_raw():
+            response_chunks = (
+                [transformed_body] if transformed_body is not None else resp.iter_raw()
+            )
+            for chunk in response_chunks:
```
lilly-luo: this was the cause of issues for relayed auth that customer reported, we should be careful changing this. also does this mean that if a relayed auth request is sent for inference on a session that has model discovery enabled, we wouldn't be doing resp.iter_raw?

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -0,0 +1,518 @@
...
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
+
+import httpx
+
+from ucode.constants import LOOPBACK_HOST
+from ucode.databricks import get_databricks_token
+
+# Header we overwrite with the freshly-minted Databricks credential. Any
+# client-supplied value is replaced, so a stale settings.json value can't leak.
+AI_GATEWAY_TOKEN_HEADER = "X-Databricks-AI-Gateway-Token"
```
lilly-luo: do we need this?

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -0,0 +1,518 @@
...
+        "keep-alive",
+        "proxy-authenticate",
+        "proxy-authorization",
+        "te",
+        "trailers",
+        "transfer-encoding",
+        "upgrade",
+        "host",
+        "content-length",
+    )
```
lilly-luo: do we need this?

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def _diagnostic_log(event: str, **fields: object) -> None:
+    if not _diagnostics_enabled():
+        return
+    payload = {"event": event, **fields}
+    sys.stderr.write(
+        f"[ucode-relay] {json.dumps(payload, sort_keys=True, separators=(',', ':'))}\n"
+    )
+    sys.stderr.flush()
```
lilly-luo: can this be shared btw the 2 proxies?

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -0,0 +1,358 @@
...
+                    return
+                # Auth rejected. Drain the (small) error body so the pooled
+                # connection can be reused, then fall through to one retry.
+                resp.read()
+            # A relayed 401/403 may be a stale Databricks swap token rather than a
+            # bad Anthropic OAuth — the two are indistinguishable from the status
+            # alone. Force-refresh the Databricks token and retry once. If it was the
+            # Anthropic layer, the retry still 401s and we relay it verbatim, so a
+            # genuine re-auth is triggered; a stale-Databricks 401 self-heals here
+            # instead of surfacing to Claude Code as a spurious Anthropic prompt.
```
lilly-luo: is this relevant? we're never doing anthropic oauth with this right, we're just talking to databricks. maybe just rm

src/ucode/anthropic_gateway_proxy.py
```diff
@@ -0,0 +1,358 @@
...
+            request_id=diagnostic_id,
+            method=self.command,
+            path=self.path.split("?", 1)[0],
+        )
+        try:
+            # First attempt with the current token.
+            headers = forwarded_request_headers(self, self.cache.token, self.token_header)
+            with self.client.stream(self.command, url, headers=headers, content=body) as resp:
+                log_proxy_diagnostic(
+                    "upstream_headers",
```
lilly-luo: can you give these a diff prefix so that we can identify it's from model discovery, not from the relayed auth? same w/ all the other labels

src/ucode/anthropic_model_discovery_proxy.py
lilly-luo, line 1: nit anthropic_gateway_proxy -> anthropic_model_discovery_proxy

lilly-luo (approved):

lilly-luo:
could you put a video of the e2e in the PR desc?

PR #398, opened 2026-08-26
claude: add model discovery launch flag

andy-xu-db:
## Summary

- add `ucode claude --enable-model-discovery`
- enable the existing Claude gateway model discovery path for that launch
- keep the flag from being forwarded to Claude Code

## Tests

- `uv run pytest tests/test_cli.py -q`
- `uv run pytest tests/test_agent_claude.py -q`
- `uv run ruff check .`
- `uv run ty check src/`

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -2213,6 +2213,7 @@ def claude_cmd(
         bool,
         typer.Option(
             "--enable-model-discovery",
+            hidden=True,
             help="Enable AI Gateway models in Claude Code's model picker.",
         ),
     ] = False,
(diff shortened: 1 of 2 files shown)


src/ucode/cli.py
```diff
@@ -2209,6 +2209,13 @@ def claude_cmd(
     skip_preflight: SkipPreflightOption = False,
     skip_managed_config: SkipManagedConfigOption = False,
     workspace: WorkspaceOption = None,
+    enable_model_discovery: Annotated[
+        bool,
+        typer.Option(
+            "--enable-model-discovery",
+            help="Enable AI Gateway models in Claude Code's model picker.",
```
lilly-luo, line 2217: is it poss to avoid adding this to the help menu but still make it work?

lilly-luo (approved):

PR #401, opened 2026-08-27
Add first-prompt routing to Codex interposer

lilly-luo:
## Summary
- route the first Codex prompt through task_v1 using the existing workspace token
- source all route options from cached Unity Catalog model services and mark them with the codex harness
- rewrite the app-server turn to the selected full system.ai model and log routing requests without credentials

## Testing
- 569 focused tests passed
- ruff checks passed
- full suite: 1,988 passed and 37 skipped; remaining failures require loopback sockets unavailable in the sandbox

### DIFF
--- a/src/ucode/smart_routing/v2.py
+++ b/src/ucode/smart_routing/v2.py
@@ -123,8 +123,8 @@ def launch_codex(
             "Smart routing v2 could not determine a starting Codex model for this workspace."
         )
 
-    token = get_databricks_token(workspace, state.get("profile"))
-    os.environ[OAUTH_TOKEN_ENV_VAR] = token
+    profile = state.get("profile")
+    os.environ[OAUTH_TOKEN_ENV_VAR] = get_databricks_token(workspace, profile)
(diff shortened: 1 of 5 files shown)


src/ucode/smart_routing/v2.py
```diff
@@ -114,7 +123,14 @@ def launch_codex(
             "Smart routing v2 could not determine a starting Codex model for this workspace."
         )
 
-    os.environ[OAUTH_TOKEN_ENV_VAR] = get_databricks_token(workspace, state.get("profile"))
+    token = get_databricks_token(workspace, state.get("profile"))
```
andy-xu-db: Would token expiration ever be an issue here?

andy-xu-db (approved):

src/ucode/smart_routing/v2.py
lilly-luo: yes - if i open my tui and leave it for an hour, then try to type something, it would fail. 

fixing to pass a lambda for token refresh. great catch

PR #404, opened 2026-08-27
Role-aware managed-config paths in `ucode configure` (AIGTWY-4338)

tt-le:
Make `ucode configure` branch on the caller's role and whether the workspace already publishes a managed coding-agent config. All new behavior is gated by ENABLE_MANAGED_AGENT_CONFIG; with the flag unset, `configure` runs the existing manual flow unchanged.

Once the target workspace is resolved and authenticated (up front, so the config read is fresh and admin detection is reliable on an expired session):

|Role | Managed Config Status | UX |
|-----|-------------------------|----|
|User | N/A | This opens the regular configure flow Screenshot 2026-08-27 at 6 43 36 PM|
|User | Present | This just shows the user the managed config and hints them to use `ucode` to start Screenshot 2026-08-27 at 6 30 58 PM|
|Admin | Present | This essentially just calls ucode setup but relabeled to ucode configuration Screenshot 2026-08-27 at 6 28 19 PM Screenshot 2026-08-27 at 6 28 44 PM|
|Admin | N/A | This also just calls ucode setup but goes straight into the setup, same as current ucode setup if no configs were found. Screenshot 2026-08-27 at 6 27 39 PM|

Also: the workspace picker and the setup flow's section headers now brand themselves to the invoking command ("ucode configure" vs "ucode setup") via a threaded label, so a configure user never sees stray "ucode setup" headers. Admin status is a best-effort workspace-admin check (SCIM `Me` group membership), mirroring the server-side authority; failures fall through safely.

Tests cover all four role/config paths, the flag-off no-op, the auth ordering, the command-label branding, and setup's Adopt option.

On non enabled workspaces:
As workspace admin: 
Screenshot 2026-08-28 at 11 44 50 AM
Non workspace admin:
image


lilly-luo:
can you test against a workspace that doesn't have the functionality enabled yet and put a screenshot? if the workspace does not have the functionality enabled yet due to server side flag, then the admin flow should be same as non-admin flow.

src/ucode/cli.py
```diff
@@ -340,14 +391,18 @@ def _print_discovery_diagnostics(state: dict) -> None:
     print_note("Re-run with `UCODE_DEBUG=1` to log raw discovery responses to ~/.ucode/debug.log.")
 
 
-def _prompt_for_configuration(tool: str | None = None) -> tuple[str, str | None]:
+def _prompt_for_configuration(
+    tool: str | None = None, *, title: str = "ucode configure"
```
lilly-luo: Image

can you change to "Configure Unity Gateway"

src/ucode/ui.py
```diff
@@ -344,6 +344,8 @@ def normalize_workspace_url(workspace: str) -> str:
 def prompt_for_workspace(
```
lilly-luo, line 344: prompt_for_workspace is a bit generic of a name. can you either make the func name specific to setup or remove the default from the title param (or both)?

src/ucode/managed_wizard.py
```diff
@@ -1305,12 +1305,18 @@ def _handle_existing_config(workspace: str, token: str) -> tuple[bool, dict | No
     choice = prompt_for_selection(
         "What would you like to do?",
         [
+            ("adopt", "Adopt the published config as-is (it applies when you run `ucode`)"),
```
lilly-luo: Adopt the published config as-is (it applies when you run `ucode`) -> Adopt the published config as your current settings. To invoke, run `ucode`.

src/ucode/cli.py
```diff
@@ -253,29 +253,45 @@ def _print_managed_summary_abridged(managed: dict, state: dict, tool: str | None
     )
 
 
+def _confirm_managed_config_in_force(managed: dict, workspace: str) -> None:
+    """Confirm the workspace's managed config is already in force, and show what it holds.
+
+    There is nothing to write locally: the launch path applies the config (agents, MCP servers, and
+    skills) on every ``ucode`` run. So both the non-admin ``configure`` path and setup's "Adopt"
+    choice land here — show the admin's config and point the user at ``ucode``.
+    """
```
lilly-luo: rm all this

src/ucode/cli.py
```diff
@@ -253,29 +253,45 @@ def _print_managed_summary_abridged(managed: dict, state: dict, tool: str | None
...
+    print_success("A managed config is published for your workspace — you're all set.")
+    _print_managed_summary(managed, {"workspace": workspace}, tool=None)
+    print_note("Configuration is complete. Just run `ucode` to launch with it applied.")
```
lilly-luo: Just run `ucode` to launch with it applied."
-> 
Run `ucode` to launch with your managed settings.

src/ucode/cli.py
```diff
@@ -253,29 +253,45 @@ def _print_managed_summary_abridged(managed: dict, state: dict, tool: str | None
...
+    * **No managed config** → a workspace admin is dropped straight into the ``ucode setup``
+      authoring flow (``configure`` is replacing ``setup``) and the command exits with its code; a
+      non-admin's own ``configure`` proceeds, with the resolved entries returned so the caller
+      reuses them instead of re-prompting.
+    * **Managed config, non-admin** (or admin status unverifiable) → they're already set: the
+      launch path applies the config on every ``ucode`` run, so just show it and point them there.
+    * **Managed config, admin** → drop into the setup flow, whose existing-config menu lets them
+      adopt it (the same "you're all set" confirmation), re-author it, or delete it; the command exits.
+
+    Without the feature enabled it returns ``workspace_entries`` unchanged and prompts nothing.
```
lilly-luo: can you specify whether feature enablement is client-side, server-side, or both?

lilly-luo (approved):

src/ucode/ui.py
tt-le: Removed the title param

PR #407, opened 2026-08-28
Speed up Claude and Codex launches

rohita5l:
## Summary
- skip repeated preflight, gateway detection, model discovery, and config rewriting for configured Claude and Codex launches
- add `--refresh` to explicitly refresh Databricks auth, gateway, models, managed config, and agent configuration
- install Databricks AI Tools only from `ucode configure`; agent launch commands never install them
- preserve full setup paths for providers, managed/local config, smart routing, and first-time setup

## Testing
- affected agent/CLI tests: 299 passed
- Ruff, formatting, and diff checks passed
- full suite: 1961 passed, 37 skipped; one unrelated Claude binary e2e test remains environment-dependent


src/ucode/agents/claude.py
```diff
@@ -600,7 +578,16 @@ def _compose(base: dict) -> dict:
...
     if web_search_model:
-        _register_web_search_mcp(state["workspace"], web_search_model, state.get("profile"))
+        web_search_entry = _web_search_mcp_entry(
+            state["workspace"], web_search_model, state.get("profile")
+        )
+        if not _web_search_mcp_is_current(state, web_search_entry):
+            if _register_web_search_mcp(state["workspace"], web_search_model, state.get("profile")):
+                state[WEB_SEARCH_MCP_STATE_KEY] = web_search_entry
+        else:
+            state[WEB_SEARCH_MCP_STATE_KEY] = web_search_entry
```
lilly-luo: this is redundant right? if the mcp is current, you dont need to reset it

src/ucode/agents/claude.py
```diff
@@ -600,7 +578,16 @@ def _compose(base: dict) -> dict:
...
+    else:
+        state.pop(WEB_SEARCH_MCP_STATE_KEY, None)
```
lilly-luo, line 630: nit for easier reading 

```
if web_search_model: 
   web_search_entry = _web_search_mcp_entry(
            state["workspace"], web_search_model, state.get("profile")
        )
    if not _web_search_mcp_is_current(state, web_search_entry):
            # _register_web_search_mcp runs multiple `claude mcp` sub processes and can take 0.8s
            _registration_success =  _register_web_search_mcp(state["workspace"], web_search_model, state.get("profile")):
            if _registration_success:
                state[WEB_SEARCH_MCP_STATE_KEY] = web_search_entry
else:
    state.pop(WEB_SEARCH_MCP_STATE_KEY, None)
```

src/ucode/cli.py
```diff
@@ -1723,10 +1597,43 @@ def _apply_managed_skills(managed: dict, tool: str, state: dict) -> None:
...
+    skip_preflight: bool,
+    workspace: str | None,
+    needs_auto_configure: bool,
+) -> bool:
+    """Return whether a normal Claude/Codex launch can trust local config."""
+    if tool not in ("claude", "codex"):
+        return False
+    if refresh or model or explicit_provider is not None:
+        return False
+    if enable_smart_routing_flag or skip_preflight:
```
lilly-luo: ? if we're only using the fast path when smart routing is disabled, this means smart routing is going to miss out on the speed up? isn't the goal to make smart routing the default on isaac so it means it would be slow?

src/ucode/cli.py
```diff
@@ -2184,6 +2091,14 @@ def codex_cmd(
             "before any `--` separator.",
         ),
     ] = None,
+    refresh: Annotated[
+        bool,
+        typer.Option(
+            "--refresh",
+            help="Refresh Databricks auth, gateway, models, managed config, and Codex configuration "
```
lilly-luo: nit put this help into a const and either drop or parameterize Codex / Claude btw the both

src/ucode/agents/claude.py
lilly-luo: fixed

src/ucode/cli.py
lilly-luo: done

src/ucode/cli.py
lilly-luo: this is for legacy smart routing, not the v2 thing w/ pty. removed skip preflight tho

src/ucode/agents/claude.py
lilly-luo: yes, rm'd

lilly-luo (approved):

PR #408, opened 2026-08-28
smart routing: route Claude subagents 

lilly-luo:
## Summary

- initialize a Claude Code `PreToolUse` hook for `Agent` calls when `ENABLE_SMART_ROUTING_V2=1`
- pass the discovered compatible Anthropic model menu and subagent prompt to Unity Gateway smart routing
- override the subagent tool input with the selected Claude model
- preserve the existing legacy Claude routing path when V2 is disabled

https://docs.google.com/document/d/1lO1l3sOH0V9_YIx2SQQIvDOBlBJfIwpITLFl1rYTZ7s/edit?tab=t.whcgscnu6a6r

## Testing

- `uv run pytest tests/test_claude_smart_routing_v2.py tests/test_claude_routing.py tests/test_cli.py -q`
- `uv run ruff check .`

Screenshot 2026-08-28 at 3 53 54 PM

logs:
```
{"decision_id": "63fe7743a93e444d867f43e4669e9b88", "session_id": "592c41aa-8cf8-4e1d-966a-92fc10d64598", "task_name": "There is a config loader in this repository (working directory /home/lilly.luo/ucode, but search the whole repo \u2014 it may be elsewhere) that parses config files of the form `key = value`. The bug: when it parses a line, it strips trailing whitespace from the value before storing it, so a setting like `token = abc123 ` (with a trailing space) gets stored as `abc123`, which no longer matches the literal value the user typed, breaking downstream equality checks.\n\nYour task:\n1. Locate the config loader / config-file parsing code. Search for where lines are split on a delimiter (`=` or similar) and where `.strip()` / `.rstrip()` / trimming is applied to the value.\n2. Fix the trimming logic so the value preserves the EXACT characters between the delimiter and the end of the line, including any trailing spaces. \n   - IMPORTANT: Preserve existing behavior for LEADING whitespace (i.e. if leading whitespace after the delimiter is currently stripped, keep stripping it; the fix is only about trailing whitespace).\n   - Do NOT change behavior for keys or any other fields \u2014 only the value's trailing-whitespace handling.\n3. Add a focused regression test that covers a value with trailing spaces (assert the stored value still contains the trailing spaces exactly). Follow the existing test conventions in the repo (this repo uses pytest; run with `uv run pytest`).\n4. Run the relevant tests to confirm the fix works and nothing else broke (`uv run pytest tests/<the_file>.py`, and consider running the loader-related tests).\n\nReport back: the exact file(s) and line(s) you changed, a diff of the change, the test you added, and the test run output. If you cannot find a config loader that does trailing-whitespace trimming, report what you found instead of guessing \u2014 do not invent a loader.", "router_model": "claude-sonnet-5", "requested_model": "system.ai.claude-sonnet-5", "rationale": "Routed to claude-sonnet-5 because [not_crosscutting AND not_mixed_change AND low_ambiguity] not all hold -> default claude-sonnet-5.", "at": 1787946656.952682}
``` 

audit log: 
```
{"agent_id": "a4bb6a9724d8df438", "agent_type": "ucode-route-claude-sonnet-5-413de6c6", "model": null, "session_id": "592c41aa-8cf8-4e1d-966a-92fc10d64598", "at": 1787946657.2221286, "decision_id": "63fe7743a93e444d867f43e4669e9b88", "router_model": "claude-sonnet-5", "requested_model": "system.ai.claude-sonnet-5", "matches_router_decision": null}
```

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -118,7 +118,7 @@ WEB_SEARCH_MCP_NAME = "web_search"
 # Matches both the AI Gateway form (`databricks-claude-opus-4-8`) and the UC
 # model-services form (`system.ai.claude-opus-4-8`).
 _CLAUDE_MODEL_RE = re.compile(
-    r"^(?:system\.ai\.)?(?:databricks-)?claude-(opus|sonnet)-(\d+)-(\d+)(.*)$"
+    r"^(?:system\.ai\.)?(?:databricks-)?claude-(opus|sonnet)-(\d+)(?:-(\d+))?(.*)$"
 )
 
 # Env keys the MLflow Stop hook reads to route traces. Written into the
@@ -438,7 +438,7 @@ def _maybe_add_1m_suffix(model: str) -> str:
 
     family, major_raw, minor_raw, _ = match.groups()
     major = int(major_raw)
-    minor = int(minor_raw)
+    minor = int(minor_raw or 0)
     should_suffix = (family == "opus" and (major, minor) >= (4, 6)) or (
         family == "sonnet" and (major, minor) >= (4, 6)
     )
@@ -1188,6 +1188,7 @@ def _launch_claude_with_gateway_proxy(
                 launch_model=_original_launch_model(state),
                 compose_settings=compose_gateway_settings,
                 launch_model_args=_launch_model_args,
+                model_name=_maybe_add_1m_suffix,
             )
             return
 
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1480,13 +1480,22 @@ def claude_router_hook_cmd(
             token = get_databricks_token(host, profile)
         except RuntimeError:
             return
-    output = route_pre_tool_use(
-        payload,
-        workspace=host,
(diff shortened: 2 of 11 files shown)


src/ucode/agents/claude.py
```diff
@@ -118,7 +118,7 @@ def _resolve_web_search_model(state: dict) -> str | None:
 # Matches both the AI Gateway form (`databricks-claude-opus-4-8`) and the UC
 # model-services form (`system.ai.claude-opus-4-8`).
 _CLAUDE_MODEL_RE = re.compile(
-    r"^(?:system\.ai\.)?(?:databricks-)?claude-(opus|sonnet)-(\d+)-(\d+)(.*)$"
+    r"^(?:system\.ai\.)?(?:databricks-)?claude-(opus|sonnet)-(\d+)(?:-(\d+))?(.*)$"
```
lilly-luo, line 121: this makes the minor version optional. so claude-sonnet-5 works. 

we don't add fable/haiku becauase they don't support 1m context

andy-xu-db (approved):

PR #409, opened 2026-08-28
smart routing: route Codex v2 subagents

lilly-luo:
## Summary
- initialize the Codex `PreToolUse` spawn-agent routing hook when `ENABLE_SMART_ROUTING_V2=1`
- reuse the existing Unity Gateway routing command and pass all cached compatible models
- preserve existing user `PreToolUse` hooks while layering the V2 hook into app-server config

## Testing
- `.venv/bin/pytest tests/test_codex_smart_routing_v2.py tests/test_codex_routing.py tests/test_agent_codex.py tests/test_lint.py -q`
- `.venv/bin/ruff check .`
- full suite: 2,021 passed; remaining failures/errors require local socket permissions unavailable in the sandbox

https://github.com/user-attachments/assets/c32b788f-ada9-4f32-ba41-0ff758d81178

### DIFF
--- a/src/ucode/smart_routing/codex_routing.py
+++ b/src/ucode/smart_routing/codex_routing.py
@@ -18,6 +18,7 @@ from typing import Any
 from ucode.config_io import APP_DIR
 from ucode.databricks import get_databricks_token
 from ucode.smart_routing import routing
+from ucode.smart_routing.codex_hooks import routing_models
 from ucode.smart_routing.routing import RoutingDecision
 
 ROUTER_NAME = routing.ROUTER_NAME
@@ -50,8 +51,8 @@ def route_launch_model(state: dict, tool_args: list[str]):
     if task is None:
         return None, None
     workspace = state.get("workspace")
-    models = state.get("codex_models")
-    if not isinstance(workspace, str) or not isinstance(models, list):
+    models = routing_models(state)
+    if not isinstance(workspace, str) or not models:
         return None, "workspace model metadata is unavailable"
     try:
         token = get_databricks_token(workspace, state.get("profile"))
--- a/src/ucode/smart_routing/codex_routing.py
+++ b/src/ucode/smart_routing/codex_routing.py
@@ -1,18 +1,20 @@
 """Databricks AI Gateway routing helpers for Codex sessions and subagents.
 
 Codex-specific configuration on top of the shared :mod:`ucode.smart_routing.routing`
-core: the frozen ``task_v1`` Codex route arms, the ``spawn_agent`` tool detector,
-the Codex model-id translation, and the artifact paths.
+core: the workspace-backed ``task_v1`` route options, the ``spawn_agent`` tool
+detector, the Codex model-id translation, and the artifact paths.
 """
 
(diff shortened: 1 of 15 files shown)


src/ucode/smart_routing/codex_routing.py
```diff
@@ -1,31 +1,31 @@
...
 from ucode.smart_routing import routing
+from ucode.smart_routing.codex_hooks import routing_models
 from ucode.smart_routing.routing import RoutingDecision
 
 ROUTER_NAME = routing.ROUTER_NAME
 ROUTING_PATH = routing.ROUTING_PATH
 REQUEST_TIMEOUT_S = routing.REQUEST_TIMEOUT_S
-CODEX_ROUTE_ARMS = ("glm-5-2", "gpt-5-6-sol", "gpt-5-6-luna")
-GLM_ROUTE_ARM = "glm-5-2"
-GLM_GATEWAY_MODEL = "system.ai.glm-5-2"
```
lilly-luo, line 28: im pulling this from model discovery instead now so its dynamic

src/ucode/smart_routing/v2.py
andy-xu-db, line 468: Token expiration might prevent the routing here

src/ucode/smart_routing/v2.py
```diff
@@ -425,12 +425,25 @@ def _codex_config_args(overlay: dict) -> list[str]:
...
+        return Path(codex_home).expanduser() / "config.toml"
+    return Path.home() / ".codex" / "config.toml"
+
+
+def _v2_pre_tool_use_hooks(state: dict, available_models: list[str]) -> list[dict]:
+    doc = read_toml_safe(_codex_home_config_path())
+    configured_hooks = doc.get("hooks")
+    existing = configured_hooks.get("PreToolUse") if isinstance(configured_hooks, dict) else None
+    return merge_pre_tool_use_hooks(
+        existing if isinstance(existing, list) else [],
```
andy-xu-db, line 443: Could the hooks be duplicated here from merging the existing ones with their copy?

andy-xu-db (approved):

src/ucode/smart_routing/v2.py
lilly-luo: yes, cleaned up

src/ucode/smart_routing/codex_routing.py
lilly-luo: Addressed. Codex route options now come from the persisted model-discovery results ( + ) via ; there are no hardcoded route arms.

src/ucode/smart_routing/v2.py
lilly-luo: good catch. we need to centralize the auth token fetch all in 1 place here...

src/ucode/smart_routing/v2.py
lilly-luo: Verified and covered in 64648b7. The CLI  override replaces that config array, so existing user hooks must be copied forward.  removes any existing ucode-marked hook before appending the fresh one; the new regression test asserts exactly one routing hook remains.

PR #413, opened 2026-08-28
Require Claude Code 2.1.248 and remove gateway auth workaround

andy-xu-db:
## Summary

- require Claude Code 2.1.248+, where gateway model discovery works with `apiKeyHelper` as the only credential
- stop setting `ANTHROPIC_AUTH_TOKEN` and stop creating a refresh cache for native gateway model discovery
- keep the model-alias proxy and preserve the existing relayed-subscription token refresh path

## Testing

- `uv run pytest tests/test_agent_claude.py tests/test_anthropic_model_discovery_proxy.py tests/test_gateway_proxy.py -q` (167 passed)
- `uv run pytest tests/test_lint.py -q` (3 passed)
- `uv run ruff check .`
- GitHub CI: unit/lint and workspace e2e jobs passed

### Live end-to-end verification

Verified on `eng-ml-inference-staging` with Claude Code 2.1.251:

- `ucode claude --enable-model-discovery --skip-preflight` launched through the PR model-alias proxy
- the Claude child used a loopback `ANTHROPIC_BASE_URL` and had no `ANTHROPIC_AUTH_TOKEN`
- `/model` displayed gateway models (`system.ai.claude-sonnet-4-6[1m]`, `system.ai.claude-haiku-4-5`, plus 13 more), proving `/v1/models` authenticated via `apiKeyHelper`
- a non-interactive inference returned the exact marker `UCODE_PR_413_OK` with exit code 0

Local full-suite note: the locally installed Claude Code 2.1.251 returned a canned response in `TestClaudeUserAgent::test_user_agent_arrives_at_gateway` without contacting its localhost capture server; the same test passed in GitHub CI.


src/ucode/agents/claude.py
```diff
@@ -1155,18 +1198,17 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
...
-        token_header=AUTHORIZATION_HEADER,
-        force_refresh_near_expiry=True,
+        token_header=None,
+        force_refresh_near_expiry=False,
     )
-    token = cache.token
-    os.environ["OAUTH_TOKEN"] = token
-    os.environ["ANTHROPIC_AUTH_TOKEN"] = token
+    if not smart_routing:
+        os.environ["OAUTH_TOKEN"] = get_databricks_token(workspace, state.get("profile"))
```
lilly-luo: why do you need this? should be orthogonal to smart routing right?

lilly-luo:
the model discovery proxy is now completely separate from the relayed auth proxy now right? model discovery is responsible for patching the model name where as relayed auth swaps headers. in that case, i think it makes sense to fully decouple them.

src/ucode/agents/claude.py
andy-xu-db: ah you're right this can be removed

andy-xu-db:
> the model discovery proxy is now completely separate from the relayed auth proxy now right? model discovery is responsible for patching the model name where as relayed auth swaps headers. in that case, i think it makes sense to fully decouple them.

makes sense, i updated it so model discovery and relayed auth use separate proxies now

lilly-luo (approved):

PR #417, opened 2026-08-30
feat(gemini): route Gemini CLI through a Model Provider Service

variablenerd:
Users need to access Model Provider Services in Unity Gateway via ucode and Gemini CLI. `--provider` routing existed for Claude and Codex but not Gemini, so a Gemini Enterprise Model Provider Service registered in Unity Gateway was unreachable.

This extends the existing `--provider` path to Gemini.

**Source (4 files):**

- **`databricks.py`** — add `gemini_enterprise` to the tool→provider-type map so the service resolves for Gemini; `resolve_provider_models` returns `None` for non-Claude tools (only Claude pins a per-family model map).
- **`agents/gemini.py`** — emit the `Databricks-Model-Provider-Service` header (via `GEMINI_CLI_CUSTOM_HEADERS`) and pin the service's target model; re-emitted on token refresh. `persisted_provider_model()` exposes the pinned target so a bare `ucode gemini` relaunch reuses it.
- **`agents/__init__.py`** — `resolve_gemini_provider_model` picks the target (explicit `--model` → persisted target → sole target → otherwise ask); `_configure_one` resolves a target before `configure_tool`, so `ucode configure` → Gemini → provider no longer crashes.
- **`cli.py`** — `--provider`/`--model` on `ucode gemini`; Gemini skips the Claude/Codex `resolve_provider_models` lookup (single control-plane call); the launch summary prints the pinned target; `ucode configure` offers Gemini the provider picker.

**Tests (4 files):** `test_agent_gemini.py`, `test_agents_init.py`, `test_cli.py`, `test_databricks.py`.

Commands:

```bash
# route Gemini CLI through a Gemini MPS (uses its target model)
ucode gemini --provider {catalog}.{schema}.{mps-name}

# pick a model when the service declares several targets
ucode gemini --provider {catalog}.{schema}.{mps-name} --model gemini-3.5-flash

# persist it as the default
ucode configure          # choose Gemini, then select the provider service
ucode gemini
```

Verified end-to-end against a Gemini Enterprise MPS: the gateway returns `HTTP 200` for `POST /ai-gateway/gemini/v1beta/models/<target>:generateContent` with the `Databricks-Model-Provider-Service` header.

Also addresses review findings: the configure-time crash, bare-relaunch model reuse, a duplicate provider-service lookup, the missing model in the launch summary, and the non-Claude provider-model contract.

Tests: `uv run pytest` (2 pre-existing smart-routing socket tests aside, unrelated to this change), `uv run ruff check .`.

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -344,6 +344,40 @@ def resolve_provider_models(
     return map_claude_family_models(service.get("targets") or []) or None, None, relayed
 
 
+def resolve_gemini_provider_model(
+    state: dict, provider: str, explicit_model: str | None
+) -> tuple[str | None, str | None]:
+    """Pick the Gemini model to pin for a provider-service launch.
+
+    A Gemini Enterprise service routes by header but the request still names a
+    concrete model in the URL, so one of the service's declared targets must be
+    pinned. Uses ``explicit_model`` (from ``--model``) when it names a target;
+    the sole target when the service declares exactly one; otherwise asks the
+    user to choose. Returns ``(model, error)``.
+    """
+    token = get_databricks_token(state["workspace"], state.get("profile"))
+    service, error = resolve_provider_service("gemini", provider, state["workspace"], token)
+    if error or service is None:
+        return None, error or f"Model provider service '{provider}' was not found."
+    targets = [t for t in (service.get("targets") or []) if isinstance(t, str) and t]
+    if explicit_model:
+        if explicit_model in targets:
+            return explicit_model, None
+        available = ", ".join(targets) or "none"
+        return None, (
+            f"Model '{explicit_model}' is not a target of provider service '{provider}'. "
+            f"Available: {available}."
+        )
+    if len(targets) == 1:
+        return targets[0], None
+    if not targets:
+        return None, f"Provider service '{provider}' exposes no models to launch."
+    return None, (
+        f"Provider service '{provider}' exposes several models "
+        f"({', '.join(targets)}); pass --model to choose one."
+    )
+
+
 def configure_tool(
     tool: str,
     state: dict,
@@ -379,7 +413,9 @@ def configure_tool(
         if not model:
             raise RuntimeError(f"A {tool} model must be selected before configuration.")
         if tool == "gemini":
-            result = gemini.write_tool_config(state, model)
+            # Gemini routes through a provider by header (like codex), but must also
+            # pin the service's target model in the URL — `model` already holds it.
+            result = gemini.write_tool_config(state, model, provider=provider)
         elif tool == "copilot":
             result = copilot.write_tool_config(state, model)
         elif tool == "pi":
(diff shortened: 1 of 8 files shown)


src/ucode/agents/__init__.py
```diff
@@ -326,6 +326,40 @@ def resolve_provider_models(
...
+        available = ", ".join(targets) or "none"
+        return None, (
+            f"Model '{explicit_model}' is not a target of provider service '{provider}'. "
+            f"Available: {available}."
+        )
+    if len(targets) == 1:
+        return targets[0], None
+    if not targets:
+        return None, f"Provider service '{provider}' exposes no models to launch."
+    return None, (
```
masonc08, line 395: does it make sense to default to the first target instead of just failing?

src/ucode/agents/__init__.py
```diff
@@ -326,6 +326,40 @@ def resolve_provider_models(
...
+) -> tuple[str | None, str | None]:
+    """Pick the Gemini model to pin for a provider-service launch.
+
+    A Gemini Enterprise service routes by header but the request still names a
+    concrete model in the URL, so one of the service's declared targets must be
+    pinned. Uses ``explicit_model`` (from ``--model``) when it names a target;
+    the sole target when the service declares exactly one; otherwise asks the
+    user to choose. Returns ``(model, error)``.
+    """
+    token = get_databricks_token(state["workspace"], state.get("profile"))
```
masonc08: why do we use a getter for one field but not the other?

src/ucode/agents/__init__.py
masonc08: it can be a followup, of course. we can pick it up on our side

src/ucode/agents/__init__.py
```diff
@@ -359,7 +393,9 @@ def configure_tool(
         if not model:
             raise RuntimeError(f"A {tool} model must be selected before configuration.")
         if tool == "gemini":
-            result = gemini.write_tool_config(state, model)
+            # Gemini routes through a provider by header (like codex), but must also
+            # pin the service's target model in the URL — `model` already holds it.
```
masonc08: comment can be removed

masonc08 (approved):

src/ucode/agents/__init__.py
variablenerd: `workspace` is required for any launch (the launch paths guarantee it — `gemini.py` even raises "No workspace configured" when it is absent), so indexing asserts its presence and fails loud. `profile` is genuinely optional (`None` → the default profile). It also matches the existing convention: `state["workspace"], state.get("profile")` is the pairing used at the other `get_databricks_token` call sites.

src/ucode/agents/__init__.py
variablenerd: Deliberate: the target is a real model choice (cost/latency/capability all differ), so auto-picking `targets[0]` risks launching on a model the user did not intend, whereas failing prints the available targets and the exact `--model` to add. Happy to leave it as a follow-up on your side as you suggested.

src/ucode/agents/__init__.py
variablenerd: Removed — and tidied the now-stale "claude/codex-only" comment just above it. Will be in the force-push after the rebase.

PR #433, opened 2026-08-31
Replace `ucode setup` with a `ucode configure`-centered workflow

david-siqi-liu:

- #438
  - **#433** ⬅ this PR
    - #440

---

Authoring a managed config (`ucode setup`) and applying one (bare `ucode`) were separate entry points
with overlapping behavior, and neither told the user which side of that line they were on. `ucode
configure` becomes the single entry point: it works out whether you are a workspace admin, then
either offers to author a draft or applies the workspace's published config. Authoring has to be
locally testable on the admin's own machine before it reaches anyone else, so the same change splits
`~/.ucode/managed-state.json` into a draft slot and a published slot.

- `ucode setup` and its sub-app are gone. `ucode configure` carries authoring, and `ucode publish`
  promotes the draft to published.
- The state file becomes a versioned map of workspace URL to `{published, draft}`. Authoring and
  `ucode export` read and write `draft` only; a launch refreshes `published` and never touches
  `draft`. Before this, both wrote the one slot, so `refresh_managed_config` on the next launch
  clobbered whatever the admin had just authored, and the admin had no way to run their own draft
  without publishing it first.
- A map rather than a single slot because one developer can be configured against more than one
  workspace. A v1 file migrates into `published` (a v1 config's provenance is unrecoverable, and
  published is the safe reading), the legacy file is copied to `managed-state.json.pre-v2.bak` once
  before the first v2 write, and every write goes through a temp file plus rename.
- README is rewritten around the new flow, including the on-disk state table.
- Reviewer focus: admin detection and its fallback in `src/ucode/cli.py` (an inconclusive check must
  not strand a non-admin in the authoring path), the phase ordering in `src/ucode/managed_wizard.py`,
  and the slot read precedence, which is draft-then-published on every authoring surface and
  published alone on the apply path.

## How do you know it works?

`tests/test_cli.py` and `tests/test_managed_wizard.py` were rebuilt around `configure`, covering the
admin and non-admin branches, the dry run, authoring from a file, and publish;
`tests/test_managed_config.py` covers the v1 to v2 migration and the slot boundary. #440 then drives
the whole flow end to end against a fake workspace. Full unit suite green locally at this layer
(2165 passed, 38 skipped).

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,52 +217,58 @@ ucode skills --location main.default,ml.prod --mcp
 Each run prints the registered server, its URL, the configured agents, and its tools, and reminds
 you to run `ucode <agent>` (existing agent sessions need a restart before the MCP tools load).
 
-### Managed config for a workspace (admins)
+### Managed config for a workspace
 
-Author the coding config your developers pick up automatically, instead of asking each of them to
-run `ucode configure` by hand. Restricted to workspace admins. `ucode setup help` prints the whole
-sequence; the short version is one command for the agents and models, then the spend-tier command,
-then publish:
+`ucode configure` is the single entry point for both configuring your own machine and (for admins)
+authoring the config your developers pick up automatically. It figures out which you need:
 
-```bash
-ucode setup                 # agents and models (start here)
-ucode setup spend-tiers     # spend-based routing
-ucode publish               # publish it to the workspace
-```
+1. It selects and authenticates your workspace and fetches the workspace's managed config once.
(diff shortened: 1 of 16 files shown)


david-siqi-liu:
Superseded by the #500/#504/#510 stack, which reverses the unified role-aware configure this PR builds: #504 removes the admin auto-routing and makes `ug configure` application-only, and #510 removes `ug setup`/`ug publish` entirely (authoring moves to the AI Gateway API and the Unity Gateway UI). Closing as obsolete.

PR #435, opened 2026-09-01
MPS: resolve relayed --model against a curated target allowlist

masonc08:
## Problem

adding enforcement for an allowed models list, if it is specified on the model provider service.

for example, if the MPS only allows for haiku, the client should launch only haiku by default, and not allow any other models to be launched

## Testing
In the tests below, we launch haiku through a parameter, or when nothing is passed in. if we try to specify opus to be launched,it gets rejected

Screenshot 2026-09-09 at 16 33 01
Screenshot 2026-09-09 at 16 32 48
Screenshot 2026-09-09 at 16 32 34

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -2109,15 +2109,24 @@ def _launch_tool(
         # The router's per-launch pick for the root session. Codex pins it as the
         # resolved model; claude pins it via ANTHROPIC_MODEL (route_root_model).
         route_root_model = None
+        relayed_forward_model = None  # forwarded to Claude Code's --model for a relayed provider
         if provider:
             # Routing through a Model Provider Service pins no Databricks model;
             # the agent uses its own canonical model names (header selects the
             # provider). Skip model resolution, which would otherwise fail when
             # the workspace has no matching Databricks models.
             resolved_model = None
-            # Relayed services forward --model to Claude Code's own flag at launch (below), not env.
-            if tool == "claude" and not relayed and (model or provider_models):
-                route_root_model = resolve_provider_launch_model(model, provider_models or {})
+            if tool == "claude" and (model or provider_models):
+                if relayed:
+                    # Resolve against a curated allowlist so the forwarded id is one the gateway
+                    # allows; an allow_all relay declares none, so forward as-is.
+                    relayed_forward_model = (
+                        resolve_provider_launch_model(model, provider_models, always_select=True)
+                        if provider_models
+                        else model
+                    )
+                else:
+                    route_root_model = resolve_provider_launch_model(model, provider_models or {})
             if tool == "gemini":
                 # Gemini is the exception: the request still names a concrete model
                 # in the URL, so pin one of the service's targets (--model or default).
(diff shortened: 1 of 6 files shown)


src/ucode/cli.py
```diff
@@ -2221,15 +2221,23 @@ def _launch_tool(
...
             # provider). Skip model resolution, which would otherwise fail when
             # the workspace has no matching Databricks models.
             resolved_model = None
-            # Relayed services forward --model to Claude Code's own flag at launch (below), not env.
             if tool == "claude" and not relayed and (model or provider_models):
                 route_root_model = resolve_provider_launch_model(model, provider_models or {})
+            elif tool == "claude" and relayed and (model or provider_models):
+                # Resolve against a curated allowlist so the forwarded id is one the gateway allows;
+                # an allow_all relay declares none, so forward as-is.
+                relayed_forward_model = (
```
lilly-luo: could you nest this for easier reading? 

```
if tool == 'claude' and (model or provider_moels): 
     if relayed:
         ...your new logic 
      else:
          route_root_model = resolve_provider_launch_model(model, provider_models or {})
```

src/ucode/cli.py
```diff
@@ -2270,10 +2278,16 @@ def _launch_tool(
...
-            ctx.args = ["--model", model, *ctx.args]
-            forwarded_model = model
+        # Relayed = a Claude subscription: forward the model to Claude Code's own flag, like `-- --model X`.
+        if (
+            tool == "claude"
+            and provider
+            and relayed
+            and relayed_forward_model
+            and not forwarded_model
+        ):
```
lilly-luo: can you combine this set of logic checks into a meaningful variable name that encapsulates what it's trying to do? 

```
should_set_model_override_to_relayed_forward_model = (
            tool == "claude"
            and provider
            and relayed
            and relayed_forward_model
            and not forwarded_model
        )
```

src/ucode/databricks.py
```diff
@@ -2472,7 +2475,7 @@ def resolve_provider_launch_model(model: str | None, provider_models: dict[str,
                 )
             return target
         return model
-    if provider_models.get("opus"):
+    if not always_select and provider_models.get("opus"):
```
lilly-luo: what does always_select mean

lilly-luo (commented):
can you test e2e with ucode and provide screenshot in pr desc?

src/ucode/databricks.py
masonc08: `always_select` will decide which model to launch from the ucode layer. we will decide, then pass it onto claude code.

if its false, then we tell claude to make the decision on what tolaunch

masonc08:
> can you test e2e with ucode and provide screenshot in pr desc?

yes, done @lilly-luo

src/ucode/databricks.py
```diff
@@ -2500,7 +2509,7 @@ def resolve_provider_launch_model(model: str | None, provider_models: dict[str,
...
```
lilly-luo: why is there special casing for opus and not haiku or sonnet?

lilly-luo: can you rename `always_select` to `use_provider_allowlist` or something similar instead?

masonc08: i removed this logic. the previous logic was to run the claude default model (opus) if it was available in the allowed model list in the MPS. but this logic is not clean because we are hardcoding opus, and assuming that the default claude model will always be that.

the new approach that i've implemented is to just default to the following model families, in preference:
1. sonnet
2. opus
3. haiku

eg. if only opus is available in the allowed model list, and ug claude is launched without a model parameter, we will launch opus by default

lilly-luo (approved):

src/ucode/databricks.py
lilly-luo: thanks

PR #437, opened 2026-09-01
Split managed state into draft and published slots

david-siqi-liu:

- #438
  - #433
    - #440

---

`~/.ucode/managed-state.json` kept one config per workspace, so an admin authoring a config and a
developer's launch-time snapshot wrote to the same key and overwrote each other. This gives every
workspace two named slots: `draft` (what an admin is authoring, never applied to any agent) and
`published` (the last snapshot fetched from the workspace, the one that gets applied on launch).

- New on-disk shape is a versioned map: `version: 2` plus a `workspaces` object, each workspace
  holding a `draft` and a `published` slot.
- v1 files migrate on first read, and the pre-migration bytes are preserved next to the file as
  `managed-state.json.pre-v2.bak`.
- `save_managed_state` and `load_managed_state` remain as thin aliases over the published slot so
  the existing call sites keep working unchanged. #433 deletes them once every caller has moved.
  Keeping them here holds this PR to one file of real logic instead of a mechanical rename across
  six more.
- The map is written through a temp file and renamed into place. Before v2 a torn write only cost a
  published snapshot that the next launch refetches; now the same file holds a draft that nothing can
  rebuild.
- Reviewer focus: `_migrated_workspaces` and `_save_slot` in `src/ucode/managed_config.py`. Everything
  else follows from those two.

Behavior preserved: an empty published config still records "this workspace has no managed config",
which is what stops a server-side removal from being reapplied from stale disk state after a
transient fetch failure.

## How do you know it works?

`tests/test_managed_config.py` covers the v2 round trip, the v1 migration and its backup file, an
unreadable or malformed file, slot isolation (writing one slot leaves the other slot and every other
workspace untouched), and a write that fails part way leaving the previous state intact. Full unit suite green locally at this layer (2179 passed, 37 skipped), and `test` plus `e2e` are green on CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,7 @@ control the installation.
 | `~/.pi/agent/models.json` | Pi |
 | `~/.cursor/mcp.json` | Cursor Agent (MCP servers only) |
 | `~/.ucode/managed-state.json` | The managed config — authored by `ucode setup` (admins) and refreshed from the workspace on launch |
+| `~/.ucode/managed-state.json.pre-v2.bak` | One-time copy of a pre-slots `managed-state.json`, kept when it is first migrated |
 | `~/.ucode/managed-backups/` | Baseline backups for OS-managed files changed by ucode |
 
 Existing files are backed up before being overwritten. `ucode revert` restores backups.
--- a/src/ucode/managed_config.py
+++ b/src/ucode/managed_config.py
@@ -36,9 +36,11 @@ from ucode.databricks import (
     fetch_model_recommendation,
     get_databricks_token,
 )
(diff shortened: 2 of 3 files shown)


david-siqi-liu:
Folded into #433. The draft/published split only has a consumer once authoring writes the draft, and keeping it separate meant this PR shipped `save_managed_state` as a deprecated alias that #433 then deleted. Squashed, that alias never exists and `save_draft_config` is introduced alongside its first caller. The stack is now #438 -> #433 -> #440, with the same cumulative diff (verified: identical trees).

PR #440, opened 2026-09-01
Add a sandbox that exercises every `ucode configure` path end to end

david-siqi-liu:

- #438
  - #433
    - **#440** ⬅ this PR

---

The configure flow spans admin detection, authoring, publish, apply, and per-agent file writes, and
the unit tests mock most of that away. This adds a sandbox that runs the real CLI against a fake
Databricks workspace inside a throwaway HOME, so the file writes, the state transitions, and the
printed output under assertion are the real ones.

- 27 scenarios across 6 groups (`draft`, `published`, `empty`, `gate`, `multi`, `section`). Each runs
  in its own subprocess with a fresh HOME, fresh module state, and fresh caches.
- Only two seams are faked: the HTTP layer in `ucode.databricks` and the questionary prompts in
  `ucode.ui`. Everything above them (cli, managed_wizard, managed_config, managed_publish, agents,
  config_io, state) executes for real.
- Any URL the fake workspace does not recognize is recorded as unrouted and returned as an error, so
  a flow that reaches an unmodelled endpoint fails loudly instead of degrading quietly. That is
  already earning its keep: the Unity AI Gateway preflight added by #453 hit the sandbox as 23 loud
  failures until `/api/2.1/unity-catalog/model-services` was modelled here.
- Tests only. No change under `src/`.
- Reviewer focus: the expectations table in `tests/test_configure_sandbox.py`. That is the actual
  specification of what each path should do; the rest is plumbing.

## How do you know it works?

`uv run pytest tests/test_configure_sandbox.py` collects 28 tests: the 27 scenarios plus a guard
that fails if a scenario is added without expectations. Full unit suite green locally at this layer
(2193 passed, 38 skipped), and the sandbox runs as part of the default suite in CI. To drive one
scenario by hand: `python -m tests.sandbox_scenarios <name>`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,7 @@ Tests live in `tests/`.
 - Run the full test suite with `uv run pytest`.
 - Run focused tests with `uv run pytest tests/<file>.py`.
 - Run e2e tests with `UCODE_TEST_WORKSPACE=<db_workspace_url> uv run pytest tests/test_e2e.py -v`.
+- Run the `ucode configure` sandbox with `uv run pytest tests/test_configure_sandbox.py` (each scenario is a `test_configure_scenario[<name>]` case). To drive one scenario by hand, run `python -m tests.sandbox_scenarios <name>` from the repo root.
 - Run lint with `uv run ruff check .`.
 - Run the CLI from the current checkout with `uv run ucode ...`.
 - Reinstall the local checkout as the `ucode` tool with `uv tool install --reinstall .`.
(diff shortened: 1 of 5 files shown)


david-siqi-liu:
Superseded by the #500/#504/#510 stack. This sandbox exercises the authoring/publish/admin-detection configure flow that #510 removes, so it no longer maps to the developer-only `ug configure`. Closing as obsolete; a sandbox for the new developer-only flow would be a fresh PR.

PR #452, opened 2026-09-02
Handle MCP transport timeouts and disconnects

anthonyivn2:
## Summary

- configure the MCP HTTP client with timeout values matching the MCP Python SDK’s recommended client defaults: 30 seconds for connect/write/pool and 300 seconds for reads
- raise exceptions yielded by MCP transport streams instead of forwarding them as messages
- treat unexpected upstream EOF as a transport failure while preserving clean shutdown when the client closes stdin

### Align the custom HTTP client with MCP SDK timeouts

**Problem:** ucode supplies its own HTTPX client to attach Databricks per-request authentication. This bypasses the SDK client factory and its timeout configuration, leaving HTTPX’s 5-second read-inactivity default in effect. A slow GitHub MCP response can therefore disconnect while the operation is still valid.

The MCP protocol does not mandate timeout values. The 30-second connect/write/pool timeout and 300-second read timeout come from the MCP Python SDK implementation in [v2.1.1](https://github.com/modelcontextprotocol/python-sdk/blob/v2.1.1/src/mcp/shared/_httpx_utils.py#L9-L11).

**Fix:** explicitly configure these values on ucode’s authenticated client so it retains the SDK-recommended timeout behavior.

**Code:** `src/ucode/mcp_proxy.py` — `_run()`

### Propagate errors yielded by MCP streams

**Problem:** MCP read streams can yield an `Exception` instead of a protocol message. The proxy previously forwarded every yielded value to stdio, so it did not recognize this value as a transport failure.

**Fix:** detect and raise yielded exceptions. The bridge terminates with the original error, allowing the client to observe that the MCP connection failed.

**Code:** `src/ucode/mcp_proxy.py` — `_pump()`

### Fail when the upstream stream closes unexpectedly

**Problem:** some transport failures surface to the proxy only as upstream EOF. Both proxy directions previously treated EOF as normal completion. The upstream pump could therefore finish successfully while the client-input pump continued waiting on stdin, leaving the proxy alive but unable to receive tool responses. Claude would continue showing the tool call as running.

**Fix:** treat completion of the upstream pump as an error. The error cancels the client-input pump and terminates the proxy. The client-input pump remains the normal control path, so Claude closing stdin still cancels the upstream task and shuts down cleanly.

**Code:** `src/ucode/mcp_proxy.py` — `_pump_upstream()` and `_run()`

## Testing

- `uv run --frozen pytest tests/test_mcp_proxy.py -q` — 27 passed
- `uv run --frozen ruff check .`
- `git diff --check`

### DIFF
--- a/src/ucode/mcp_proxy.py
+++ b/src/ucode/mcp_proxy.py
@@ -117,13 +117,24 @@ async def _pump(
     source: MemoryObjectReceiveStream,
     dest: MemoryObjectSendStream,
 ) -> None:
-    """Forward every message (or transport exception) from ``source`` to ``dest``.
+    """Forward every message from ``source`` to ``dest``.
 
     The proxy is transport-level: it never inspects or rewrites MCP method
     payloads, so new methods and capabilities pass through untouched."""
     async with source, dest:
         async for message in source:
+            if isinstance(message, Exception):
+                raise message
             await dest.send(message)
+
+
+async def _pump_upstream(
+    source: MemoryObjectReceiveStream,
+    dest: MemoryObjectSendStream,
+) -> None:
+    """Forward upstream messages, failing if the transport closes first."""
+    await _pump(source, dest)
+    raise RuntimeError("upstream MCP transport closed unexpectedly")
 
 
 async def _run(url: str, workspace: str, profile: str | None) -> None:
@@ -132,7 +143,15 @@ async def _run(url: str, workspace: str, profile: str | None) -> None:
     # 2.x-native shape: hand the transport a pre-built AsyncClient carrying our
     # per-request auth. Works on mcp 1.28+ and 2.x; `streamable_http_client`
     # yields a (read, write) pair in both.
-    async with httpx.AsyncClient(auth=auth) as http_client:
+    async with httpx.AsyncClient(
+        auth=auth,
+        timeout=httpx.Timeout(
+            connect=30.0,
+            read=300.0,
+            write=30.0,
+            pool=30.0,
+        ),
+    ) as http_client:
         async with streamable_http_client(url, http_client=http_client) as streams:
             # mcp 1.x yields (read, write, get_session_id); mcp 2.x drops the
             # trailing callback and yields (read, write). Take the first two
@@ -141,8 +160,9 @@ async def _run(url: str, workspace: str, profile: str | None) -> None:
             async with stdio_server() as (stdio_read, stdio_write):
                 # Bidirectional bridge: client stdin -> Databricks, Databricks -> client stdout.
                 async with anyio.create_task_group() as tg:
-                    tg.start_soon(_pump, stdio_read, http_write)
-                    tg.start_soon(_pump, http_read, stdio_write)
+                    tg.start_soon(_pump_upstream, http_read, stdio_write)
+                    await _pump(stdio_read, http_write)
+                    tg.cancel_scope.cancel()
 
 
 def _preflight_token(workspace: str, profile: str | None) -> None:
--- a/src/ucode/mcp_proxy.py
+++ b/src/ucode/mcp_proxy.py
@@ -43,9 +43,8 @@ from mcp.server.stdio import stdio_server
 
 from ucode.databricks import ensure_pat_bearer, get_databricks_token
 
-# Exit code used when the proxy cannot authenticate. MCP clients surface a
-# non-zero exit far more usefully than a startup timeout, so bail out with this
-# instead of letting the process hang until the client's timeout fires.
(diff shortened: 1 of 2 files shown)


src/ucode/mcp_proxy.py
```diff
@@ -117,22 +117,41 @@ async def _pump(
...
 ) -> None:
-    """Forward every message (or transport exception) from ``source`` to ``dest``.
+    """Forward every message from ``source`` to ``dest``.
 
     The proxy is transport-level: it never inspects or rewrites MCP method
     payloads, so new methods and capabilities pass through untouched."""
     async with source, dest:
         async for message in source:
+            if isinstance(message, Exception):
+                raise message
```
sunishsheth2009: `_pump` is used for both directions, and `stdio_server`'s read stream sends an `Exception` (a `ValidationError`) into the stream when a client line fails JSON-RPC decode. So on the **client -> upstream** path this now tears down the entire proxy on a single malformed client message, whereas before it was forwarded (and would have failed downstream differently).

For the **upstream** direction this raise is exactly what we want. For the client-input direction it's more aggressive than before — probably fine in practice since Claude sends well-formed JSON-RPC, but worth a conscious decision. If you want to keep the stdin path tolerant, the raise could live in `_pump_upstream` instead of the shared `_pump`.

src/ucode/mcp_proxy.py
```diff
@@ -117,22 +117,41 @@ async def _pump(
...
             await dest.send(message)
 
 
+async def _pump_upstream(
+    source: MemoryObjectReceiveStream,
+    dest: MemoryObjectSendStream,
+) -> None:
+    """Forward upstream messages, failing if the transport closes first."""
+    await _pump(source, dest)
+    raise RuntimeError("upstream MCP transport closed unexpectedly")
```
sunishsheth2009: This `RuntimeError` propagates out of the task group as an `ExceptionGroup`. In `serve`, `_unwrap_auth_error` returns `None` for it, so it hits the bare `raise` and the user sees a raw Python traceback + nonzero exit. That's strictly better than the old silent hang, but inconsistent with the clean stderr UX of the auth path (`_fail_fast`). Consider surfacing a one-line stderr message for the "upstream transport closed" case too.

Minor: there's also a benign race — if the client closes stdin (clean shutdown) and upstream EOFs at nearly the same moment, `_pump_upstream` could win and turn a clean shutdown into an error exit. Low impact, just noting it.

src/ucode/mcp_proxy.py
```diff
@@ -117,22 +117,41 @@ async def _pump(
...
     auth = _build_token_auth(workspace, profile)
     # 2.x-native shape: hand the transport a pre-built AsyncClient carrying our
     # per-request auth. Works on mcp 1.28+ and 2.x; `streamable_http_client`
     # yields a (read, write) pair in both.
-    async with httpx.AsyncClient(auth=auth) as http_client:
+    async with httpx.AsyncClient(
+        auth=auth,
+        timeout=httpx.Timeout(
+            connect=30.0,
+            read=300.0,
```
sunishsheth2009, line 156: Confirming intent: httpx `read` here is the between-bytes inactivity timeout, not a total-operation cap, so a long-lived idle server->client stream that's silent for >300s would trip it. That matches the MCP SDK's own default, so aligning is correct — just flagging so it's a deliberate choice. Also worth a quick check that `httpx2.Timeout` (mcp 2.x path) accepts the same `connect/read/write/pool` kwargs (it should, as a fork), since `_httpx()` can return either flavor.

src/ucode/mcp_proxy.py
anthonyivn2: Good point. I moved the exception check into `_pump_upstream()`. Client input keeps the previous forwarding behavior, while upstream exceptions now stop the proxy. I added tests for both paths.

src/ucode/mcp_proxy.py
anthonyivn2: Confirmed. `read` is an inactivity timeout. I checked MCP 2.1.1 with httpx2 2.12.0; `connect`, `read`, `write`, and `pool` accept these values and resolve to 30/300/30/30.

src/ucode/mcp_proxy.py
anthonyivn2: Updated. Upstream errors and unexpected EOF now become `ProxyTransportError`. `serve()` prints one line to stderr and exits, so users do not get a raw traceback. I left the rare simultaneous shutdown race unchanged because both sides are already closing.

sunishsheth2009 (approved):

sunishsheth2009:
Approved the PR, can we confirm that this PR will resolve this issue?

PR #453, opened 2026-09-02
Improve AI Gateway diagnostics during configure

david-siqi-liu:

**Change:** `ucode configure` now reports whether it can access a Unity Catalog model service. If no accessible model service is returned, it probes legacy endpoints internally to determine whether Unity AI Gateway is still usable. The user-facing success output remains focused on the model service result.

**Why:** Today, `ucode configure` provides little telemetry about the discovery path it can access. Users can eventually see a CSP-related fallback error even when the actionable problem is missing access to `system.ai`. Reporting the model-service result makes that failure clear, while avoiding an unnecessary legacy probe when model-service discovery already succeeds.

Related: ES-2185388

### An accessible model service is returned

```text
✔ Unity AI Gateway detected
  Model service: reachable, accessible model service returned
```

The legacy endpoint API is not probed.

### Model-service discovery is reachable but returns nothing accessible, and the legacy endpoint API is reachable

```text
✔ Unity AI Gateway detected
  Model service: reachable, no accessible model services returned; check USE CATALOG on system, and USE SCHEMA and EXECUTE on system.ai
```

This applies whether the legacy endpoint listing is empty or contains an accessible endpoint. The fallback is checked internally and is not printed.

### Model-service discovery is unavailable or forbidden, and the legacy endpoint API is reachable

```text
✔ Unity AI Gateway detected
  Model service: HTTP 403 Forbidden: ...
```

The exact model-service reason is printed after replacing any API-version terminology with resource-oriented language. The reachable legacy endpoint fallback is not printed.

### Model-service discovery is reachable but empty, and the legacy endpoint API is unavailable

No success summary is printed. Configuration raises an actionable model-service access error with this guidance:

```text
Verify USE CATALOG on `system`, and USE SCHEMA and EXECUTE on `system.ai`.
```

### Model-service discovery is unavailable, and legacy endpoint access is forbidden

No success summary is printed. Configuration raises an actionable legacy endpoint permission error.

### Neither discovery path is reachable

No success summary is printed. Configuration reports that Unity AI Gateway is not enabled and includes both resource-oriented failure reasons.

### The access token is invalid

No success summary is printed. Configuration stops immediately with `databricks auth logout` and `databricks auth login` guidance. If model-service discovery identifies the invalid token, the legacy endpoint API is not probed.

A successful empty model-service response, including `HTTP 200 {}`, confirms that the API is reachable. It does not prove that the caller can access a model service.

## How do you know it works?

- Added deterministic local scenarios that drive the real probe formatter through `configure_shared_state`.
- Verifies that an accessible model service skips the legacy endpoint request.
- Verifies that legacy endpoint results are used only to validate the fallback and are not printed.
- Covers empty and forbidden model-service results with reachable, empty, forbidden, and unavailable legacy endpoint discovery.
- Covers neither discovery path being reachable and invalid access tokens.
- Verifies that user-facing errors do not expose API-version terminology.
- Added focused probe coverage for non-empty and empty responses, HTTP failures, and the `system.ai` permission hint.
- Isolated publish CLI wiring tests from the real Databricks CLI installer so CI does not depend on network bootstrap.
- `uv run ruff check .` and `uv run ruff format --check src/ tests/` passed.
- The suite passed with `2176 passed, 37 skipped, 1 deselected`. The deselected live Claude user-agent E2E is blocked in this development environment because system-managed Claude settings override its localhost capture server.

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -3049,11 +3049,12 @@ def fetch_codex_models(workspace: str, token: str) -> list[str]:
 class GatewayProbe(NamedTuple):
     reachable: bool
     detail: str
+    resource_available: bool = False
 
 
 class GatewayCapabilities(NamedTuple):
     v3: GatewayProbe
-    v2: GatewayProbe
+    v2: GatewayProbe | None
 
 
 def _gateway_probe_result(
@@ -3067,7 +3068,7 @@ def _gateway_probe_result(
         return GatewayProbe(False, reason or "unknown error")
     resources = payload.get(collection_key) if isinstance(payload, dict) else None
     if resources:
-        return GatewayProbe(True, f"reachable, accessible {resource_name} returned")
+        return GatewayProbe(True, f"reachable, accessible {resource_name} returned", True)
     detail = f"reachable, no accessible {resource_name}s returned"
     if empty_hint:
         detail = f"{detail}; {empty_hint}"
@@ -3125,10 +3126,12 @@ def _raise_ai_gateway_v2_permission_failure(
 
 
 def ensure_ai_gateway(workspace: str, token: str) -> GatewayCapabilities:
-    """Return both gateway probe results if either V2 or V3 is reachable."""
+    """Return gateway probe results, checking V2 only when V3 has no accessible resource."""
     v3 = _probe_ai_gateway_v3(workspace, token)
     if not v3.reachable and _looks_like_definitive_auth_failure(v3.detail):
         _raise_ai_gateway_auth_failure(workspace, v3.detail)
+    if v3.resource_available:
+        return GatewayCapabilities(v3=v3, v2=None)
 
     v2 = _probe_ai_gateway_v2(workspace, token)
     capabilities = GatewayCapabilities(v3=v3, v2=v2)
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -3126,7 +3126,7 @@ def _raise_ai_gateway_v2_permission_failure(
 
 
 def ensure_ai_gateway(workspace: str, token: str) -> GatewayCapabilities:
-    """Return gateway probe results, checking V2 only when V3 has no accessible resource."""
+    """Probe model services first, falling back to legacy endpoints when needed."""
     v3 = _probe_ai_gateway_v3(workspace, token)
     if not v3.reachable and _looks_like_definitive_auth_failure(v3.detail):
         _raise_ai_gateway_auth_failure(workspace, v3.detail)
@@ -3134,9 +3134,8 @@ def ensure_ai_gateway(workspace: str, token: str) -> GatewayCapabilities:
         return GatewayCapabilities(v3=v3, v2=None)
 
     v2 = _probe_ai_gateway_v2(workspace, token)
-    capabilities = GatewayCapabilities(v3=v3, v2=v2)
     if v3.reachable or v2.reachable:
-        return capabilities
+        return GatewayCapabilities(v3=v3, v2=v2)
     if _looks_like_definitive_auth_failure(v2.detail):
         _raise_ai_gateway_auth_failure(workspace, v2.detail)
     if _looks_like_permission_failure(v3.detail):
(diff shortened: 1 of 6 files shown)


src/ucode/databricks.py
```diff
@@ -3046,18 +3046,53 @@ def fetch_codex_models(workspace: str, token: str) -> list[str]:
...
     hostname = workspace_hostname(workspace)
     url = f"https://{hostname}/api/2.1/unity-catalog/model-services?page_size=1"
     payload, reason = _http_get_json(url, token)
-    return payload is not None, reason
+    return _gateway_probe_result(
+        payload,
+        reason,
+        "model_services",
+        "model service",
+        "check USE CATALOG on system, and USE SCHEMA and EXECUTE on system.ai",
```
lilly-luo: nit add kwargs so easier to read 

```
_gateway_probe_result(
        payload=payload,
        reason=reason,
        collection_key="model_services",
        resource_name="model service",
        empty_hint="check USE CATALOG on system, and USE SCHEMA and EXECUTE on system.ai",
```

src/ucode/databricks.py
```diff
@@ -3090,27 +3125,28 @@ def _raise_ai_gateway_v2_permission_failure(
...
-    v2_ok, v2_reason = _probe_ai_gateway_v2(workspace, token)
-    if v2_ok:
-        return
-    if v2_reason and _looks_like_definitive_auth_failure(v2_reason):
-        _raise_ai_gateway_auth_failure(workspace, v2_reason)
-    if v3_reason and _looks_like_permission_failure(v3_reason):
-        _raise_ai_gateway_v3_permission_failure(workspace, v3_reason, v2_reason)
-    if v2_reason and _looks_like_permission_failure(v2_reason):
-        _raise_ai_gateway_v2_permission_failure(workspace, v2_reason, v3_reason)
+def ensure_ai_gateway(workspace: str, token: str) -> GatewayCapabilities:
```
lilly-luo: nit rename ensure_ai_gateway -> probe_unity_gateway_capabilities. ensure sounds like it's gonna throw if it fails vs actually return smth

src/ucode/databricks.py
```diff
@@ -3090,27 +3125,28 @@ def _raise_ai_gateway_v2_permission_failure(
...
+    v3 = _probe_ai_gateway_v3(workspace, token)
+    if not v3.reachable and _looks_like_definitive_auth_failure(v3.detail):
+        _raise_ai_gateway_auth_failure(workspace, v3.detail)
+    if v3.resource_available:
+        return GatewayCapabilities(v3=v3, v2=None)
+
+    v2 = _probe_ai_gateway_v2(workspace, token)
+    capabilities = GatewayCapabilities(v3=v3, v2=v2)
+    if v3.reachable or v2.reachable:
+        return capabilities
```
lilly-luo: this has to be v2 capabilities right? given v3 resource isnt avail

src/ucode/databricks.py
```diff
@@ -3090,27 +3125,28 @@ def _raise_ai_gateway_v2_permission_failure(
...
+    if _looks_like_definitive_auth_failure(v2.detail):
+        _raise_ai_gateway_auth_failure(workspace, v2.detail)
+    if _looks_like_permission_failure(v3.detail):
+        _raise_ai_gateway_v3_permission_failure(workspace, v3.detail, v2.detail)
+    if _looks_like_permission_failure(v2.detail):
+        _raise_ai_gateway_v2_permission_failure(workspace, v2.detail, v3.detail)
 
     raise RuntimeError(
         "Databricks AI Gateway is not enabled on this workspace: neither V3 "
```
lilly-luo: can we not expose v2/v3 to the customer? given we're trying to get ppl to migrate to v3, can we just show the v3 detail and leave as is without including anything about v2 and v3

src/ucode/databricks.py
david-siqi-liu: ya at this point `v3.reachable` only means it's reachable, BUT returned an empty listing. let me change the condition to require `v2.reachable`

lilly-luo (approved):

Issue #454, opened 2026-09-02
feat(codex): support amazon_bedrock Model Provider Services that expose OpenAI-compatible models


bjanssen11:
## Problem

`ucode codex --provider <bedrock-mps>` fails with:

```
ERROR Model provider service '<name>' is a 'amazon_bedrock' provider, which codex can't route to (supported: openai).
```

This happens even when the Bedrock MPS exposes OpenAI-compatible models, because `_TOOL_PROVIDER_TYPES` hardcodes codex to only `("openai",)`, and a secondary check in `resolve_provider_service` rejects any Bedrock service that doesn't declare Claude model targets.

## Root cause

Two places in `src/ucode/databricks.py` prevent this from working:

1. **`_TOOL_PROVIDER_TYPES`** (`~line 2097`) — `"codex": ("openai",)` never matches `amazon_bedrock`.
2. **Bedrock-Claude-targets guard** in `resolve_provider_service` and `service_usable_for_tool` — rejects any Bedrock MPS without Claude targets, regardless of which tool is asking.

Codex uses the OpenAI-compatible endpoint. Bedrock exposes this same endpoint for non-Claude models (e.g. Titan, GPT-class models hosted on Bedrock). There's no reason to require Claude targets when codex is the active tool.

## Fix

- Add `"amazon_bedrock"` to codex's allowed types in `_TOOL_PROVIDER_TYPES`.
- Gate the "exposes no Claude models" check on `tool == "claude"` in both `resolve_provider_service` and `service_usable_for_tool`.

PR forthcoming.

PR #455, opened 2026-09-02
feat(codex): support amazon_bedrock Model Provider Services that expose OpenAI-compatible models

bjanssen11:
Related to #454

## Summary

**Fix: codex + Bedrock MPS validation**
- Add `"amazon_bedrock"` to codex's allowed provider types in `_TOOL_PROVIDER_TYPES` (was `("openai",)` only)
- Gate the "exposes no Claude models" check in `resolve_provider_service` on `tool == "claude"` so a Bedrock MPS with non-Claude targets isn't rejected when codex picks it
- Apply the same `tool == "claude"` guard in `service_usable_for_tool` so Bedrock services without Claude targets appear in the listing for codex

Codex uses the OpenAI-compatible endpoint. Bedrock exposes this same interface for non-Claude models. There's no reason to require Claude targets when codex is the active tool.

**Fix: Bedrock model picker shows OpenAI models**

When `ucode codex --provider <bedrock-mps>` launched, Codex's built-in model picker queried OpenAI and showed `gpt-5-codex`/`gpt-5` instead of the Bedrock targets declared on the MPS.

- At launch time, fetch the MPS targets and offer a picker (or auto-pin when there's only one)
- Honor explicit `--model` for codex in the provider path (was previously a no-op)
- Make `codex.write_tool_config` use the `model` parameter when a provider is active (was silently ignored)

**New: `ucode providers` commands**
- `ucode providers list [--tool TOOL]` — lists all MPS services on the workspace with name, provider type, and declared targets. `--tool claude|codex` filters to services the given tool can route through.
- `ucode providers show <catalog.schema.service>` — shows full detail for one MPS: provider type, relay flag, allow_all_targets, and the complete targets list.

## Test plan

- `test_codex_bedrock_openai_compat_ok` — codex + Bedrock MPS exposing only non-Claude targets now resolves successfully
- `test_codex_bedrock_with_claude_targets_ok` — codex + Bedrock MPS with Claude targets also works
- `test_codex_anthropic_rejected` — codex still can't route to an Anthropic MPS
- `test_codex_filters_to_openai_and_bedrock` — updated from `test_codex_filters_to_openai` to verify Bedrock services appear in codex's listing

```
uv run pytest tests/test_databricks.py tests/test_agent_codex.py tests/test_agents_init.py tests/test_lint.py  # 409 passed
uv run ruff check .   # no issues
```

https://claude.ai/code/session_017G9kjrbhqvWucH26GwykSn

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -350,6 +350,7 @@ def configure_tool(
     relayed: bool = False,
     route_root_model: str | None = None,
     custom_model: str | None = None,
+    bedrock_targets: list[str] | None = None,
 ) -> dict:
     result: dict | tuple[dict, str]
     if tool == "codex":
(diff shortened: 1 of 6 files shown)


bjanssen11:
Closing this. Codex cannot use Amazon Bedrock through the Databricks AI Gateway: the gateway serves the amazon_bedrock provider only over the Converse API, and Codex speaks the OpenAI Responses API, which the gateway refuses for Bedrock. Evidence and details in #476. Pi Bedrock support, which works over Converse, is tracked in its own PR.

PR #458, opened 2026-09-02
Introduce ug command and Phase 1 migration path

rohita5l:
## Summary
- add the `ug` console entry point while retaining `ucode` as a supported alias
- make the README and CLI commands/guidance lead with `ug`, while preserving Unity Gateway branding in agent launch banners and Smart Router notices
- make `ug upgrade` distribution-aware: upgrade `ucode` normally before cutover, migrate only on uv's confirmed metadata-name mismatch, then upgrade `unity-gateway` normally afterward
- preserve the existing installation on unrelated upgrade failures and verify both `ug` and `ucode` after every successful upgrade
- add focused coverage for both entry points, invocation-aware help, all three upgrade lifecycle stages, failure safety, renamed guidance, and launch branding

## Testing
- `.venv/bin/pytest tests/test_cli.py tests/test_entry_points.py -q` (284 passed)
- `.venv/bin/ruff check .`
- `.venv/bin/ruff format --check src/ucode/cli.py tests/test_cli.py`
- `.venv/bin/ty check src/ucode/cli.py`
- `git diff --check`
- isolated real-uv lifecycle probe: confirmed `ucode` metadata mismatch at cutover, successful reinstall as `unity-gateway`, and normal post-cutover `uv tool upgrade unity-gateway`

## Release requirement
Merge and tag this commit while `[project].name` is still `ucode`. That immutable tag is the bridge release and must exist before the distribution-name cutover.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -5,8 +5,10 @@ from __future__ import annotations
 
 import os
 import shutil
+import subprocess
 from collections.abc import Iterator
 from contextlib import contextmanager
+from importlib import metadata
 from typing import Annotated
 
 import typer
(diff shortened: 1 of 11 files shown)


src/ucode/cli.py
```diff
@@ -356,13 +358,13 @@ def _run_setup_and_exit(workspace: str, profile: str | None, token: str | None =
...
-        # Brand the flow as "Configure Unity Gateway": it was reached through `ucode configure`,
-        # not a bare `ucode setup`, so its section headers use the product name rather than the
+        # Brand the flow as "Configure ug": it was reached through `ug configure`,
+        # not a bare `ug setup`, so its section headers use the product name rather than the
         # bare command.
         code = setup_command(
             workspace=workspace,
             profile=profile,
-            command_label="Configure Unity Gateway",
+            command_label="Configure ug",
```
lilly-luo: leave as `Configure Unity Gateway` ?

rohita5l: Configure ug sounded better ? Configure Unity Gateway sounded weird ? thats the whole product

david-siqi-liu: Configure unity-gateway?

src/ucode/cli.py
```diff
@@ -3249,24 +3251,126 @@ def usage_cmd(
 
 @app.command("upgrade")
 def upgrade_cmd() -> None:
-    """Upgrade ucode to the latest version from GitHub."""
-    import subprocess
-
+    """Upgrade ug to the latest version from GitHub."""
+    legacy_distribution = "ucode"
+    current_distribution = "unity-gateway"
     git_url = "git+https://github.com/databricks/ucode"
```
david-siqi-liu, line 3404: after we rename the package, this will correctly re-direct right?

src/ucode/cli.py
```diff
@@ -3249,24 +3251,126 @@ def usage_cmd(
...
+        print_err("`uv` was not found on PATH. Install uv to upgrade ug.")
         raise typer.Exit(1) from None
     except subprocess.CalledProcessError as exc:
-        print_err(f"Upgrade failed (exit code {exc.returncode}).")
+        if legacy_removed:
+            print_err(
+                "The legacy `ucode` tool was removed, but installing `unity-gateway` failed "
+                f"(exit code {exc.returncode})."
+            )
+            print_note(f"Recover by running `uv tool install --force {git_url}`.")
```
david-siqi-liu, line 3458: should we not recommend `--force`?

src/ucode/cli.py
```diff
@@ -3249,24 +3251,126 @@ def usage_cmd(
...
+                migrated = True
+                installed_distribution = current_distribution
+            else:
+                detail = _upgrade_failure_detail(result)
+                print_err(
+                    f"Upgrade failed (exit code {result.returncode})"
+                    f"{f': {detail}' if detail else '.'}"
+                )
+                print_note("The existing installation was left unchanged.")
+                raise typer.Exit(1)
```
david-siqi-liu, line 3443: not 100% sure about the difference between this and
```
raise typer.Exit(1) from None
```
which seems to be used more widely below

src/ucode/cli.py
```diff
@@ -3249,24 +3251,126 @@ def usage_cmd(
...
+
+        _verify_upgraded_commands()
```
david-siqi-liu: nit: i think this should only be ran `if migrated` is true?

david-siqi-liu (approved):

src/ucode/cli.py
david-siqi-liu: hmm probably ok here, since this is a recovery command

src/ucode/cli.py
david-siqi-liu: agent telling me this is ok, but 

> typer.Exit subclasses RuntimeError, so this gets caught by the except RuntimeError below and adds a trailing ERROR 1 after the real error. Minor, but an except typer.Exit: raise before that handler would avoid it.

lilly-luo (approved):

PR #474, opened 2026-09-03
Restrict interactive MCP picker to MCP services for consumer access

sunishsheth2009:

**Change:** The interactive MCP picker (`ug configure mcp` / `ug mcp add`) no longer offers V2 AI Gateway sources — Vector Search, UC Functions, external connections, Genie spaces, and Databricks apps (everything under `/api/2.0/mcp/*`). It now offers only **MCP services** (the `/ai-gateway/mcp-services/` path) and skips the source-selection step since that's the sole source. Workspace users can still add a V2 server on request, non-interactively, by naming it in `--services` with a typed selector:

- `vector-search:<catalog>.<schema>`
- `uc-functions:<catalog>.<schema>`
- `external:<connection>`
- `genie-space:<space-id>`
- `app:<name>`

**Why:** Consumer-only identities lack the `workspace-access` entitlement and can't use the V2 AI Gateway, so offering those sources in the default flow leads to failures. MCP services is the consumer-safe path. This implements the team "Option 4" decision and the AIGTWY-4471 graceful-skip pattern.

**On consumer access (deliberately simple):** the `workspace-access` entitlement 403 is emitted only by *guarded AI Gateway / Model Serving inference and model-listing* paths, which `ug` already exercises at model setup. The listing calls this flow uses (`GET /api/2.0/apps`, UC catalogs, vector-search endpoints, `ai-gateway/v2/endpoints`) are **not** guarded — a consumer gets an empty list or a generic ACL denial, with no reliable, matchable signal. So this PR does **not** detect consumer-vs-missing-grant at MCP-add time (no sound signal; a consumer is already gated upstream at model setup). `PermissionDeniedError` is used only so discovery skips a 403 gracefully; an `app:` add that 403s gives an actionable "needs workspace access" error.

## Reviewer feedback addressed

- **Slow loading → progressive streaming.** The picker used to block on the workspace-wide MCP-services walk (up to a 30s budget) before showing anything. Now it opens **immediately** on the fast `system.ai` list (+ SQL + already-configured servers), and the walk runs on a daemon thread that **streams each schema's services in as they're found**, with a live `⏳ loading more MCP services… (N found)` footer. Nothing is hidden; no one waits for the full walk. Selection/scroll/filter are preserved as rows arrive (questionary reads its choice list live; append + thread-safe `app.invalidate()`).
- **Remove.** `ug mcp remove` is unchanged and still lists **every** configured server (including V2 ones added via `--services`) so you pick which to remove.

## How do you know it works?

**Testing:** `ruff check`, `ruff format --check`, and the `test_mcp.py` / `test_databricks.py` suites pass (full suite: only pre-existing E2E/codex-env failures remain). Added/updated tests cover: picker offering only MCP services and skipping the source prompt; the typed V2 `--services` selectors registering the right server URL; V2 selectors rejected when combined with `--location`/plain names; `list_databricks_apps` raising `PermissionDeniedError` on a 403; discovery skipping it quietly while warning on other errors; an `app:` add surfacing an actionable error; the workspace-wide walk streaming services into the picker via the background loader (`on_services` + `_mcp_services_background_loader`); and `_mcp_service_choice` building the right add/toggle/disabled row. The interactive repaint itself is verified manually.

## Notes

- Follow-up (separate PR): remove the now-dead V2 discovery functions in `mcp.py` and the underlying `list_*` helpers in `databricks.py`. It's interwoven — the VS/UC-Fn walk shares helpers with the retained mcp-services walk, and ~40 test monkeypatch stubs reference the `discover_*` names — so it needs its own diff + test refactor.


src/ucode/mcp.py
```diff
@@ -899,23 +1017,17 @@ def known_choice(name: str, title: str | None = None) -> questionary.Choice:
     choices: list[questionary.Choice | questionary.Separator] = []
     displayed_names: set[str] = set()
 
-    if "databricks-sql" in known_names:
-        choices.append(known_choice("databricks-sql", "Databricks SQL"))
-    else:
-        choices.append(_add_choice(SQL_MCP_VALUE, "Databricks SQL"))
-    displayed_names.add("databricks-sql")
+    # Databricks SQL is intentionally NOT offered as an up-front add-choice — we don't promote
```
xsh310, line 1069: **Possible blocker: the picker can crash when it has no selectable choices.**

Removing the always-present "Databricks SQL" entry means `build_mcp_picker_choices(...)` can now return `[]` (the new `test_picker_is_empty_when_nothing_discovered_or_configured` asserts exactly that). That empty list reaches `_scrolling_checkbox`, which builds `InquirerControl(choices)` synchronously, before the background loader thread starts.

In questionary 2.1.1, `InquirerControl.__init__` only sets `self.pointed_at` when it finds a non-disabled choice, then calls `is_selection_valid()`, which dereferences it. With an empty (or all-disabled) list it raises `AttributeError: 'InquirerControl' object has no attribute 'pointed_at'` (reproduced against questionary==2.1.1).

Why it matters here: this is the exact path the PR targets. On a fresh `ug configure mcp` where the fast `system.ai` list comes back empty (a consumer-only identity whose listing 403s, or any workspace with no system.ai MCP services), both `discovered["services"]` and the configured servers are empty, so the picker opens with zero choices and crashes. The background walk cannot rescue it, since the crash is at picker construction. Same crash for `ug mcp add` when every offered service is already configured (all rows become `disabled="already configured"`).

Not covered by tests: the picker is stubbed everywhere, and the new empty-picker test only checks that the builder returns `[]`, never renders it.

Suggested fix: make `_scrolling_checkbox` tolerate an empty or all-disabled list (guard `pointed_at` before `ask()`, or short-circuit with a note when nothing is selectable). A guard inside `_scrolling_checkbox` is preferable to a caller-side early return, since the streaming design intentionally opens on an empty list and fills it in.

README.md
```diff
@@ -113,14 +113,25 @@ ug configure mcp
...
 
 Add Databricks MCP servers to installed MCP-capable tools: Codex, Claude Code, Gemini CLI, OpenCode, GitHub Copilot CLI, and Cursor Agent.
-Options are shown in this order:
 
-- Discovered external MCP connections
-- Databricks SQL
-- Managed Databricks MCPs (Vector Search, UC Functions, etc.)
-- Custom MCP server URL
+The interactive picker discovers **MCP services** (the `system.ai.*` and workspace-wide
+`<catalog>.<schema>` Unity Catalog MCP services), plus Databricks SQL and a custom MCP server URL.
```
xsh310: README still lists "Databricks SQL" as a picker option, but `build_mcp_picker_choices` no longer offers it up front (see the "intentionally NOT offered" comment); drop it here.

src/ucode/mcp.py
```diff
@@ -1890,13 +2129,30 @@ def configure_mcp_command(
 
     # Two-step wizard: (1) choose which sources to search, (2) pick servers from
     # the results. Pressing Left (←) in the picker returns to step 1, so the user
-    # can revise their source selection without restarting the command.
+    # can revise their source selection without restarting the command. When only
+    # one search source is available (MCP services — the V2 sources were removed),
+    # step 1 has nothing to choose, so skip it and go straight to the picker.
+    available_source_keys = [k for k, _, _ in MCP_SEARCH_SOURCES if k not in excluded_sources]
+    prompt_sources = len(available_source_keys) > 1
```
xsh310: It seems like prompt_sources is always true after this change? Do we still need the prompt_for_mcp_search_sources call or is it safe to clean up?

src/ucode/mcp.py
```diff
@@ -1366,11 +1489,35 @@ def message() -> str:
...
+    def loader(append: Callable[[list[questionary.Choice]], None]) -> None:
+        def on_services(new_names: list[str]) -> None:
+            append([_mcp_service_choice(name, known_names, additive) for name in new_names])
+
+        discover_all_mcp_service_names(workspace, profile, on_services=on_services)
+
+    return loader
+
+
 def _discover_selected_mcp_sources(
```
xsh310, line 1513: Do we still need the discovery helper for other sources?

src/ucode/mcp.py
```diff
@@ -1890,13 +2129,30 @@ def configure_mcp_command(
...
+            sources = set(available_source_keys)
         discovered = _discover_selected_mcp_sources(workspace, profile, sources)
 
+        # The fast `system.ai` list is shown immediately; the slow workspace-wide walk
+        # streams in behind it so the picker never blocks on it.
+        services_loader = (
+            _mcp_services_background_loader(
+                workspace, profile, set(original_by_name), additive=append
+            )
+            if "mcp-services" in sources
```
xsh310: nit: would it be better if we create a constant for this string

xsh310:
Tried this with ml inference staging, it still seems to be taking a while before it shows the first non system ai mcp. It's probably because we have too many catalog and it takes time to iterate all catalogs

https://github.com/user-attachments/assets/c04251df-5866-4047-b80e-1e179f195709

xsh310 (approved):

src/ucode/mcp.py
sunishsheth2009: Fixed. Added `_StreamingInquirerControl`, an InquirerControl subclass that tolerates an empty or all-disabled choice list (the picker intentionally opens empty and streams rows in via the background loader), and guarded the space/arrow handlers against no-selectable-choice. Verified headlessly with a piped-input app session (empty, all-disabled, streaming, and normal cases) plus a regression test.

README.md
sunishsheth2009: Done — dropped the "Databricks SQL" mention from the picker-options line.

src/ucode/mcp.py
sunishsheth2009: Cleaned up. With the single MCP-services source, `prompt_sources` was always false so the prompt/back-nav loop never ran. Removed `prompt_for_mcp_search_sources` + `MCP_SEARCH_SOURCES`; `configure_mcp_command` now opens the picker directly.

src/ucode/mcp.py
sunishsheth2009: Done — extracted `MCP_SERVICES_SOURCE = "mcp-services"` and used it for the source key.

src/ucode/mcp.py
sunishsheth2009: Confirmed dead — with the single source these branches never run. Removing the `discover_*` helpers for the other sources (and the databricks `list_*` walkers they call) touches ~40 test monkeypatch stubs and the walk helpers shared with the retained mcp-services walk, so I'll do it as a focused follow-up PR to keep this one reviewable. Leaving this thread open to track it.

sunishsheth2009: Done in https://github.com/databricks/unity-gateway/pull/550 — removed the dead discover_* helpers for the other sources (and the databricks list_* walkers they used).

PR #475, opened 2026-09-03
feat(pi): Amazon Bedrock Model Provider Service support

bjanssen11:
## What this adds

Amazon Bedrock as a model provider for Pi, routed through the Databricks AI Gateway, plus two fixes that were needed to make it work end to end.

Pi talks to Bedrock over Amazon's native Converse API (`bedrock-converse-stream`). That is the only dialect the gateway serves for the `amazon_bedrock` provider type. I verified this against the live gateway: OpenAI Responses and MLflow chat-completions both return "not supported for provider Amazon Bedrock", while `/ai-gateway/model/{id}/converse-stream` returns a real reply. The Model Provider Service is picked per request through the `Databricks-Model-Provider-Service` header. Any Bedrock chat model the MPS exposes works: Claude, Nova, Llama, and the GPT-family targets all responded in testing.

## The two fixes

1. Bedrock config survives the token refresh. The launch-time refresh and the 30-minute background refresh both re-rendered `models.json` without the provider arguments, which dropped the `databricks-bedrock` block and sent the session to a system-hosted model instead. `_refresh_token_once` now reads the existing config, and when it finds a `databricks-bedrock` block it re-applies that block with a freshly refreshed token.

2. No duplicate `User-Agent`. Pi's Converse client sets its own `User-Agent`, and ucode set one too. The gateway rejected the request with "Header field 'user-agent' must only have a single value". The Bedrock block now sends only the MPS header.

## How to use it

```
ucode pi --provider <your-bedrock-mps>
```

If the MPS declares target models, they populate Pi's `/model` picker. If it is `allow_all_targets` with none declared, ucode prompts for a Bedrock model ID, for example `us.amazon.nova-lite-v1:0` or `us.anthropic.claude-sonnet-4-20250514-v1:0`. Add or remove targets on the MPS and the next launch reflects the change.

## Install and try it locally

```
git fetch origin feat/pi-bedrock
git checkout feat/pi-bedrock
uv tool install --reinstall .
ucode pi --provider eng_dev.ai_gateway.amazonbedrock
```

Tests for the fixes:

```
uv run pytest tests/test_agent_pi.py
```

## Notes for reviewers

Codex stays incompatible with Bedrock on this gateway. Codex speaks the OpenAI Responses API, which the gateway refuses for the `amazon_bedrock` provider. Only Converse works, and Pi speaks it.

This branch builds on the shared Model Provider Service code from #455. It is a draft against `main`, so until #455 merges the diff also shows that shared foundation. Once #455 lands, this collapses to the Pi provider commits plus the two fixes above.

https://claude.ai/code/session_017G9kjrbhqvWucH26GwykSn

### DIFF
--- a/src/ucode/agents/__init__.py
+++ b/src/ucode/agents/__init__.py
@@ -350,6 +350,7 @@ def configure_tool(
     relayed: bool = False,
     route_root_model: str | None = None,
     custom_model: str | None = None,
+    bedrock_targets: list[str] | None = None,
 ) -> dict:
     result: dict | tuple[dict, str]
     if tool == "codex":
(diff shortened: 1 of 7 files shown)


bjanssen11:
Superseded by #478, which is the clean, Pi-only version off main. This draft branched on top of the now-closed #455 and carried the codex launch code; #478 drops all of it and contains only the shared MPS support plus Pi Bedrock.

PR #493, opened 2026-09-04
fix nested config items in codex app

lilly-luo:
https://github.com/databricks/ucode/pull/473 was a previous attempt to fix codex app in ucode. i tested by running `uv run ucode codex app` and it worked. however, when i installed a new ucode and ran `ucode codex app`, it failed with the error below. 

this is because when i tested previously, `uv run` uses the uv.lock where as a freshly installed ucode does not respect uv.lock. we specify tomlkit>=0.13.0. the uv.lock has 0.14 pinned which works but a freshly installed ucode has tomlkit 0.15.1 which is not compatible with this change. 

i've since updated the uv.lock https://github.com/databricks/ucode/pull/496 so that uv run will be consistent with future installs so we don't have this discrepancy in the future. this PR makes a fix such that `ucode codex app` will actually work 

before:

```
➜  ucode git:(lilly/fix-codex-app-again) ucode codex app

╭────────────────────────────────────╮
│ Launching Codex with Unity Gateway │
╰────────────────────────────────────╯
✔ Starting Codex
╭────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────╮
│ /Users/lilly.luo/.local/share/uv/tools/ucode/lib/python3.12/site-packages/ucode/cli.py:2402 in codex_cmd                                       │
│                                                                                                                                                │
│   2399 │   │   print_success("Codex smart routing disabled; ucode routing hooks removed")                                                      │
.......
ValueError: Inline tables cannot contain a table

``` 

after (with the updated uv.lock):
```
➜  ucode git:(lilly/fix-codex-app-again) uv run codex app 
Opening Codex Desktop at /Applications/Codex.app...
Opening workspace /Users/lilly.luo/ucode...
```

Screenshot 2026-09-04 at 12 48 52 PM

### DIFF
--- a/src/ucode/codex_config.py
+++ b/src/ucode/codex_config.py
@@ -3,19 +3,26 @@
 from __future__ import annotations
 
 import tomlkit
+from tomlkit.items import Item
+
+
+def _toml_item(value: object) -> Item:
+    """Convert nested Python values without creating regular TOML tables."""
+    if isinstance(value, dict):
+        item = tomlkit.inline_table()
+        for key, entry in value.items():
+            item[key] = _toml_item(entry)
+        return item
+    if isinstance(value, list):
+        item = tomlkit.array()
+        for entry in value:
+            item.append(_toml_item(entry))
+        return item
+    return tomlkit.item(value)
 
 
 def _toml_value(value: str | int | float | bool | list[object] | dict[str, object]) -> str:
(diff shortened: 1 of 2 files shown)


lilly-luo:
this will be fixed here: https://github.com/databricks/ucode/pull/502

PR #495, opened 2026-09-04
fix(copilot): route Claude models through Copilot's native anthropic provider

### LINKED ISSUE #494: Copilot CLI gets zero prompt-cache hits on Claude models — ucode forces the openai provider type instead of Copilot's native anthropic one
## The problem

Copilot CLI never caches anything when talking to Claude models through ucode. Every request pays full price, even though the same big system/tool prompt repeats on every turn.

## Why

ucode always tells Copilot to use its `openai` connection mode for every model, including Claude. That mode has no way to mark "cache this part" — Anthropic's caching only works over Anthropic's own API format, which Copilot also supports (`anthropic` mode), but ucode never uses it for Copilot.

## Complication

Copilot's `anthropic` mode only works right on newer versions. On anything below Copilot **1.0.81-6**, Copilot sends a request parameter that current Claude models (Sonnet 5, Opus 5) reject outright — every request would just fail. So the fix has to check the installed Copilot version and only switch modes when it's new enough; older Copilot stays on the old (working, uncached) path.

## Fix

PR #495. Verified against a real workspace: caching now shows up in Databricks' own usage table for Copilot + Claude, on Copilot 1.0.81-6 and newer.

larsmoan:
Fixes #494.

## What this does

For Claude models, Copilot now talks to Databricks using Copilot's `anthropic` connection mode instead of `openai` mode. That's the mode that lets Copilot's own caching logic actually run. Codex (gpt-5) is unchanged, still on `openai` mode.

This only kicks in on Copilot CLI **1.0.81-6 or newer**. Below that version, Copilot sends a parameter current Claude models reject, so every request would just fail — the code checks the installed Copilot version and keeps older installs on the old (working, uncached) path.

## Proof

Tested against a real workspace across Copilot versions 1.0.79 through 1.0.83, and checked Databricks' own usage table (`system.ai_gateway.usage`), filtered to Copilot's actual requests (not this session's).

- **Before, on 1.0.79/1.0.80** (what most people are likely running): requests succeed, but never cache — matches the original bug report exactly.
- **After, on 1.0.81-6+, with this fix**: first request per model writes the cache, every request after that reads it back (confirmed in the usage table, not just Copilot's own token counter).

## Test plan

- `uv run pytest -q` passes (one unrelated pre-existing failure, same on `main`)
- `uv run ruff check .` passes
- Verified live as above


larsmoan:
**Update: validated against a live workspace, and the original fix had two real bugs.**

I ran actual `copilot` sessions with and without this fix against a real Databricks workspace, and checked the resulting requests. The original version of this PR did *not* work end-to-end:

1. **401, not a cache miss.** Copilot's `anthropic` provider authenticates with `x-api-key` by default (`COPILOT_PROVIDER_API_KEY`, what this PR originally set). Databricks' AI Gateway rejects that outright — confirmed directly with curl against `/ai-gateway/anthropic/v1/messages`: `x-api-key` → `401 Credential was not sent or was of an unsupported type`; `Authorization: Bearer` → `200`. Fixed by using `COPILOT_PROVIDER_BEARER_TOKEN` for both provider types (`copilot help providers` confirms it "takes precedence over API key" regardless of provider type).
2. **400 on every current-gen Claude model.** With auth fixed, every request to `claude-sonnet-5`/`claude-opus-5` still 400'd: `` `temperature` is deprecated for this model ``. Per `copilot help providers`, an unrecognized `COPILOT_MODEL` (a Databricks catalog id like `system.ai.claude-sonnet-5` isn't one Copilot knows) makes it "fall back to safe defaults" — which includes sending `temperature`, and current-generation Claude models reject that outright. Fixed by splitting `COPILOT_PROVIDER_MODEL_ID` (a canonical name extracted from the model id, e.g. `claude-sonnet-5`) from `COPILOT_PROVIDER_WIRE_MODEL` (the actual Databricks id sent on the wire), and leaving `COPILOT_MODEL` unset for the Claude path.

With both fixes, live requests against `claude-sonnet-5` and `claude-opus-5` succeed, and Copilot's own usage line shows real cache reads on repeated invocations against the same session config, e.g.:

```
run 0: Tokens ↑ 25.8k (25.8k written)
run 1: Tokens ↑ 25.8k (13.7k cached, 12.1k written)
run 2: Tokens ↑ 25.8k (13.7k cached, 12.1k written)
```

That "cached" figure is Copilot's own display of the Anthropic response's `usage.cache_read_input_tokens` field (confirmed in the runtime binary strings), not something ucode or I computed — so this is the model actually reading from cache, not an estimate.

I have *not* yet been able to cross-check this against `system.ai_gateway.usage` itself — the table on this workspace hadn't ingested anything from today, even ~1 hour after the test requests (last row was from the previous day), so there's real ingestion lag I couldn't wait out in this session. The live client-side evidence above is strong, but the PR stays a draft until someone (or a later session) confirms `cache_read_input_tokens > 0` server-side too.

Pushed the corrected commit: 8d7bc13.

larsmoan:
@lilly-luo @rohita5l @andy-xu-db

This PR is current with main and the focused Copilot, Databricks, CLI, and lint tests pass. The automated reviews approved the implementation.

Can you please confirm who owns the merge decision and what blocks progress? Unity Gateway cannot currently support Claude models correctly through the Copilot CLI path. The current OpenAI-compatible path cannot carry Copilot cache-control breakpoints. Current Claude models can also reject the fallback request parameters. This makes Copilot unusable for the intended Unity Gateway Claude workflow today.

Please state the remaining technical or product concern, or approve and merge this fix.

larsmoan:
Addressed the review finding. The Copilot launch environment removes inherited legacy authentication and model-selection variables before applying the selected provider configuration. Added regression coverage for stale provider and model variables.

PR #496, opened 2026-09-04
upgrade uv.lock

lilly-luo:
- [x] uv run ucode claude works 
- [x] uv run ucode codex works 

when testing codex app server, i ran into a problem where ucode codex app worked with uv run but not when i actually installed the new ucode. this was because uv run was using uv.lock which had toml at 0.14 but when users actually instead ucode, they get toml 0.15.1 (on which ucode codex app is broken). 

this PR:
1. updates uv.lock
2. makes the necessary changes for mcp to match the new versions


src/ucode/cli.py
```diff
@@ -2990,10 +2990,12 @@ def configure(
             # The workspaces were just configured, so enable tracing for them
             # directly instead of re-prompting. Fall back to the workspace that
             # `configure_workspace_command` made current (the interactive pick).
-            tracing_workspaces = workspace_entries
+            tracing_workspaces: list[tuple[str, str | None]] | None = workspace_entries
             if tracing_workspaces is None:
                 current = load_full_state().get("current_workspace")
-                tracing_workspaces = [(current, None)] if current else None
+                tracing_workspaces = (
+                    [(current, None)] if isinstance(current, str) and current else None
```
sunishsheth2009, line 2997: This `tracing_workspaces` re-typing + `isinstance(current, str)` guard is unrelated to the uv.lock / mcp purpose of the PR, and it's a silent behavior change: previously `[(current, None)] if current` accepted any truthy `current`; now only a non-empty `str` qualifies. If `current_workspace` in state is ever a non-str truthy value, it's now dropped instead of passed through. Low risk (it's arguably a hardening), but it's scope creep in a lockfile-bump PR — consider splitting it out, or add a line to the description explaining why it's bundled here.

sunishsheth2009 (commented):
Reviewed the code and verified the proxy against the new lock (checked out `lilly/upgrade-uv-lock`, `uv sync --frozen`, ran the suite). Summary:

**✅ Proxy verified working on the new lock.** `uv sync --frozen` resolves **mcp 2.1.1 / httpx2 2.12.0 / anyio 4.14.2**; `tests/test_mcp_proxy.py` is 29/29 green and `ty` is clean. The `mcp_proxy.py` change is annotation-only — concrete `MemoryObjectReceiveStream`/`MemoryObjectSendStream` types swapped for structural `_ReceiveStream`/`_SendStream` Protocols, and `_pump`/`_pump_upstream` made generic. The function bodies are unchanged, so runtime behavior is identical. Sensible decoupling from the concrete anyio stream types (which is what lets the same code path work under both mcp majors).

**🔴 Description item 3 ("excludes gpt-5-6- astra from tests") is not in this PR.** The diff touches only `cli.py`, `mcp_proxy.py`, and `uv.lock` — no test file. The astra-skip logic (`test_astra_is_skipped`, `test_web_search_skips_astra`, the `"astra"` frozenset in `tests/test_e2e.py`) already exists on `main`; this PR adds nothing to it. So either the description is stale and should drop item 3, or a commit was dropped and needs pushing. Please reconcile — approving off the description would approve something that isn't here.

One inline note on the unrelated `cli.py` change below.

sunishsheth2009 (approved):

PR #500, opened 2026-09-04
Stop `ug configure` erroring on feature-disabled workspaces

david-siqi-liu:
## 🥞 Stack (ug configure + managed-config)

- **#500** ⬅ this PR
  - #504
    - #510

---

Customers running `ug configure` on a workspace where server-managed coding-agent config is disabled (FEATURE_DISABLED) hit an alarming, circular error and then dead-ended:

    ERROR Workspace-managed coding agent configuration is not available on this workspace. Use `ug configure` to set up agents for individual users instead.

Root cause: `refresh_managed_config` treated FEATURE_DISABLED like a transient read failure and fell back to a stale cached config, returning it with the feature-disabled flag unset. That masked the disabled state, so an admin got routed into the managed setup flow, which re-read the workspace, got FEATURE_DISABLED again, and raised the message.

The fix is in `managed_config.py`: FEATURE_DISABLED is now authoritative. It returns no config with the feature-disabled flag set and clears the persisted cache, so `ug configure` stays on the normal per-user flow, a launch never re-applies a policy the workspace turned off, and a later transient read or token failure cannot resurrect the disabled policy through the fallback. Also initializes the feature-disabled flag on the `--dry-run` path (a pre-existing UnboundLocalError on an empty local cache, surfaced by review).

Hiding `ug setup`/`ug publish` from help and pruning the README, previously bundled here, moved to standalone #510 since it has no dependency on this fix.

## How do you know it works?

Unit suites (managed_config, cli) green, including new coverage for the feature-disabled cache clear and the dry-run no-crash path. ruff check and format clean. PR CI green (test + e2e).

Live-verified end to end against a real feature-disabled staging workspace (eng-ml-agent-platform, with the `codingAgentConfigCrudEnabled` SAFE flag turned off), where the gateway returns:

```json
{"error_code":"FEATURE_DISABLED","message":"Coding agent config APIs are not enabled for this workspace."}
```

Invoking `ug configure` against that workspace as an admin, with a non-empty stale managed config cached (the exact condition that triggered the report):

Old build (`0.1.0+81.g95999aa`) exits 1 with the circular dead-end error:

```
ERROR Workspace-managed coding agent configuration is not available on this workspace. Use `ug configure` to set up agents for individual users instead.
```

This branch exits 0 with no error and drops into the normal per-user configure flow. Under the hood `refresh_managed_config` now returns `(None, True)` and clears the stale cache to `{}`, so the admin is never routed into the managed-setup dead end.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -74,6 +74,7 @@ from ucode.managed_budget import (
 )
 from ucode.managed_config import (
     MANAGED_CONFIG_ENV_VAR,
+    ManagedConfigResult,
     get_model_recommendation,
     load_managed_state,
     managed_agent_config_enabled,
@@ -1802,15 +1803,16 @@ def _reject_disabled_agent(managed: dict | None, tool: str) -> None:
         )
 
 
-def _fetch_managed_config(state: dict) -> tuple[dict | None, bool]:
-    """The workspace's managed config for this launch, or ``(None, _)`` when there is none.
+def _fetch_managed_config(state: dict) -> ManagedConfigResult:
+    """The workspace's managed config for this launch, or a manifest-less result when there is none.
 
-    Returns ``(None, False)`` when managed configs are switched off — either the feature is disabled
-    or the launch passed ``--skip-managed-config`` (which clears the enabling env var for the process).
+    Returns a feature-disabled-free result when managed configs are switched off — either the feature
+    is disabled or the launch passed ``--skip-managed-config`` (which clears the enabling env var for
+    the process).
     """
 
     if not managed_agent_config_enabled():
-        return None, False
+        return ManagedConfigResult(None, False)
     with spinner("Loading..."):
         return refresh_managed_config(state)
 
@@ -2493,7 +2495,7 @@ def _launch_managed_default(
     if not current:
         raise RuntimeError("No workspace configured. Run `ug configure` first.")
     apply_pat_environment(state)
-    # --dry-run avoids the fetch but still applies the last saved config.
+    coding_agent_config_feature_disabled = False
     if dry_run:
         managed = load_managed_state(current)
     else:
--- a/src/ucode/managed_config.py
+++ b/src/ucode/managed_config.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 import json
 import os
 from pathlib import Path
-from typing import cast
+from typing import NamedTuple, cast
 
 import ucode.config_io as config_io
 from ucode.databricks import (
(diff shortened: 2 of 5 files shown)


src/ucode/managed_config.py
```diff
@@ -431,10 +435,20 @@ def refresh_managed_config(state: dict) -> tuple[dict | None, bool]:
...
-        # A refused read leaves the cached config alone: it says nothing about whether the admin's
-        # config still exists, unlike a successful "no config" answer below.
+        if _is_feature_disabled(reason):
+            # FEATURE_DISABLED is an authoritative "the feature is off server-side", not a transient
+            # read failure — so don't fall back to a config cached from when it was enabled. Doing so
+            # would keep applying a policy the workspace has turned off, and (via a stale cache
+            # masking the disabled state) route `ug configure` into a managed-setup flow that
+            # immediately dead-ends. Clear the persisted copy too (same as the "no config" branch
+            # below): the file doubles as the fallback, so leaving it would let a later transient
+            # read/token failure resurrect the disabled policy. Report "no config, feature off".
```
lilly-luo: can you rm all these comments

src/ucode/cli.py
```diff
@@ -1171,6 +1171,10 @@ def revert() -> int:
 app.add_typer(
     setup_app,
     name="setup",
+    # Hidden from `ug --help` while workspace-managed configs are still in development and not
+    # ramped: the commands stay runnable (the flow is gated behind ENABLE_MANAGED_AGENT_CONFIG),
+    # they just aren't advertised to customers yet. Unhide alongside the public rollout.
```
lilly-luo: rm this

src/ucode/cli.py
```diff
@@ -2419,7 +2423,11 @@ def _launch_managed_default(
     if not current:
         raise RuntimeError("No workspace configured. Run `ug configure` first.")
     apply_pat_environment(state)
-    # --dry-run avoids the fetch but still applies the last saved config.
+    # --dry-run avoids the fetch but still applies the last saved config. It doesn't reach the
+    # gateway, so the feature-disabled signal is unknown here — default it to False so the guidance
+    # check below is well-defined (without this, a dry run with no cached config raised
+    # UnboundLocalError).
+    coding_agent_config_feature_disabled = False
```
lilly-luo: can you just return here and comment out everything below? or at least make this a hidden env var that we  can use to test/

src/ucode/cli.py
```diff
@@ -2419,7 +2419,8 @@ def _launch_managed_default(
     if not current:
         raise RuntimeError("No workspace configured. Run `ug configure` first.")
     apply_pat_environment(state)
-    # --dry-run avoids the fetch but still applies the last saved config.
+    # --dry-run doesn't fetch, so default the feature-disabled flag rather than leave it unbound.
```
lilly-luo: maybe rm this comment

src/ucode/managed_config.py
lilly-luo: nit maybe make this return a dataclass instead of a tuple

lilly-luo (approved):

PR #503, opened 2026-09-05
use models from model picker on CC as priority for smart routing 

lilly-luo:
as titled. i tested: 

```
  "modelPicker": {
    "replaceBuiltInOptions": true,
    "options": [
      {
        "model": "system.ai.claude-opus-4-8"
      },
      {
        "model": "system.ai.glm-5-2"
      },
      {
        "model": "system.ai.glm-5-3-flash"
      },
      {
        "model": "system.ai.claude-sonnet-4-6"
      }
    ]
  }
```

- [x] added the 4 models in /etc/claude-code/managed-settings.json and confirmed that smart routing worked and these were the only 4 models sent 
- [x] added the 4 models in ~/.claude/ucode-settings.json and confirmed that smart routing worked and these were the only 4 models sent (nothing in managed-settings.json)
- [x] added the 4 models in ~/.claude/settings.json and confirmed that smart routing worked and these were the only 4 models sent (nothing in ucode-settings.json or managed settings).


src/ucode/agents/claude.py
```diff
@@ -653,6 +653,9 @@ def _compose(base: dict, *, enforce_model_default_hierarchy: bool) -> dict:
                 else:
                     target_env[key] = selected_default_model
         merged = deep_merge_dict(base, overlay_for_merge)
+        if enforce_model_default_hierarchy and "modelPicker" in base:
+            # Claude's managed modelPicker is administrator-owned; ucode must leave it intact.
```
david-siqi-liu: > ucode must leave it intact.

but managed settings (and `modelPicker` inside) is updated by `ug configure` right?

src/ucode/smart_routing/v2.py
```diff
@@ -348,7 +390,8 @@ def launch_claude(
     os.environ[OAUTH_TOKEN_ENV_VAR] = token
     os.environ[GATEWAY_MODEL_DISCOVERY_ENV_VAR] = "1"
     os.environ["CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY"] = "1"
-    catalog = list_anthropic_model_catalog(workspace, token)
+    # modelPicker takes priority over model discovery.
+    catalog = _model_picker_catalog() or list_anthropic_model_catalog(workspace, token)
```
david-siqi-liu, line 394: i thought we are always using modelPicker?

lilly-luo: not if automatic discovery is enabled

src/ucode/agents/claude.py
lilly-luo: yes you're right, rm'ing this

david-siqi-liu (approved):

PR #515, opened 2026-09-07
smart routing: fix codex first prompt route

lilly-luo:
the intended flow for smart routing is :
1/ user sends first prompt 
2/ smart router returns recommended model 
3/ first prompt runs on the recommeded model 
4/ subsequent prompts run on the recommended model. 

instead, what was happening was: 
1/ user sends first prompt 
2/ smart router returns recommended model 
3/ first prompt runs on the DEFAULT model for the session (not the recommended model) 
4/ subsequent prompts run on the recommended model. 

this is because the [interposer](https://docs.google.com/document/d/1lO1l3sOH0V9_YIx2SQQIvDOBlBJfIwpITLFl1rYTZ7s/edit?tab=t.whcgscnu6a6r) only overwrote the top-level model in turn/start which was not used for the first prompt, only for subsequent prompts. this PR makes it so that when the smart router picks a model, we send a `thread/settings/update` request to the app-server (similar to invoking /model) before sending the first prompt so that the model actually changes. 

test
before:
Screenshot 2026-09-07 at 7 38 56 PM

after:
Screenshot 2026-09-07 at 7 38 36 PM

### DIFF
--- a/src/ucode/smart_routing/codex_interposer.py
+++ b/src/ucode/smart_routing/codex_interposer.py
@@ -48,7 +48,7 @@ def _prompt_from_turn(params: dict) -> str | None:
 @dataclass
 class TuiFrameResult:
     frame: str
-    needs_settings_update: bool = False
+    needs_settings_update: bool
 
 
 class _Session:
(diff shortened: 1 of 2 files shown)


david-siqi-liu (approved):

src/ucode/smart_routing/codex_interposer.py
```diff
@@ -225,11 +244,32 @@ def route_decision(prompt: str):
...
+                        update_req = json.dumps(
+                            {
+                                "id": update_id,
+                                "method": SETTINGS_UPDATE,
+                                "params": {
+                                    "threadId": sess.thread_id,
+                                    "model": sess.target,
+                                },
+                            }
+                        )
```
lilly-luo, line 259: this is the main change - we invoke a /update call to the codex thread

PR #520, opened 2026-09-08
Support per-client skill MCP scopes

xsh310:
## What

Store the skills MCP connection's scope as a **per-agent map** (`skill_locations_by_client`) instead of a single flat `skill_locations` list. Each configured coding agent (Claude, Codex, …) gets its own set of skill schemas and is registered with its own scoped gateway URL, so different agents can carry different skills.

This PR is the **state model** that lets later PRs in the stack target individual agents (`skill add`/`skill remove --agents …`). On its own it changes no user-facing command behavior — every command still operates on all configured agents — it just stores the scope per agent and reads/writes it coherently.

Backward compatible: connections written by older builds (flat `skill_locations` only) still read correctly, and `skill_locations` is kept as a union mirror for older readers.

## State shape

The skills connection is a single entry in `state.json`'s `mcp_servers` (`kind: "skills"`, name `databricks-skill-registry`):

```jsonc
{
  "name": "databricks-skill-registry",
  "kind": "skills",
  "clients": ["claude", "codex"],
  "skill_locations_by_client": {                // NEW — authoritative, per-agent scope
    "claude": ["system.ai", "team.ml"],
    "codex":  ["system.ai"]
  },
  "skill_locations": ["system.ai", "team.ml"],  // union across agents (legacy readers only)
  "url": ".../ai-gateway/skills/?schema=system.ai&schema=team.ml",
  "auth": "proxy"
}
```

- **`skill_locations_by_client`** — the source of truth: one `catalog.schema` list per agent.
- **`skill_locations`** — derived, the union of every agent's list. Kept only so an older build or any legacy reader still sees a coherent scope.
- Each agent is registered with a gateway URL scoped to **its own** list.

## What happens on a skill command

- **`ug skill add --mcp <catalog.schema,…>`**: unions the given schemas into every configured agent's list (keeping what each already has), re-registers only the agents whose scope actually changed (each with its own scoped URL), and saves the per-agent map plus a refreshed union mirror.
- **`ug configure skills --mcp <catalog.schema,…>`**: *sets* (replaces) every configured agent's list to exactly the given schemas.
- **`ug configure skills` with no `--location`, and the on-launch download path**: registers/keeps the schema-less connection (utility tools only) while preserving each agent's existing scope.
- **Reads** : `skill_locations_for_client(entry, client)` returns that agent's list: from `skill_locations_by_client` when present, otherwise the flat `skill_locations` mirrored to every agent, so pre-upgrade state reads identically.
- **Migration**: no eager rewrite: a legacy flat entry is upgraded to the per-agent map the next time any skill command writes the connection.
- **PAT users:** registration threads `use_pat` from state (`use_pat=bool(state.get("use_pat"))`), like every other `apply_mcp_server_changes` caller. For anyone with `use_pat` set this changes the generated `ug mcp-proxy` argv and fixes the skills proxy for PAT-only users; non-PAT registration is unchanged.

## Stacking

**Stacked on #519** (`skill-managed-download-only`). With managed skills delivered by download only, the skills MCP connection holds only a developer's own `skill add --mcp` schemas, so this per-agent scope is purely developer state — there is no global managed set to union in.

First PR of the per-agent skills stack (this → `add --agents` → per-agent `status` → `remove` → per-agent `remove` → deprecate `configure skills`). Base moves to `main` once #519 merges.

Original design and initial implementation by Arthur Jenoudet; reworked into the per-client model here.

### DIFF
--- a/src/ucode/mcp.py
+++ b/src/ucode/mcp.py
@@ -100,6 +100,7 @@ MCP_CLIENTS = {
 }
 SKILLS_MCP_KIND = "skills"
 SKILLS_MCP_SERVER_NAME = "databricks-skill-registry"
+SKILL_LOCATIONS_BY_CLIENT_KEY = "skill_locations_by_client"
 # MCP-only clients ucode never launches for model routing, so they never land in
 # `available_tools`; they're eligible for MCP config purely on being installed.
 MCP_ONLY_CLIENTS = ("cursor",)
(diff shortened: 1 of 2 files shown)


src/ucode/mcp.py
```diff
@@ -2106,56 +2164,117 @@ def _print_skills_summary(entry: dict) -> None:
...
     workspace, profile, clients = setup_mcp_clients(state, "Skills MCP")
-    _update_skills_mcp(state, workspace, profile, clients, locations)
+    locations_by_client = _skill_locations_by_client_from_state(state)
+    for client in clients:
+        locations_by_client[client] = list(locations)
+    _update_skills_mcp(state, workspace, profile, clients, locations_by_client)
     return 0
 
 
 def _skill_mcp_locations(state: dict) -> list[str]:
```
xsh310, line 2283: Is this function still used anywhere?

xsh310: Yes, still used — just not on the skills-MCP read path anymore.

Its one remaining caller is `managed_wizard._skill_names_from_state` (imported locally), which `setup_skills_command` (`ug setup skills`) calls to populate the published manifest's `skills.names`:

```python
configure_skills_mcp_command(locations)
skill_names = _skill_names_from_state(load_state()) or locations
manifest["skills"] = {"names": skill_names}
```

It looks orphaned in this diff because this PR moved the skills-MCP read path onto the per-client map (`_skill_locations_by_client_from_state`), so both former in-module callers migrated off it — leaving `managed_wizard` as the sole caller. That manifest only needs the union of schema names (not per-agent scopes), so reading the flat `skill_locations` mirror via `_skill_mcp_locations` is the correct source there. Leaving it as-is.

src/ucode/mcp.py
```diff
@@ -2106,56 +2164,117 @@ def _print_skills_summary(entry: dict) -> None:
...
-    working = _resolve_skills_mcp_servers(workspace, clients, locations, original)
-    changed = apply_mcp_server_changes(original, working, clients, workspace, profile)
+    working = _resolve_skills_mcp_servers(workspace, clients, locations_by_client, original)
+    original_entry = _skills_entry(original)
+    working_entry = _skills_entry(working)
+    if working_entry is None:
+        raise RuntimeError("Failed to build the Skills MCP connection.")
+
+    changed = False
+    for client in clients:
```
sunishsheth2009: **P1 (perf):** this per-client loop replaced the single `apply_mcp_server_changes(..., clients)` call, and that function was built to run the per-client CLI re-registrations *concurrently across clients* (its `ThreadPoolExecutor`). Iterating here serializes them, and opens a separate "Configuring MCP servers…" spinner per changed client. In every flow this stack ships — and any time agents share a scope — all clients resolve to the same URL, so we pay N sequential `claude mcp add-json` / `codex` / etc. shell-outs for no benefit (worst with two CLI-backed agents like claude + gemini).

Could we group clients by their resolved scope-tuple and make one `apply_mcp_server_changes` call per *distinct* scope? Same-scope agents share a URL so they go in one concurrent call (common case → a single call, old behavior restored), and genuinely divergent per-agent scopes still get their own call.

src/ucode/mcp.py
```diff
@@ -2106,56 +2164,117 @@ def _print_skills_summary(entry: dict) -> None:
...
+                )
+            ]
+        changed = (
+            apply_mcp_server_changes(
+                original_view,
+                working_view,
+                [client],
+                workspace,
+                profile,
+                use_pat=bool(state.get("use_pat")) if use_pat is None else use_pat,
```
sunishsheth2009: **P2 (note, not a bug):** this now threads `use_pat=bool(state.get("use_pat"))`, where the pre-stack skills path called `apply_mcp_server_changes` without it (defaulting `False`). It aligns skills registration with every other `apply_mcp_server_changes` caller and actually fixes the skills proxy for PAT-only users, so it looks intentional — just worth a line in the description, since it changes the generated `mcp-proxy` argv for anyone with `use_pat` in their state.

sunishsheth2009 (approved):

src/ucode/mcp.py
xsh310: Done in `368a5ab`. Rather than group by scope-tuple, I extracted the concurrent runner out of `apply_mcp_server_changes` into a shared `_run_client_work(work)`, and added `apply_skills_mcp_changes` that builds a single per-client work map (each client → its own scoped URL) and runs it in one `_run_client_work` batch. All changed clients are now (re)configured concurrently under one spinner — uniform or divergent scopes alike, common case back to a single pass — and the per-client `_update_skills_mcp` loop (with its per-client spinner) is gone.

src/ucode/mcp.py
xsh310: Good call — added a note to the description ("What happens on a skill command" → PAT users bullet): skills registration now threads `use_pat` from state to match the other `apply_mcp_server_changes` callers, which changes the generated `mcp-proxy` argv for `use_pat` users and fixes PAT-only setups.

PR #522, opened 2026-09-08
Add per-agent scoping to skill add --mcp

xsh310:
## What

Adds an `--agents` option to `ucode skill add --mcp` so a skill schema can be added to a chosen subset of configured coding agents instead of all of them. This mirrors `ucode mcp add --agents`, closing the gap between how the skills MCP path and the general MCP path scope work per agent.

## How

- `add_skills_command(locations, agents=None)` forwards `agents` to `setup_mcp_clients`, which scopes the returned client set. The per-client skill-location map is updated only for the targeted clients; the rest keep their scope untouched.
- In `cli.py`, `skill add` gains the `--agents` option, bootstraps any named agent that is not configured yet via `_configure_agents_for_mcp`, and guards that `--agents` is only valid with `--mcp` (downloaded skills use shared directory families and cannot be agent scoped).
- Without `--agents`, behavior is unchanged: every configured agent is updated.

## Tests

- `test_mcp.py`: per-agent add touches only the named client's scope and re-registers only that client; a global add reaches every configured client while preserving per-client divergence; adding a schema an agent already has is a no-op.
- `test_cli.py`: `--agents` is parsed, bootstrapped, and forwarded for `--mcp`; an empty `--agents` list is rejected; `--agents` is rejected in download mode.

Installed the stack tip with `uv tool install --reinstall .` (ucode `0.1.0+91.g5c0dc1e`) and exercised the command in an isolated sandbox: a temporary `HOME`, stub `claude`/`codex` binaries that record every registration ucode issues, and `DATABRICKS_BEARER` set so auth is satisfied offline. State was seeded with both agents configured and no skills.

| Scenario | Command | Result |
|---|---|---|
| Global add (all agents) | `ucode skill add --location shared.skills --mcp` | Both agents scoped to `shared.skills`; each re-registered once |
| Per-agent add | `add_skills_command(["claude.only"], agents={"claude"})` | `claude=[shared.skills, claude.only]`, `codex=[shared.skills]`; only claude re-registered; summary showed divergent per-client URLs |
| Guard: no `--location` | `ucode skill add` | `ERROR --location is required for ucode skill add.` |
| Guard: `--agents` without `--mcp` | `ucode skill add --location a.b --agents claude` | `ERROR --agents is only supported when using --mcp` |
| Guard: empty `--agents` | `ucode skill add --location a.b --mcp --agents ,` | `ERROR No agents provided for --agents ...` |

The `--agents` add path first runs an interactive agent bootstrap, so the per-agent case drove `add_skills_command(..., agents=...)` directly against the installed package with only that prompt stubbed; state I/O and per-client agent registration ran for real. CLI parsing and bootstrap are covered by the mocked unit tests.

## Stacking

This is the first of a stacked series that adds per-agent scoping to the skills MCP connection, built on top of `xshen/per-client-skill-scopes` (per-client skill scope map) and `skill-managed-download-only` (managed skills delivered by download only). Reviewing the diff against its base branch shows just this change. It rebuilds behavior originally designed by Arthur Jenoudet on the current per-client-map state model.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -107,8 +107,10 @@ from ucode.mcp import (
     add_mcp_command,
     add_skills_command,
     apply_managed_mcp_servers,
+    available_mcp_clients,
     configure_mcp_command,
     configure_skills_mcp_command,
+    configured_mcp_clients,
     purge_cross_workspace_mcp_residue,
     remove_mcp_command,
     revert_mcp_configs,
@@ -1281,6 +1283,15 @@ def skills_add(
             "Not valid with --mcp.",
         ),
     ] = None,
+    agents: Annotated[
+        str | None,
+        typer.Option(
+            "--agents",
+            help="(--mcp only) Comma-separated coding agents whose skills MCP scope should "
+            "be updated. Any that aren't configured yet are set up first. Without --agents, "
+            "every configured agent is updated.",
+        ),
+    ] = None,
 ) -> None:
     """Add Databricks Skills to your coding tools, keeping any already configured.
 
@@ -1296,6 +1307,16 @@ def skills_add(
         requested_skills = (
             None if skills is None else {s.strip() for s in skills.split(",") if s.strip()}
         )
+        requested_agents = None
+        if agents is not None:
+            requested_agents = {
+                agent.strip().lower() for agent in agents.split(",") if agent.strip()
+            }
+            if not requested_agents:
+                raise RuntimeError(
+                    "No agents provided for --agents. Use a comma-separated list like "
+                    "`--agents claude,codex`."
+                )
         if mcp and path is not None:
             raise RuntimeError("--path is not supported when using --mcp")
         if mcp and requested_skills is not None:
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -1305,7 +1324,15 @@ def skills_add(
             None if requested_skills is None else {s.split(".")[-1] for s in requested_skills}
         )
         if mcp:
-            add_skills_command(locations)
+            if requested_agents:
+                scope = {a if a == "cursor" else normalize_tool(a) for a in requested_agents}
```
sunishsheth2009: **P1:** this reimplements the agent normalization + cursor special-case + bootstrap inline, whereas the sibling `mcp add --agents` (just above) routes through `_configure_agents_for_mcp(sorted(requested_agents))` and uses its return as the scope. Same intent, two code paths that can drift — e.g. here we bootstrap only `scope - ready`, while `mcp add` reconfigures all named agents. Any reason not to reuse `_configure_agents_for_mcp` (and its returned canonical set) here too? Minor: `load_state()` also runs here and again inside `add_skills_command`.

sunishsheth2009 (approved):

src/ucode/cli.py
xsh310: Done in `caa792e`. Moved the normalize + cursor + bootstrap logic into `_configure_agents_for_mcp` — it now skips already-configured agents and returns the full canonical scope, so `skill add --mcp --agents` and `mcp add --agents` share one path: the inline reimplementation is gone, and `mcp add` now also skips reconfiguring already-set-up agents, so the two no longer drift.

On the `load_state()` nit: it still runs once in the helper and once in `add_skills_command`, but that now matches `mcp add`'s shape. Fully deduping would mean threading state through both shared signatures for negligible gain, so I left it.

PR #523, opened 2026-09-08
Show skill MCP scopes per agent in status

xsh310:
## What

`ug status` rendered a single `Skill MCP Locations` line from the flat scope mirror. Once agents carry different skill scopes (which per-agent add now makes possible), that single line hides the divergence. This makes `status` report each agent's real scope.

## How

- The Skills section reads each configured client's scope with `skill_locations_for_client`.
- When every agent shares the same scope, it collapses to the original single `Skill MCP Locations` + `Configured` lines.
- When scopes diverge, it prints a per-agent line (`Claude Code skill MCP locations: ...`, `Codex skill MCP locations: ...`), matching how `_print_skills_summary` already reports scopes.

## Tests

- `test_cli.py`: a state whose per-client scopes diverge renders one line per agent with that agent's own locations. The existing uniform-scope and no-locations status tests still pass unchanged.

Using the sandbox from the per-agent add PR (agents scoped to divergent locations: claude `shared.skills, claude.only`, codex `shared.skills`), ran `ucode status`. The Skills section rendered one line per agent:

```
Skills
  Claude Code skill MCP locations: shared.skills, claude.only
  Codex skill MCP locations: shared.skills
```

The new `ug skill add` / `ug skill remove --mcp` hint also appeared. The uniform-scope collapse path is covered by the existing status tests.

## Stacking

Second in the stacked per-agent skills series, based on `xshen/skill-per-agent-add` (#522). Reviewing the diff against that base shows just this change. It rebuilds behavior originally designed by Arthur Jenoudet on the current per-client-map state model.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -114,6 +114,7 @@ from ucode.mcp import (
     purge_cross_workspace_mcp_residue,
     remove_mcp_command,
     revert_mcp_configs,
+    skill_locations_for_client,
 )
 from ucode.skills_download import (
     configure_skills_download_command,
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -955,17 +956,32 @@ def status() -> int:
...
-            if client in MCP_CLIENTS
+        configured_clients = [
+            client for client in (skill_mcp_entry.get("clients") or []) if client in MCP_CLIENTS
         ]
-        print_kv("Configured", ", ".join(configured_agents) if configured_agents else "none")
+        scopes = {
+            client: skill_locations_for_client(skill_mcp_entry, client)
+            for client in configured_clients
+        }
+        # Collapse to one line when every agent shares a scope; split per agent when they diverge.
```
sunishsheth2009: **P2 (cleanup):** this uniform-vs-diverged collapse is the same logic `_print_skills_summary` already runs (`skill_locations_for_client` per client → single line when the scope-tuples match, else one line per agent). Worth factoring the shared shape into a small helper so `status` and the post-command summary can't drift in how they render per-agent scopes.

sunishsheth2009 (approved):

src/ucode/cli.py
xsh310: Done in `db6d6ef`. `status` and `_print_skills_summary` now share the collapse decision via `agents_share_one_scope(scopes)` (one line when scopes match, else one per agent) and use the same empty-scope fallback (`[]`) — `status` previously fell back to the flat `skill_locations` mirror. I kept the per-client scope dict inline in each caller rather than add another `*_by_client` helper next to the existing `_skill_locations_by_client`, to avoid a near-duplicate name.

PR #525, opened 2026-09-08
Scope skill remove --mcp to named agents

xsh310:
## What

Adds an `--agents` option to `ucode skill remove --mcp` so a schema can be removed from a chosen subset of configured agents and kept on the rest, mirroring `ucode mcp remove --agents`. This is the core bug fix in the series.

## The bug

Before per-agent scopes, an all-agent add wrote a shared layer while a per-agent remove edited a separate one, so they did not compose: `skill add --mcp X` (all) followed by `skill remove --mcp --agents claude` was a silent no-op, and often printed a misleading "nothing to remove". Now that developer scope is a single per-agent map, the two operations act on the same layer.

## How

- `remove_skills_command(agents=None)` forwards `agents` to `setup_mcp_clients`, which scopes the client set. The picker offers only those clients' schemas and removal edits only their maps; other agents keep the schema and are not re-registered.
- The "nothing to remove" note names the scope when `--agents` was given.
- `cli.py`: `skill remove` gains the `--agents` option; without it, removal stays global.

## Tests

- `test_mcp.py`: add-all then `remove --agents claude` removes the schema for claude only and re-registers only claude; the picker is offered only the named clients' scope; an empty named scope is a no-op.
- `test_cli.py`: `--mcp` forwards `agents=None`; `--mcp --agents claude,codex` forwards the parsed set.

This is the headline fix, so I verified the exact previously-failing sequence end to end. In the sandbox (installed build `0.1.0+91.g5c0dc1e`, stub `claude`/`codex` binaries recording registrations, offline auth via `DATABRICKS_BEARER`), `shared.skills` had been added to both agents. Then I invoked `remove_skills_command(agents={"claude"})` with the picker stubbed to select `shared.skills`:

| Check | Result |
|---|---|
| Picker input | Offered only the named client's scope: `{claude:[shared.skills, claude.only]}` |
| State after | `shared.skills` removed from claude only: `{claude:[claude.only], codex:[shared.skills]}` |
| Registrations | Only claude re-registered; codex untouched |

Before this change the same sequence was a silent no-op (add wrote a shared layer, per-agent remove edited a different one). It now removes for the named agent while the other keeps the schema.

## Stacking

Fourth in the stacked per-agent skills series, based on `xshen/skill-remove` (#524). Reviewing the diff against that base shows just this change. It rebuilds behavior originally designed by Arthur Jenoudet on the current per-client-map state model.

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -1400,15 +1400,33 @@ def skills_remove(
             help="Remove schemas from the skills MCP connection instead of downloaded files.",
         ),
     ] = False,
+    agents: Annotated[
+        str | None,
+        typer.Option(
+            "--agents",
+            help="Comma-separated coding agents to remove the schemas from (e.g. claude,codex). "
+            "A schema scoped to several agents is removed only from the named ones and kept on "
+            "the rest. Without --agents, a selected schema is removed from every agent it's on.",
+        ),
+    ] = None,
 ) -> None:
-    """Interactively remove Skill schemas from the skills MCP connection."""
+    """Interactively remove Skill schemas from the skills MCP connection.
+
+    Without ``--agents`` a selected schema is removed from every configured agent; ``--agents``
+    scopes the removal to the named agents and keeps the schema on the rest.
+    """
     try:
         if not mcp:
             raise RuntimeError(
                 "Removing downloaded skills is not supported yet. Pass --mcp to remove "
                 "schemas from the skills MCP connection."
             )
-        remove_skills_command()
+        requested_agents = (
+            None
+            if agents is None
+            else ({agent.strip().lower() for agent in agents.split(",") if agent.strip()} or None)
+        )
+        remove_skills_command(agents=requested_agents)
     except RuntimeError as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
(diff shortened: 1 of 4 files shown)


src/ucode/cli.py
```diff
@@ -1370,15 +1370,33 @@ def skills_remove(
...
                 "Removing downloaded skills is not supported yet. Pass --mcp to remove "
                 "schemas from the skills MCP connection."
             )
-        remove_skills_command()
+        requested_agents = (
+            None
+            if agents is None
+            else ({agent.strip().lower() for agent in agents.split(",") if agent.strip()} or None)
+        )
+        remove_skills_command(agents=requested_agents)
```
sunishsheth2009, line 1429: **P2 (consistency, pre-existing — worth fixing in both `skill` and `mcp`):** `--agents` here only lowercases the names and then `setup_mcp_clients` does a membership check, so `--agents claude-code` errors with "not configured". But the *add* paths (`skill add`, `mcp add`) run names through `normalize_tool`, so aliases like `claude-code` / `gemini-cli` resolve there. `mcp remove` has the identical gap, so skill remove is at least consistent with its sibling — but the add-normalizes / remove-doesn't asymmetry can trip a user who added with an alias and then can't remove with it. Not blocking this stack; worth normalizing on the remove paths too.

xsh310: Good catch. Fixed in #563 (stacked on this PR): `skill remove` and `mcp remove` now normalize `--agents` through a shared `normalize_agents` helper, so aliases like `claude-code` / `gemini-cli` resolve on removal just like on add. The pre-existing `mcp remove` gap is covered too.

sunishsheth2009 (approved):

PR #530, opened 2026-09-08
fix smart routing detection for claude code 

lilly-luo:
before the below command did not trigger smart routing. this is a problem because it's how isaac launches cc. 

```
ENABLE_SMART_ROUTING_V2=1 ucode claude --session-id 123
``` 

this PR fixes so that unless --model is specified, we launch smart routing for cc  

- [x] `ENABLE_SMART_ROUTING_V2=1 uv run ucode claude --session-id 123` -> smart routing enabled
- [x] `ENABLE_SMART_ROUTING_V2=1 uv run ucode claude update` -> no smart routing or tui. just works ✅ 
- [x] `ENABLE_SMART_ROUTING_V2=1 uv run ucode claude --session-id 123 --model system.ai.glm-5-2` -> no smart routing 
- [x] `ENABLE_SMART_ROUTING_V2=1 uv run ucode claude --model system.ai.glm-5-2` -> no smart routing


src/ucode/cli.py
```diff
@@ -1824,6 +1825,31 @@ def _download_managed_skills(managed: dict, state: dict) -> None:
...
+
+    Values belonging to arbitrary Claude options cannot be identified without duplicating
+    Claude's option parser, so the first forwarded argument determines whether the invocation
+    begins with an option or a command.
+    """
+    if model is not None or has_explicit_model_arg(tool_args):
+        return False
+    if not tool_args or explicit_prompt:
+        return True
+    return tool == "claude" and tool_args[0].startswith("-")
```
lilly-luo, line 1839: this is the main change. it says: after ucode consumes all its relevant args (like --workspace), if the first arguments starts with -, then invoke smart routing. 

the problem before was: 
`ENABLE_SMART_ROUTING_V2=1 ucode claude --verbose` and `ENABLE_SMART_ROUTING_V2=1 ucode claude --session-id xxx` wasn't invoking smart routing. but it should.

david-siqi-liu (approved):

PR #539, opened 2026-09-09
fix typer._click module does not exist 

lilly-luo:
problem: typer added the _click module in 0.26.0. if <0.25 is cached, then _click will fail. 

the fix is: 
- remove the use of the _click module (wasn't needed...)
- pin typer to >=0.26 so that this doesnt recur in the future 

repro to confirm that 0.26 is the right version: 
```
$ uv run --with "typer==0.25.1" python -c \
  'import typer; print(typer.__version__); from typer._click import Context as ClickContext; print(ClickContext)'
warning: `VIRTUAL_ENV=.venv` does not match the project environment path `/home/lilly.luo/ucode/.venv` and will be ignored; use `--active` to target the active environment instead
Installed 8 packages in 8ms
0.25.1
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'typer._click'
``` 

``` 
$ uv run --with "typer==0.26.0" python -c \
  'import typer; print(typer.__version__); from typer._click import Context as ClickContext; print(ClickContext)'
warning: `VIRTUAL_ENV=.venv` does not match the project environment path `/home/lilly.luo/ucode/.venv` and will be ignored; use `--active` to target the active environment instead
Installed 7 packages in 9ms
0.26.0
<class 'typer._click.core.Context'>
```

### DIFF
--- a/src/ucode/cli.py
+++ b/src/ucode/cli.py
@@ -9,11 +9,10 @@ import subprocess
 from collections.abc import Iterator
 from contextlib import contextmanager
 from importlib import metadata
-from typing import Annotated
+from typing import Annotated, Any
 
 import typer
 from rich.panel import Panel
-from typer._click import Context as ClickContext
 from typer.core import TyperCommand
 
 from ucode.agents import (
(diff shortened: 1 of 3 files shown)


src/ucode/cli.py
```diff
@@ -9,11 +9,10 @@
 from collections.abc import Iterator
 from contextlib import contextmanager
 from importlib import metadata
-from typing import Annotated
+from typing import Annotated, Any
 
 import typer
 from rich.panel import Panel
-from typer._click import Context as ClickContext
```
lilly-luo, line 16: main fix

rohita5l (approved):

PR #540, opened 2026-09-09
[AIGTWY-4564] Add Claude parent schema discovery

andy-xu-db:
## Summary
- add `--parent <catalog>.<schema>` to Claude launches
- enable model discovery and scope it to the selected schema
- keep model provider routing authoritative

## Testing
- focused Claude parent-schema suites (526 passed)
- mixed Bedrock Claude/GPT target test
- Ruff check and format check
- LiteSwap: Claude `/model` matched all 14 models under `main.andy`

## e2e test with LiteSwap
Claude model picker


src/ucode/agents/claude.py
```diff
@@ -358,6 +360,8 @@ def render_overlay(
     ]
     if provider:
         header_lines.append(f"Databricks-Model-Provider-Service: {provider}")
+    elif parent_schema:
```
lilly-luo: add a comment saying why you can't have both a provider service head and a model service parent schema

src/ucode/cli.py
```diff
@@ -1965,10 +1965,18 @@ def _launch_tool(
...
 ) -> None:
     try:
         tool = normalize_tool(tool_name)
+        if parent_schema is not None:
+            parts = parent_schema.split(".")
+            if len(parts) != 2 or any(
+                not part or any(ch in " /" or ord(ch) < 32 or ord(ch) == 127 for ch in part)
+                for part in parts
+            ):
+                raise RuntimeError("--parent must be `<catalog>.<schema>`.")
```
lilly-luo: this looks....concerning. can you see if ucode already has some validation like this? if no, can you put it into a string utils file and add unit tests for that util file only

lilly-luo:
can u test with a bedrock provider that has claude and gpt models as well?

src/ucode/agents/claude.py
```diff
@@ -167,6 +167,7 @@ def _resolve_web_search_model(state: dict) -> str | None:
         "x-databricks-use-coding-agent-mode",
         "user-agent",
         "databricks-model-provider-service",
+        "databricks-model-service-parent-schema",
```
lilly-luo: looks like theres a const in the codex PR? can we share that across both places and put into a const file? https://github.com/databricks/unity-gateway/pull/541/changes/f914af76d0f6319127b2cbdbd5564622abbeb40e..f456c19c08ae51cf1493d8ac763246897258aff5#diff-6dfc964820cd6e8cfb07173a500b45493cf33dbb9d62bd6c6a1439d5d89734b8R62

src/ucode/cli.py
andy-xu-db: moved to string utils file

src/ucode/agents/claude.py
andy-xu-db: added comment

src/ucode/agents/claude.py
andy-xu-db: added as constants in const file

src/ucode/cli.py
```diff
@@ -2063,6 +2067,7 @@ def _launch_tool(
                 )
             if managed_provider:
                 provider = managed_provider
+        effective_parent_schema = None if provider else parent_schema
```
lilly-luo: should we raise an error if someone tries to pass a parent_schema and a provider at the same time?

src/ucode/cli.py
```diff
@@ -2500,6 +2506,13 @@ def claude_cmd(
             "before any `--` separator.",
         ),
     ] = None,
+    parent: Annotated[
+        str | None,
+        typer.Option(
+            "--parent",
+            help="Discover model services in `<catalog>.<schema>`.",
```
lilly-luo: can you put an ex in the help text? `Discover model services in `<catalog>.<schema>`. Example: main.default`

src/ucode/constants.py
```diff
@@ -2,3 +2,6 @@
 
 LOCALHOST = "localhost"
 LOOPBACK_HOST = "127.0.0.1"
+
+MODEL_PROVIDER_SERVICE_HEADER = "databricks-model-provider-service"
+MODEL_SERVICE_PARENT_SCHEMA_HEADER = "databricks-model-service-parent-schema"
```
lilly-luo: love!

src/ucode/model_service_headers.py
```diff
@@ -0,0 +1,17 @@
...
+def model_service_routing_headers(
+    provider: str | None,
+    parent_schema: str | None,
+) -> dict[str, str]:
+    # A provider selects one MPS; a parent discovers Model Services.
+    if provider:
+        return {MODEL_PROVIDER_SERVICE_HEADER: provider}
+    if parent_schema:
+        return {MODEL_SERVICE_PARENT_SCHEMA_HEADER: parent_schema}
+    return {}
```
lilly-luo: can you throw an error or smth if both are set? bc right now if u set provider and not parent_schema, it will pick provider and parent_schema will be silently ignored 

also is a separate file + util necessary for this? seems like it can be inlined

tests/test_agent_claude.py
```diff
@@ -288,7 +288,7 @@ def test_relayed_sends_mps_header_but_not_swap_token(self):
             relayed_base_url="http://127.0.0.1:9",
         )
         headers = overlay["env"]["ANTHROPIC_CUSTOM_HEADERS"]
-        assert "Databricks-Model-Provider-Service: c.s.mps" in headers
+        assert "databricks-model-provider-service: c.s.mps" in headers
```
lilly-luo: why is this being lowercased

lilly-luo (approved):

PR #545, opened 2026-09-09
[AIGTWY-4573] Consume the coding-agent config + version-gate the launch apply

david-siqi-liu:
Stacked PR: #547 (`david/AIGTWY-4573-model-picker`) applies the model picker on top of this.

Teaches the `ug` client to read the `CodingAgentConfig` the AI Gateway will emit (AIGTWY-4572, estore v4 #2580860) without regressing today's path, and stops a launch from re-applying the config (and re-prompting for sudo) when nothing changed.

- `normalize_managed_config` reads the config shape. `enabled_agents` is a repeated list of `{agent enum, config}` (verified against the #2580860 proto, since proto map keys can't be enums), and the inner `AgentConfig` carries the model source (`models`, with `model_provider_service` / `unity_catalog_location` / `model_services`), `default_models`, and `http_headers`. The current fields win over the deprecated predecessors the proto still retains: `http_headers` over `custom_headers`, and the `models` / `default_models` set over the `model_config` oneof, so a config carrying only the old fields still normalizes.
- `spec_version` forward-compat gate: this build supports up to `MAX_SPEC_VERSION` (currently 1). A newer or malformed `spec_version` is refused as an unresolved read, so the launch keeps its last-known-good cache and the developer is told to `ug upgrade`.
- `UCODE_MANAGED_CONFIG_STUB` reads a local JSON config so the client can be exercised before the server emits one.
- Version-gate the launch apply. `refresh_managed_config` still fetches on every launch, but re-applying (the only step that can prompt for a sudo password, because it rewrites the OS-managed files) now happens only when the fetched config is newer than the last one applied here. `managed_config_is_newer` compares the config's update-time watermark against the last-applied watermark (`get_applied_managed_update_time` / `set_applied_managed_update_time`): an unchanged launch skips the re-apply and never prompts, while `ug configure` and `ug <agent> --refresh` always re-apply. A failed fetch falls back to the last persisted config, a `FEATURE_DISABLED` response drops the cache, and a workspace with no config persists an empty marker so a removed policy doesn't return after a transient outage.
- Stop `ug configure` from validating agents with a probe message. Remove the post-configure agent validation (`validate_all_tools` / `validate_tool`) and the `--skip-validate` flag: it isn't configure's job to test models, the probe also reverted a freshly-written config on failure (for example a gateway-rejected `web_search` tool), and the gateway enforces model access regardless. The now-unused `validate_*` helpers are left for a follow-up cleanup.
- Drop the bare-`ug` managed-launch banner (the "Using managed config" line). It printed before the version-gate re-apply and its sudo prompt, reading as if the agent were launching before the config that selects it was settled. The launch panel (`Launching <agent> with Unity Gateway`) plus the recommended-agent note are the announcement now, so bare `ug` matches `ug <agent>`. Also drop the redundant post-panel "Starting <agent>" line.
- Rename the user-facing product string "CLI Managed Configuration" to "managed configuration" (lowercase), tighten the budget-tier recommendation line, and remove the sample `examples/managed-config.stub.json` (the `UCODE_MANAGED_CONFIG_STUB` env var stays).

The static `models.names` allow-list and `model_service_location` are parsed here but not yet applied to each agent's `/model` picker; that lands in #547.


examples/managed-config.stub.json
```diff
@@ -0,0 +1,53 @@
...
+  ],
+  "mcp_servers": {
+    "names": ["system.ai.github"],
+    "tags": ["MCP_SERVER_TYPE_UC_SERVICE"]
+  },
+  "skills": {
+    "names": ["system.ai.pdf-extraction"],
+    "tags": []
+  },
+  "tracing": { "enabled": true },
```
tt-le: should this live inside the agent specific config?

david-siqi-liu:
Superseded by the finer-grained AIGTWY-4573 stack: this PR's content is now split into #590 (parse the v2 config) and #593 (apply at launch + version-gate). See #590 for the stack map.

PR #547, opened 2026-09-09
[AIGTWY-4573] Apply static/auto model curation to each agent's /model picker

david-siqi-liu:
Stacked on #545 (branch `david/AIGTWY-4573`), merge that first. This is the second half of AIGTWY-4573: now that the client parses the config (including the static `models.names` list and `model_service_location`), turn each agent's model source into its own `/model` picker.

- `managed_resolve`: expose `managed_static_models` + `managed_model_service_location`, count them in `managed_supplies_models` so ucode skips its own discovery, and layer them into state for the Claude/Codex writers.
- Claude: a static list becomes `availableModels` + `enforceAvailableModels` + a `modelPicker` that replaces the built-in options, so `/model` shows exactly the named ids (including non-Claude gateway models, which discovery alone omits); a discovery location instead turns on gateway model discovery. Both are skipped under a provider or relayed launch, and stale picker keys are pruned when the config drops its list.
- Codex: a static list is written as a `model_catalog_json` catalog of full `ModelPreset`s (Codex's StaticModelsManager replaces remote discovery); auto needs no catalog, since the provider base_url already drives Codex's discovery. The catalog file is removed when the list goes away.

Verified: `render_overlay` writes the expected Claude keys, and the generated Codex catalog is accepted by the real Codex 0.146.1 (`codex debug models` renders exactly the listed models with `visibility: "list"`). The pickers are soft curation (they do not fence an explicit `--model`); hard governance stays gateway-side. The Codex preset's runtime-behavior fields (base_instructions, truncation_policy, etc.) are generic defaults, sufficient for the `/model` listing this PR targets; per-model behavior tuning is a follow-up.


david-siqi-liu:
Superseded by the finer-grained AIGTWY-4573 stack: this PR's content is now split into #591 (shared managed_resolve helpers), #592 (Claude picker), and #594 (Codex catalog). See #590 for the stack map.

PR #548, opened 2026-09-10
Drive MCP connection login (RFC 8707 resource) from ucode mcp-proxy

sunishsheth2009:

**Change:** Teach `ucode mcp-proxy` to complete an MCP **connection login** when the AI Gateway asks for one.

**Why:** When an AI Gateway MCP service is backed by a per-user connection (e.g. `system.ai.github`) and the user hasn't logged in to that connection yet, the gateway answers a tools call with an RFC 9728 challenge — HTTP 401 + `WWW-Authenticate: Bearer resource_metadata="…"`. A valid Databricks *workspace* token isn't enough; the *connection* needs its own login. Today the proxy just forwards that response to the coding agent, which shows `connected · tools fetch failed` and the user is stuck with no way to log in.

The proxy's `httpx` auth now detects that challenge and drives an interactive **resource-scoped** OAuth login: authorization-code + PKCE against the workspace `/oidc`, using the published `databricks-cli` app, sending the MCP endpoint URL as the RFC 8707 `resource` indicator. That indicator makes `/oidc` route the browser through the connection's `/mcp-service-login` page → the user signs in to the backing SaaS → the connection credential is stored server-side → the flow returns → the proxy retries the original request, which now succeeds. No coding-agent OAuth (RFC 9728 / dynamic client registration) support required.

## How do you know it works?

**Testing:** `uv run pytest tests/test_resource_login.py tests/test_mcp_proxy.py` — 35 passing. New tests cover the authorize-URL + PKCE construction and the `401`/`resource_metadata` challenge detection; existing proxy tests still pass. `ruff check`/`format` clean. The interactive browser + token-exchange path is validated end-to-end against staging separately (it requires the gateway-side RFC 9728 challenge to be enabled for the connection).

**Notes:**
- Interactive login is skipped under `DATABRICKS_BEARER` (headless/CI) — no browser to drive.
- One connection-login attempt per proxy lifetime, guarding against a retry loop.
- `resource_login.py` is stdlib-only (no new dependency).

### DIFF
--- a/src/ucode/mcp_proxy.py
+++ b/src/ucode/mcp_proxy.py
@@ -33,16 +33,20 @@ especially confusing -- the user needs to be told to re-run
 
 from __future__ import annotations
 
+import functools
+import os
 import sys
 from collections.abc import AsyncIterator
 from types import ModuleType, TracebackType
 from typing import Protocol, Self
 
 import anyio
+from anyio import to_thread
 from mcp.client.streamable_http import streamable_http_client
 from mcp.server.stdio import stdio_server
 
 from ucode.databricks import ensure_pat_bearer, get_databricks_token
+from ucode.resource_login import ConnectionLoginError, login_for_connection
 
 # Exit code used when the proxy cannot continue. MCP clients surface a non-zero
 # exit far more usefully than a timeout, so bail out instead of hanging.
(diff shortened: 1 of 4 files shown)


sunishsheth2009:
Closing: folded into the simpler setup-time login approach. Rather than driving the login from inside the proxy, ucode will run `databricks auth login --resource <mcp-url>` once at `ug mcp add` time (needs CLI flag databricks/cli#6621). Tracked in the reworked #557.

PR #557, opened 2026-09-10
mcp-proxy: drive per-connection login on a 401 (generic, all agents)

sunishsheth2009:
> Dependency **merged**: `databricks auth login --resource` shipped in databricks/cli#6763, so the connection sign-in works with a current Databricks CLI. Rebased on `main`.

## What

Give **every** coding agent (codex, cursor, gemini, claude, …) a working login flow for connection-backed AI Gateway `mcp-services` endpoints (e.g. `system.ai.github`) — implemented in the `ug mcp-proxy` stdio bridge every agent already spawns. No new library (this is what a generic OAuth MCP bridge like [`mcp-remote`](https://www.npmjs.com/package/mcp-remote) does, done in ucode with the Databricks CLI), and no per-agent OAuth app.

## How

A connection-backed `mcp-services` endpoint needs a per-user connection credential before its tools can be used. The proxy drives that login **at connect time**, before it opens the stdio↔HTTP bridge:

1. `serve()` detects a connection-backed URL and runs a **blocking** `databricks auth login --resource <mcp-url>`. The RFC 8707 `resource` indicator makes a resource-aware `/oidc` route the browser through the connection's own SaaS login (`/mcp-service-login`) before minting the token. It uses the CLI's **own** default client + registered loopback redirect — **no `--client-id`, no auth-side redirect change**.
2. Once the login completes, the credential exists server-side, so the bridge opens already authenticated and AI Gateway is only ever called with a valid credential (it never has to elicit a login). The login is idempotent — once signed in it returns immediately.

This is the `mcp-remote` pattern: the CLI opens the browser natively (macOS/Windows/Linux-desktop) and, when it can't (a headless/remote box), prints the authorize URL to **stderr** so the user can open it manually. The proxy never writes to stdout (the MCP JSON-RPC wire).

## Why this shape (supersedes the earlier bridges in this PR's history)

Earlier iterations tried a proxy that faked a `sign_in` tool / rewrote `tools/list`, a login-on-401 httpx auth-hook retry, and a launch-time check — each either more complex or subtly wrong (the auth hook ran a blocking subprocess inside the event loop; launch-time fired a login per connection at every launch). Driving the login once at connect, and letting the agent block on "connecting…" while it runs, is the minimal generic realization and matches how `mcp-remote` behaves.

## Dependencies

- **CLI `--resource`**: databricks/cli#6763. Until released, `databricks auth login` ignores `--resource`.
- **`/oidc` resource-indicator handling (login §2)** + **`/mcp-service-login` return_to (§3)** — the universe side of this flow. AI Gateway §1 (the 401) is already on staging.

## Scope

- Connection-backed mcp-services endpoints get the connect-time login. PAT profiles skip it (no connection OAuth to drive).
- Plain Databricks MCP services keep the existing transparent token-injection (no extra login).

## Tests

`test_mcp_connection_login` — connection-FQN parsing, and the CLI login runner with the subprocess mocked (sends `--resource`/`--host` and no `--client-id`; routes output to stderr; reports nonzero-exit, timeout, and missing-binary). `test_mcp_proxy` — connect-time login ordering (login → bridge), non-connection URLs skip it, a login failure exits before the bridge, and `--use-pat` skips it. `uv run pytest tests/test_mcp*.py` green; ruff + ty clean.

## Example output

`ug mcp-proxy` is launched by the agent (e.g. Codex) as a stdio subprocess. On a connection-backed MCP service whose per-user credential is missing, it drives the one-time sign-in on stderr before serving tools (stdout stays the clean JSON-RPC stream), so the tools list appears already-authenticated instead of failing:

```
ucode mcp-proxy: 'system.ai.github' needs a one-time connection sign-in. Opening your
browser to complete it — if it doesn't open, use the authorization URL printed below.
https://eng-ml-inference.staging.cloud.databricks.com/oidc/v1/authorize?...&resource=...%2Fmcp-services%2Fsystem.ai.github
ucode mcp-proxy: signed in.
```

If the Databricks CLI is too old to support `--resource`, it fails fast with a clear message instead of a cryptic connection error:

```
ucode mcp-proxy: 'system.ai.github' connection sign-in can't run. Upgrade the CLI
(needs databricks/cli#6763 for `databricks auth login --resource`).
```

### DIFF
--- a/src/ucode/mcp_connection_login.py
+++ b/src/ucode/mcp_connection_login.py
@@ -19,6 +19,7 @@ needed. Requires the CLI ``--resource`` flag (databricks/cli#6621).
 from __future__ import annotations
 
 import subprocess
+import sys
 
 # AI Gateway MCP service endpoints look like
 # ``https://<ws>/ai-gateway/mcp-services/<catalog>.<schema>.<service>``.
@@ -58,8 +59,13 @@ def run_connection_login(
     it is sent as ``--resource`` so a resource-aware ``/oidc`` drives the
     connection's SaaS login before issuing the token. Uses the Databricks CLI's
     own default client, whose loopback redirect is already registered — no
-    ``--client-id`` needed. Returns ``(ok, message)``; ``message`` is the CLI's
-    own output on failure so the caller can surface it.
+    ``--client-id`` needed.
+
+    The CLI opens the browser to complete the login and prints the authorize URL.
+    We route its output to **stderr** (never stdout — that is the proxy's MCP
+    JSON-RPC wire), so a coding agent surfaces it in the server's log and the URL
+    stays visible when the browser can't open (e.g. a headless remote). Returns
+    ``(ok, message)``; on failure ``message`` points at that log.
     """
     argv = [
         login_binary,
@@ -72,22 +78,35 @@ def run_connection_login(
     ]
     if profile:
         argv += ["--profile", profile]
+    connection = connection_from_url(resource_url) or resource_url
+    print(
+        f"ucode mcp-proxy: '{connection}' needs a one-time connection sign-in. Opening your "
+        "browser to complete it — if it doesn't open, use the authorization URL printed below.",
+        file=sys.stderr,
+        flush=True,
+    )
     try:
+        # stdout -> stderr: the CLI's prompts and authorize URL reach the agent's
+        # MCP log (fd 2) without corrupting this process's stdout (fd 1, the MCP
+        # JSON-RPC stream). stdin is closed since the flow is browser-driven.
         result = subprocess.run(
             argv,
             check=False,
-            capture_output=True,
-            text=True,
             timeout=_LOGIN_TIMEOUT_SECONDS,
+            stdin=subprocess.DEVNULL,
+            stdout=sys.stderr,
+            stderr=sys.stderr,
         )
     except OSError as exc:
         return False, f"could not run '{login_binary} auth login': {exc}"
     except subprocess.TimeoutExpired:
-        return False, "login timed out waiting for the browser flow to complete"
+        return False, "connection sign-in timed out waiting for the browser flow to complete"
     if result.returncode == 0:
         return True, "signed in"
-    detail = (result.stderr or result.stdout or "").strip()
-    return False, detail or f"login exited with code {result.returncode}"
+    return (
+        False,
+        f"connection sign-in did not complete (CLI exited {result.returncode}; see the log above)",
+    )
 
 
 __all__ = [
--- a/src/ucode/mcp_connection_login.py
+++ b/src/ucode/mcp_connection_login.py
@@ -1,0 +1,97 @@
+"""Per-connection login for AI Gateway MCP services, driven by the proxy on a 401.
+
+A connection-backed AI Gateway MCP service (e.g. ``system.ai.github``) needs a
+per-user connection credential before its tools can be used. Until the user has
+logged in to the underlying SaaS, AI Gateway answers requests with an HTTP 401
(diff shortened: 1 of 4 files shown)


src/ucode/mcp_connection_login.py
```diff
@@ -0,0 +1,140 @@
...
+    ``--client-id`` needed.
+
+    The CLI opens the browser to complete the login and prints the authorize URL.
+    We route its output to **stderr** (never stdout — that is the proxy's MCP
+    JSON-RPC wire), so a coding agent surfaces it in the server's log and the URL
+    stays visible when the browser can't open (e.g. a headless remote). Returns
+    ``(ok, message)``; on failure ``message`` points at that log.
+    """
+    connection = connection_from_url(resource_url) or resource_url
+    if not _cli_supports_resource_flag(login_binary):
```
xsh310: nit: maybe we can add a databricks CLI version auto upgrade (install_databricks_cli) when configuring the MCP if we don't have that yet?

src/ucode/mcp_proxy.py
```diff
@@ -111,29 +113,72 @@ def _fail_fast(message: str) -> None:
...
+        return connection is not None and response.status_code == 401 and not login["attempted"]
+
+    def _login_or_fail() -> None:
+        login["attempted"] = True
+        ok, detail = run_connection_login(url, workspace, profile=profile)
+        if not ok:
+            raise ProxyAuthError(f"connection login for '{connection}' failed: {detail}")
 
     class _DatabricksTokenAuth(httpx.Auth):
         def auth_flow(self, request):
```
xsh310: Do we still need this sync one?

xsh310 (approved):

xsh310:
Might be helpful to add a video recording of this

src/ucode/mcp_proxy.py
```diff
@@ -111,29 +113,72 @@ def _fail_fast(message: str) -> None:
...
+            raise ProxyAuthError(str(exc)) from exc
+        request.headers["Authorization"] = f"Bearer {token}"
+
+    def _needs_connection_login(response) -> bool:
+        # A 401 from a connection-backed service means the per-user credential is
+        # missing. Only drive the login once per session — if it doesn't resolve
+        # the 401, re-running it on every subsequent request would loop browsers.
+        return connection is not None and response.status_code == 401 and not login["attempted"]
+
+    def _login_or_fail() -> None:
```
xsh310: nit: would _connection_login_or_fail() be clearer?

sunishsheth2009: Done — renamed to `_connection_login_or_fail` (commit d99ae4b).

src/ucode/mcp_proxy.py
sunishsheth2009: The proxy only ever uses httpx's `AsyncClient`, so `async_auth_flow` is the real runtime path. I kept the sync `auth_flow` because (a) httpx.Auth is defined for both client types and dropping it would make a sync client silently skip bearer injection (base `auth_flow` yields unchanged), and (b) the unit tests drive the generator synchronously. Happy to drop it and convert those tests to async-driven if you'd prefer async-only — let me know.

src/ucode/mcp_connection_login.py
sunishsheth2009: Nice idea. Keeping it out of this PR since auto-upgrading the user's Databricks CLI is a behavior change worth its own review; today an old CLI fails fast with an explicit 'upgrade the CLI (databricks/cli#6621)' message rather than a cryptic error. I'll track it as a follow-up alongside the other MCP cleanups.

PR #560, opened 2026-09-11
ug mcp add: native OAuth (direct HTTP) for Claude Code + Cursor, else generic proxy

sunishsheth2009:
## What

Give **Claude Code and Cursor** their *native* connection-login experience for connection-backed AI Gateway `mcp-services` endpoints (e.g. `system.ai.github`): the agent itself is the OAuth client and drives the `/oidc` → `/mcp-service-login` login, so `/mcp` shows a real "Authenticate" (Claude) / login (Cursor) prompt instead of a token silently injected by the proxy. This is a much better UX than the generic stdio-proxy path for the agents that can do it.

It is **probe-driven with a safe fallback**: only used when the agent's OAuth app is actually published on the workspace; otherwise (and for every other agent) it falls back to the generic `ug mcp-proxy` stdio path (databricks/ucode#557).

## How

`/oidc` has **no** dynamic client registration, so the agent must present a **pre-registered public client** whose loopback `/callback` redirect is registered:
- **Claude Code** → `claude mcp add --transport http --client-id claude-code` (Claude drives MCP OAuth with the RFC 8707 `resource` indicator).
- **Cursor** → a `url` + `auth: {CLIENT_ID: cursor-desktop}` entry in `~/.cursor/mcp.json` (Cursor supports a pre-registered client; its fixed `http://localhost:8787/callback` redirect matches by path, RFC 8252 §8.4).

`mcp_oauth.oauth_client_available(workspace, client_id)` back-channel-probes `/oidc/v1/token` (401 `invalid_client` = absent, else present) and caches the answer per workspace (weekly TTL). `configure_client_mcp_server` registers direct-HTTP only when: the URL is a connection-backed `mcp-services` endpoint, the agent has a mapped OAuth client (`AGENT_OAUTH_CLIENT`), that client is published, and it isn't a `--use-pat` / skills-registry registration. Everything else keeps the stdio proxy.

## Relationship to the generic proxy (#557)

Complementary. #557 gives **every** agent a working login via the proxy (connect-time `databricks auth login --resource`). This PR is a UX upgrade layered on top: where an agent *can* be the OAuth client and its app exists, we let it drive the login natively. Same server-side flow underneath (`/oidc` → `/mcp-service-login`), same §2/§3 dependency.

## Dependencies

Same server side as #557: AI Gateway §1 (`WWW-Authenticate` 401, already on staging) + login §2 (`/oidc` resource-indicator redirect) + webapp §3 (`/mcp-service-login` `return_to`). No CLI dependency for the native path (the agent is the OAuth client, not the CLI).

## Scope / fallback

- Connection-backed `mcp-services` + Claude/Cursor + published app → native direct-HTTP.
- No published app, other agents (codex/gemini/copilot/opencode), PAT, non-connection MCPs, skills registry → generic stdio proxy (unchanged).

## Tests

`test_mcp_oauth` (probe: present/absent/network-fail, cache TTL). `test_mcp.TestConfigureClientMcpServer` (Claude + Cursor each: direct-HTTP when the app is available, proxy fallback when not, proxy under `--use-pat`, proxy for non-connection URLs; an OAuth-less agent proxies and never probes). `test_agent_cursor` (the `url`+`auth` entry shape and merge). `uv run pytest` green (pre-existing unrelated failures in `test_codex_smart_routing_v2` / `test_e2e_user_agent` confirmed on `main`); ruff + ty clean.


src/ucode/mcp_oauth.py
```diff
@@ -0,0 +1,127 @@
...
+        method="POST",
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+    )
+    try:
+        urllib.request.urlopen(request, timeout=10)  # noqa: S310 - fixed https workspace URL
+        return True  # a 2xx for a dummy code is unexpected, but means the client is valid
+    except urllib.error.HTTPError as exc:
+        # 401 invalid_client => not registered; any other error (400 for the bad
+        # code) => the client IS registered.
+        return exc.code != 401
```
xsh310: **The probe fails open, and the answer is cached for a week.** `exc.code != 401` treats any status other than 401 as "client is published." That includes transient statuses like 429 or 5xx (rate limiting, an incident) and 404 or redirects on older workspaces. When one of those happens we register direct HTTP and then cache that decision for 7 days, so the user gets a broken "Authenticate" that fails with `invalid_client`, in exactly the case where the proxy fallback would have just worked.

Since the proxy is always safe, the probe should stay on "absent" unless it sees a clear positive. A result with three states would be cleaner:

* `400` (invalid_request / invalid_grant): present
* `401`: absent
* anything else: unknown, so return absent and skip caching it (or cache with a short TTL)

The `OSError` branch just below has the same "cache a transient failure for a week" concern: it returns `False` correctly, but `oauth_client_available` then persists that `False` for 7 days, so one network blip keeps the user on the proxy for a week. Folding network errors and unexpected statuses into one "unknown, do not persist" bucket fixes both.

src/ucode/mcp_oauth.py
```diff
@@ -0,0 +1,127 @@
...
+def _read_cache() -> dict:
+    try:
+        return json.loads(_CACHE_PATH.read_text(encoding="utf-8"))
+    except (OSError, ValueError):
+        return {}
+
+
+def _write_cache(cache: dict) -> None:
+    try:
+        _CACHE_PATH.write_text(json.dumps(cache, indent=2), encoding="utf-8")
```
xsh310: **`_write_cache` doesn't create `~/.ucode` before writing.** Every other writer under `APP_DIR` ensures the directory exists first (`state.save_state`, `config_io.backup_existing_file`, and `write_json_file` via `ensure_parent_dir`). This one doesn't, so on a fresh machine where `~/.ucode` doesn't exist yet, `write_text` raises `FileNotFoundError`, the `except OSError` swallows it, and the cache never persists. `ug mcp add` then probes the network again (a 10s timeout call) on every run until some other code path happens to create the directory.

Small fix: call `ensure_parent_dir(_CACHE_PATH)` (or `APP_DIR.mkdir(parents=True, exist_ok=True)`) before the write.

src/ucode/mcp_oauth.py
```diff
@@ -0,0 +1,127 @@
...
+            "grant_type": "authorization_code",
+            "code": "ucode-probe-not-a-real-code",
+            "redirect_uri": f"http://localhost:{MCP_OAUTH_CALLBACK_PORT}/callback",
+            "client_id": client_id,
+            # `/oidc` rejects a code_verifier shorter than 43 chars *before* it
+            # validates the client, so pad past that to reach the client check.
+            "code_verifier": "u" * 43,
+        }
+    ).encode("ascii")
+    request = urllib.request.Request(
```
xsh310, line 72: Wondering whether we can reuse _http_post_json in databricks.py?

src/ucode/mcp.py
```diff
@@ -151,6 +179,46 @@ def add_claude_mcp_server(
         raise RuntimeError(f"Failed to add MCP server '{name}' via claude CLI.") from exc
 
 
+def add_claude_http_mcp_server(
```
xsh310, line 183: nit: given that the cursor logic is added to cursor.py, should we move this claude specific logic to claude.py as well?

src/ucode/mcp_oauth.py
```diff
@@ -0,0 +1,127 @@
...
+        return exc.code != 401
+    except OSError:
+        # Network failure: don't claim availability — the caller falls back to the
+        # stdio proxy, which is always safe.
+        return False
+
+
+def _read_cache() -> dict:
+    try:
+        return json.loads(_CACHE_PATH.read_text(encoding="utf-8"))
```
xsh310: `json.loads` can return valid JSON that isn't a dict (an empty list, or a hand edited value), and then `oauth_client_available` runs `cache.get(ws, {}).get(...)` and crashes `ug mcp add` with `AttributeError`. `config_io.read_json_safe` already handles exactly this (returns `{}` for any non-dict), so reuse it here.

xsh310 (approved):
Overall looks good, if someone with more auth context can sanity check that'd be best

src/ucode/mcp_oauth.py
sunishsheth2009: Fixed in 00b17f7. `_probe_oauth_client` now returns `True`/`False`/`None`: **400** → registered, **401** → not registered, and **anything else** (429/5xx, 404/redirect, a network error, or an unexpected 2xx) → `None` (inconclusive). `oauth_client_available` treats `None` as "fall back to the stdio proxy" and does **not** cache it — so a rate-limit/incident can't flip a workspace onto the direct-HTTP path, and a transient failure isn't sticky for the week-long TTL (we re-probe next run). Added tests for the transient/404/unexpected-2xx/network cases and the no-cache behavior.

src/ucode/mcp_oauth.py
sunishsheth2009: Fixed in 00b17f7. Cache I/O now goes through `config_io.read_json_safe` / `write_json_file`, and `write_json_file` ensures `APP_DIR` exists before writing (via `ensure_parent_dir`) — consistent with the other `APP_DIR` writers, and fixes the write on a fresh `~/.ucode`.

src/ucode/mcp_oauth.py
sunishsheth2009: Good thought — I looked at reusing it, but `_http_post_json` doesn't fit this probe: it takes an auth `token` (the probe is deliberately **unauthenticated**), sends a **JSON** body (the `/oidc/v1/token` endpoint requires `application/x-www-form-urlencoded`), and returns `(payload, reason)` without surfacing the **HTTP status code** — whereas the probe's whole job is to distinguish 400 (registered) vs 401 (absent) vs everything-else (inconclusive). So I kept the small self-contained urllib call. Happy to extract a shared helper if you'd prefer, but it'd need to expose the raw status and support unauth/form-encoded bodies.

src/ucode/mcp.py
sunishsheth2009: Fair point on symmetry. The wrinkle: `add_claude_http_mcp_server` sits next to its siblings `add_claude_mcp_server` and `remove_claude_mcp_server` in `mcp.py` (all called by `configure_client_mcp_server`), whereas cursor's config lives in `cursor.py` because cursor is MCP-only and owns its own `mcp.json`. Moving just the HTTP helper would split claude's registration across two files. I'd rather move all three claude helpers to `claude.py` together as a small dedicated follow-up (with the matching test monkeypatch updates) than do a partial move in this PR — happy to do that next if you agree, or keep them centralized. LMK your preference.

sunishsheth2009: Follow-up after digging in: `agents/claude.py` already imports `add_claude_mcp_server` / `remove_claude_mcp_server` **from `mcp.py`** (for the web_search MCP), so `mcp.py` is the owner of the claude registration helpers and `claude.py` is a consumer. Moving them into `claude.py` would invert that and create a circular import (mcp.py's dispatch would import from claude.py, which imports back from mcp.py), plus churn ~15 test references. Given it's a nit, I'll keep them centralized in `mcp.py` for now — a proper consolidation (untangling that dependency) is better as its own change. Resolving; happy to revisit in a dedicated cleanup.

src/ucode/mcp_oauth.py
sunishsheth2009: Fixed in 00b17f7 — `_read_cache` now uses `config_io.read_json_safe`, which returns `{}` for any non-dict / missing / invalid content, so `oauth_client_available`'s `cache.get(...)` can't hit an AttributeError. (Same commit switched the write side to `write_json_file`, which also ensures `APP_DIR` exists.)

PR #565, opened 2026-09-11
Route Databricks-hosted models to gateway auth in relayed sessions (hybrid OSS + relayed)

masonc08:
## What

Lets a **relayed** (subscription-relay) Claude Code session *also* reach Databricks-hosted (`system.ai` / OSS) models in the same session. The loopback proxy now routes per request by the requested model.

A relayed session authenticates with the caller's subscription OAuth in `Authorization`; Databricks-hosted models authenticate with a Databricks token in `Authorization`. Claude Code sends only one `Authorization` per session, so today a session is one or the other. This makes a single session serve both.

## How

All modes already hit the same upstream (`{workspace}/ai-gateway/anthropic/`); they differ only in auth + the provider-service header. So the proxy can multiplex per request:

- **Databricks-hosted model** (namespace-qualified id — `system.ai.*`, `catalog.schema.model`, `databricks-*`) → gateway auth: the Databricks token goes in `Authorization`, and the swap + `Databricks-Model-Provider-Service` headers are dropped so the gateway serves the model directly. Byte-identical to the normal non-relayed path.
- **Relayed subscription model** (bare Anthropic id — `claude-opus-4-1`, …) → today's behavior: OAuth passed through untouched, Databricks token injected as the `X-Databricks-AI-Gateway-Token` swap header.

Gated by `start_proxy(hybrid_oss_routing=...)`, enabled only for the relayed launch. A pure-relay session only sends bare Anthropic ids, so its behavior is unchanged (the classifier relays anything without a dot). Users surface Databricks-hosted models through Claude Code's existing `modelPicker` / `--model`; the proxy makes those selections route correctly.

## Testing

Screenshot 2026-09-11 at 12 09 01

**Automated (hermetic, `test_gateway_proxy.py`):**
- `TestHybridOssRouting` drives the handler end to end and asserts the forwarded headers per model: a Databricks-hosted id gets the gateway token in `Authorization` with the swap + provider-service headers dropped; a bare Anthropic id keeps the OAuth passthrough + swap header; with routing off, even a qualified id relays (pure-relay sessions unchanged).
- `is_databricks_routed_model` (qualified vs bare vs missing id), `_request_model` (body parsing), and the `forwarded_request_headers` header shapes. Full hermetic suite green.

**Live backend validation:** confirmed against a staging workspace that the gateway serves both a Databricks-hosted Claude model and an OSS model when a Messages request carries a Databricks bearer in `Authorization` and no provider-service header — the exact request shape the proxy's Databricks route emits. So both halves are covered: the per-model header logic by the unit tests, and the gateway serving those requests by the live check. The proxy's `route` diagnostic (`UCODE_RELAYED_PROXY_DIAGNOSTICS=1`) distinguishes `relay` vs `databricks` per request.

### DIFF
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -1287,7 +1287,7 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
         # the proxy re-routes those requests to gateway auth while relayed subscription
         # models keep the OAuth passthrough. Bare Anthropic ids are unaffected, so a
         # pure-relay session behaves exactly as before.
-        hybrid_oss_routing=True,
+        relayed_oss_routing=True,
     )
     # start_proxy falls back to an OS-assigned port when the cached one is taken
     # (stale proxy from a killed session). Reconcile settings + state to whatever
--- a/src/ucode/agents/claude.py
+++ b/src/ucode/agents/claude.py
@@ -1283,6 +1283,11 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
         port,
         token_header=gateway_proxy.AI_GATEWAY_TOKEN_HEADER,
         force_refresh_near_expiry=False,
+        # Let a relayed session also reach Databricks-hosted (OSS / system.ai) models:
+        # the proxy re-routes those requests to gateway auth while relayed subscription
+        # models keep the OAuth passthrough. Bare Anthropic ids are unaffected, so a
+        # pure-relay session behaves exactly as before.
+        hybrid_oss_routing=True,
     )
     # start_proxy falls back to an OS-assigned port when the cached one is taken
     # (stale proxy from a killed session). Reconcile settings + state to whatever
--- a/src/ucode/gateway_proxy.py
+++ b/src/ucode/gateway_proxy.py
@@ -7,6 +7,11 @@ header. Native gateway discovery instead carries the Databricks credential in
 `Authorization`. The proxy refreshes the applicable header and streams responses
 back verbatim.
 
+In a hybrid relayed session the proxy picks per request by the requested model:
+Databricks-hosted ids (system.ai / OSS) take the gateway-auth path while relayed
+subscription models keep the OAuth passthrough, so one Claude Code session can use
+both.
+
 Security invariants (mirroring `databricks.py` token handling):
   - Binds 127.0.0.1 only; never exposed off-host.
   - Never logs header values or bodies. The Databricks token lives in memory,
@@ -34,6 +39,9 @@ from ucode.databricks import get_databricks_token
 # client-supplied value is replaced, so a stale settings.json value can't leak.
 AI_GATEWAY_TOKEN_HEADER = "X-Databricks-AI-Gateway-Token"
 AUTHORIZATION_HEADER = "Authorization"
+# Header that routes a request to a specific Model Provider Service. Dropped when a
+# request is re-routed to a Databricks-hosted model so the gateway serves it directly.
+MODEL_PROVIDER_SERVICE_HEADER = "Databricks-Model-Provider-Service"
 # Hop-by-hop headers must not be forwarded across a proxy.
 HOP_BY_HOP_HEADERS = frozenset(
     h.lower()
@@ -186,8 +194,9 @@ def forwarded_request_headers(
     handler: BaseHTTPRequestHandler,
     token: str,
     token_header: str = AI_GATEWAY_TOKEN_HEADER,
+    extra_strip: frozenset[str] = frozenset(),
 ) -> dict[str, str]:
-    strip_on_forward = HOP_BY_HOP_HEADERS | {token_header.lower()}
+    strip_on_forward = HOP_BY_HOP_HEADERS | {token_header.lower()} | extra_strip
     headers = {
         key: value for key, value in handler.headers.items() if key.lower() not in strip_on_forward
     }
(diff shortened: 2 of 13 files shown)


lilly-luo:
can u add 1 of rohits integration tests to cover this pls

.github/workflows/integration.yml
```diff
@@ -181,6 +181,9 @@ jobs:
     env:
       UCODE_TEST_WORKSPACE: ${{ secrets.UCODE_TEST_WORKSPACE }}
       DATABRICKS_BEARER: ${{ secrets.DATABRICKS_BEARER }}
+      # Subscription OAuth token for the relayed hybrid CUJ in the Claude lane; the same secret
+      # the e2e workflow uses. The Codex lane ignores it.
```
lilly-luo: rm this

.github/workflows/integration.yml
```diff
@@ -181,6 +181,9 @@ jobs:
...
+      CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
```
lilly-luo, line 184: maybe CLAUDE_CODE_OAUTH_TOKEN -> CLAUDE_CODE_RELAYED_OAUTH_TOKEN

scripts/run_integration.py
```diff
@@ -464,6 +476,10 @@ def run(command, *, cwd=output, env=base_env, timeout=600) -> str:
                 "UG_INTEGRATION_RUN_DIR": str(output),
                 "UG_INTEGRATION_AGENTS": ",".join(agents),
                 "UG_INTEGRATION_CLAUDE_PROVIDER": args.claude_provider,
+                "UG_INTEGRATION_CLAUDE_RELAYED_PROVIDER": args.claude_relayed_provider,
+                # The relayed CUJ reads this and injects it as CLAUDE_CODE_OAUTH_TOKEN for that
+                # session only; a distinct name keeps non-relayed sessions from seeing it.
```
lilly-luo: rm this

src/ucode/agents/claude.py
```diff
@@ -1396,6 +1396,11 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
         port,
         token_header=gateway_proxy.AI_GATEWAY_TOKEN_HEADER,
         force_refresh_near_expiry=False,
+        # Let a relayed session also reach Databricks-hosted (OSS / system.ai) models:
+        # the proxy re-routes those requests to gateway auth while relayed subscription
+        # models keep the OAuth passthrough. Bare Anthropic ids are unaffected, so a
+        # pure-relay session behaves exactly as before.
```
lilly-luo: rm this

scripts/run_integration.py
```diff
@@ -202,9 +207,15 @@ def terminate(signum, frame):
     base_env["UV_CACHE_DIR"] = str(output / "cache")
     base_env["UV_DEFAULT_INDEX"] = args.default_index
     bearer = os.environ.get("DATABRICKS_BEARER", "").strip()
+    # The relayed hybrid CUJ forwards the subscription OAuth token to the session; keep it
+    # out of logs and the archived report just like the bearer.
```
lilly-luo: rm this

src/ucode/gateway_proxy.py
```diff
@@ -186,20 +194,57 @@ def forwarded_request_headers(
...
+def is_databricks_routed_model(model: str | None) -> bool:
+    """True when ``model`` is a Databricks-hosted (gateway-served) id rather than a model
+    the relayed Anthropic subscription serves.
+
+    Databricks ids are namespace-qualified (``system.ai.*``, ``catalog.schema.model``,
+    ``databricks-*``); the relayed subscription uses Anthropic's bare canonical names
+    (``claude-opus-4-1``, ``claude-sonnet-4-5``, ...), which never carry a dot."""
+    if not model:
+        return False
+    return "." in model or model.startswith("databricks-")
```
lilly-luo, line 224: can you check that it's a fqn instead of just checking for 1 dot? can use this util: https://github.com/databricks/unity-gateway/blob/030e632c208be133d43c8970aab263b37db5bd71/src/ucode/managed_setup.py#L267 , maybe move it into a shared util file

src/ucode/gateway_proxy.py
```diff
@@ -374,6 +436,7 @@ def start_proxy(
     port: int,
     token_header: str,
     force_refresh_near_expiry: bool,
+    relayed_oss_routing: bool = False,
```
lilly-luo: isn't this like always true? why would someone want to do relayed auth and not have oss routing? i think in the only call site of start_proxy, it's true? 

if it's always going to be true, can we remove the parameterization?

src/ucode/agents/claude.py
```diff
@@ -1396,6 +1396,11 @@ def _launch_relayed(state: dict, binary: str, tool_args: list[str]) -> None:
...
+        relayed_oss_routing=True,
```
lilly-luo: this is the only call site right? so it's always true?

src/ucode/gateway_proxy.py
```diff
@@ -218,15 +263,32 @@ def _handle(self) -> None:
         length = int(self.headers.get("Content-Length", 0) or 0)
         body = self.rfile.read(length) if length else None
         url = self.path.lstrip("/")
+        # Databricks-hosted models (when relayed OSS-routing is on) authenticate with the
+        # gateway token in `Authorization`; everything else keeps the relay path.
+        route_databricks = self.relayed_oss_routing and is_databricks_routed_model(
+            _request_model(body)
+        )
```
lilly-luo: this won't ever work for a custom model service that points to system.ai.glm-5-2 right? :(

.github/workflows/integration.yml
masonc08: a claude code relayed oauth token is the same as a claude code oauth token right?
also this is already being used in existing e2e tests so id rather just not change it anymore

src/ucode/gateway_proxy.py
masonc08: maybe we can followup with that

lilly-luo (approved):

PR #572, opened 2026-09-11
make databricks-sdk in the default dependencies 

lilly-luo:
so we don't run into confusion from the install


pyproject.toml
```diff
@@ -30,8 +31,6 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-# Custom-client OAuth uses the Databricks SDK's PKCE and token-cache helpers.
-custom-oauth = ["databricks-sdk>=0.135.0"]
```
lilly-luo, line 34: i will update isaac's cmd with the binary change so that removing this isn't a problem

rohita5l (approved):

PR #573, opened 2026-09-11
Bugfix: Discover model services outside the system.ai schema

### LINKED ISSUE #569: ug claude can't find any LLMs when model services live outside system.ai
Note: I wrote this with AI but I read through it and everything looks clear. TL/DR if you disable access to system catalog llms (many clients do this) and enable them as model services, ug/ucode won't be able to find them and fail on startup 

Image

## Summary

`ug claude` reports "AI Gateway returned no Claude model ids" and "No coding agents are available on this workspace" even when the workspace has Claude model services available. In my Hershey sandbox I have three Claude model services (`il-claude-haiku-4-5`, `il-claude-opus-5`, `il-claude-sonnet-5`) and none of them are discovered.

## Root cause

Model-services discovery is hardcoded to the `system.ai` schema in two places, so any Claude model service living in a different UC catalog/schema is invisible. My three models live in `il_model_service.default`, not `system.ai`.

The two hardcoded constants in `src/ucode/databricks.py`:

- [`_MODEL_SERVICE_REQUIRED_PREFIX = "system.ai."`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/databricks.py#L1453) (client-side name filter)
- [`_MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/databricks.py#L1461) (server-side query scope)

Both locks apply:

1. [`list_model_services()`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/databricks.py#L1631-L1637) sends `parent=schemas/system.ai`, so the API only returns services in `system.ai`. The `il_model_service.default` services never come back.
2. Even if they did, [`_model_service_id()`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/databricks.py#L1531-L1533) drops any name that doesn't start with `system.ai.`.

Family bucketing itself is fine (`il-claude-opus-5` contains `claude-opus-`), the ids just never reach it.

## Why the fallback doesn't cover it

For Claude, when model-services comes back empty, [`cli.py`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/cli.py#L612-L615) falls back to `discover_claude_models()`, which hits the AI Gateway's Anthropic foundation-models endpoint. This workspace has no foundation models, so that returns nothing too. Both paths assume the models are either in `system.ai` or exposed as foundation-model endpoints. These are neither, they are UC model services in a custom catalog.

Worth noting: [`model_service_exists()`](https://github.com/databricks/unity-gateway/blob/03aefdc796aaf7332ec97289701663a6391d5650/src/ucode/databricks.py#L1674-L1700) already scopes to arbitrary schemas via `parent=schemas/{catalog}.{schema}`, so the codebase already knows model services can live outside `system.ai`. The main discovery walk just doesn't.

## Steps to reproduce

1. Create Claude model services in a catalog/schema other than `system.ai` (mine are in `il_model_service.default`).
2. Have no foundation-model endpoints in the workspace.
3. Run `ug claude`.
4. Auth and gateway connect succeed, then: `ERROR No coding agents are available on this workspace` / `Claude models ...: AI Gateway returned no Claude model ids`.

## Suggested fix

A few options, trading speed vs completeness:

1. Broaden the walk and relax the `system.ai.` prefix requirement, bucketing purely by the `claude-<family>-` substring. Most robust, but an unscoped metastore walk can be slow on busy workspaces (the existing code comment notes ~50s).
2. Make the catalog/schema configurable (e.g. an env var like `UCODE_MODEL_SERVICE_SCHEMA`). Cheapest, keeps the fast single-page query.
3. `system.ai` first, then widen: keep the fast `system.ai` query, and only if it (and the foundation-model fallback) come up empty, do the broader walk. Keeps the common fast path and self-heals for custom-catalog workspaces.

Option 3 seems like the best balance.

## Environment

- CLI: `ug` / `ucode`
- Workspace: Azure Databricks (Hershey sandbox)
- Repo at commit `03aefdc`

mpkrass7:
## Problem

`ug claude` (and the other agents) report "No coding agents are available on this workspace" even when the workspace has Claude model services available. Repro: a workspace with three Claude model services (`il-claude-haiku-4-5`, `il-claude-opus-5`, `il-claude-sonnet-5`) whose Location is a custom UC catalog (`il_model_service.default` in the report), and no foundation-model endpoints. `ug` finds nothing:

```
ERROR No coding agents are available on this workspace.
 Claude models (needed for: claude, opencode, copilot, pi): AI Gateway returned no Claude model ids
 ...
 OSS models (needed for: opencode): model-services listing returned no models
```

See #569 for the full trace.

## Root cause

Model-service discovery is hardcoded to the `system.ai` schema in two places:

- `_MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"` scopes the server-side query, so the API only returns services in `system.ai`.
- `_model_service_id()` drops any name that doesn't start with `system.ai.`.

Model services in a custom UC catalog are invisible to both. This is family-agnostic: `discover_model_services()` buckets claude/codex/gemini/oss out of the same single `system.ai`-scoped walk, so a custom-catalog GPT/Gemini/OSS service would be missed the same way. The per-family fallbacks don't help either, because they look at foundation-model endpoints, not UC model services in other catalogs.

## Fix

Add an opt-in `UCODE_MODEL_SERVICE_SCHEMAS` env var: a comma-separated list of `catalog.schema` locations to search in addition to `system.ai`. Each is walked with its own `parent=schemas/{catalog}.{schema}` scope (one fast page, same as `system.ai`) and merged into the result.

This deliberately stays opt-in. `system.ai` remains the default, and when the var is unset the behavior is unchanged, so the common case never scoops up unrelated user schemas (the existing `test_ignores_non_system_ai_schemas` still holds). The paging loop is refactored into `_walk_model_services` so the `system.ai` walk and the extra-schema walks share it.

Usage — point discovery at the `catalog.schema` where your model services live:

```
export UCODE_MODEL_SERVICE_SCHEMAS=model_service_catalog.default   # your catalog.schema
ug claude
```

## Testing

- New `TestExtraModelServiceSchemas` covers env parsing (trim/dedupe/blank), discovery of Claude models in an extra schema, per-schema scoping of the extra walk, merging with `system.ai` ids, and the env-unset case still ignoring custom catalogs.
- Full `tests/test_databricks.py` passes (266), `ruff check` and `ruff format --check` clean.
- The 3 `test_claude_smart_routing_v2` PTY failures and the `test_e2e_user_agent` failure reproduce on a clean checkout in this sandbox (local socket / live-gateway env issues), unrelated to this change.

## Notes / open questions

- Scope is discovery only, which is the reported symptom. A custom-catalog id like `model_service_catalog.default.claude-opus-5` flows through launch fine; the only cosmetic gap is `_maybe_add_1m_suffix` not recognizing it for the auto `[1m]` suffix (its regex expects `system.ai.`/`databricks-` prefixes). Happy to widen that too if you'd like.
- Open to an alternative shape if you'd prefer auto-discovery of catalogs over an env var, but I leaned opt-in to preserve the intentional "don't pick up scratch schemas" behavior.

Fixes #569

### DIFF
--- a/src/ucode/databricks.py
+++ b/src/ucode/databricks.py
@@ -1460,6 +1460,29 @@ _MODEL_SERVICE_REQUIRED_PREFIX = "system.ai."
 # is the only server-side narrowing that works.
 _MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"
 
+# Some workspaces host their coding-agent model services in a custom UC catalog
+# rather than `system.ai` (e.g. `model_service_catalog.default`). Those are invisible
+# to the `system.ai`-scoped walk above, so discovery reports no models even when
+# several exist. This env var opts a workspace into searching additional
+# `catalog.schema` locations: a comma-separated list, each walked with its own
+# `parent=schemas/{catalog}.{schema}` scope (one fast page, like system.ai). It's
+# empty by default so the common case never scoops up unrelated user schemas.
+_MODEL_SERVICE_EXTRA_SCHEMAS_ENV = "UCODE_MODEL_SERVICE_SCHEMAS"
+
+
+def _extra_model_service_schemas() -> list[str]:
+    """Additional `catalog.schema` locations to search, from the env var.
+
+    De-duplicated, order-preserving; empty when the var is unset or blank."""
+    raw = os.environ.get(_MODEL_SERVICE_EXTRA_SCHEMAS_ENV, "")
+    schemas: list[str] = []
+    for part in raw.split(","):
+        part = part.strip()
+        if part and part not in schemas:
+            schemas.append(part)
+    return schemas
+
+
 # Supported OSS chat families, matched by name substring. Add an entry to
 # support a new family.
 _OSS_MODEL_FAMILIES = ("kimi-", "glm-", "deepseek-")
@@ -1517,18 +1540,24 @@ def model_token_limits(model_id: str) -> dict[str, int] | None:
     return None
 
 
-def _model_service_id(service: dict) -> str | None:
-    """Extract the `system.ai.<model-name>` id from one model-service entry.
+def _model_service_id(
+    service: dict, *, required_prefix: str = _MODEL_SERVICE_REQUIRED_PREFIX
+) -> str | None:
+    """Extract the `<catalog>.<schema>.<model-name>` id from one model-service entry.
 
-    Returns None for services in any other schema, so user/internal model
-    services don't leak into the family buckets."""
+    ``required_prefix`` defaults to ``system.ai.`` so services in any other
+    schema are dropped and user/internal model services don't leak into the
+    family buckets. Pass ``required_prefix=""`` to accept every id after the
+    ``model-services/`` prefix — used for the schema-scoped extra walks
+    (see :func:`_extra_model_service_schemas`), where the ``parent`` query has
+    already narrowed the listing to a single trusted schema."""
     name = service.get("name")
     if not isinstance(name, str):
         return None
     name = name.strip()
     if name.startswith(_MODEL_SERVICE_NAME_PREFIX):
         name = name[len(_MODEL_SERVICE_NAME_PREFIX) :]
-    if not name.startswith(_MODEL_SERVICE_REQUIRED_PREFIX):
+    if required_prefix and not name.startswith(required_prefix):
         return None
     return name or None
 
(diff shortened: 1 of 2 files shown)


src/ucode/databricks.py
```diff
@@ -1460,6 +1460,29 @@ def build_auth_shell_command(
 # is the only server-side narrowing that works.
 _MODEL_SERVICE_PARENT_SCHEMA = "schemas/system.ai"
 
+# Some workspaces host their coding-agent model services in a custom UC catalog
+# rather than `system.ai` (e.g. `il_model_service.default`). Those are invisible
+# to the `system.ai`-scoped walk above, so discovery reports no models even when
+# several exist. This env var opts a workspace into searching additional
+# `catalog.schema` locations: a comma-separated list, each walked with its own
+# `parent=schemas/{catalog}.{schema}` scope (one fast page, like system.ai). It's
+# empty by default so the common case never scoops up unrelated user schemas.
```
mpkrass7, line 1469:
```suggestion
# Handle situation where only model services and no system.ai models are available
# Discovery will look in environment variable location in addition to system models
```

mpkrass7:
Closing, Andy working on a better branch

PR #574, opened 2026-09-11
Extract walk_catalog_schemas scaffold in databricks.py

xsh310:
## What changed, and why?

**Change:** Extract the catalogs → schemas → parallel-per-schema-probe walk out of `list_all_mcp_services` into a generic `walk_catalog_schemas[T](*, deadline, probe, collect, skip_catalogs=…)` in `databricks.py`, and rewrite `list_all_mcp_services` on top of it. The scaffold owns catalog/schema enumeration, the worker pools, and the wall-clock deadline drain; the caller supplies the `probe` (`list_mcp_services`) and a `collect` that owns accumulation, dedup, progress, and streaming.

**Why:** Pure refactor (PR B in the `ug skill add` interactive-picker stack). Extracting the walk lets PR D's skills discovery reuse it as a second caller. `list_all_mcp_services`'s signature and every return value are unchanged.

Note: the design doc's PR B also mentioned rewriting `list_uc_functions_catalog_schemas`, but that function does not exist in this repo — there is a single walk today, so the scaffold has one caller now and the skills walk becomes the second in PR D.

## How do you know it works?

**Testing:** `ruff check` and `ruff format` clean; full `uv run pytest` green except the two pre-existing e2e failures (`test_e2e_user_agent`, `test_claude_smart_routing_v2`) that also fail on `main`. Added `TestWalkCatalogSchemas` (skips skip-catalogs + `information_schema`, probes each user schema, reports progress, returns the right phase-1 reason); the existing `TestListAllMcpServices` tests stay green unchanged.


src/ucode/databricks.py
```diff
@@ -2520,30 +2520,24 @@ def _paginated_json_items(
...
-    This walk is the slow, workspace-wide counterpart to `list_mcp_services`
-    (single schema)."""
+    deadline: float,
+    probe: Callable[[str, str], T],
+    collect: Callable[[T, int, int], None],
+    skip_catalogs: frozenset[str] = _UC_FUNCTIONS_SKIP_CATALOGS,
+) -> str | None:
+    """Walk every user `<catalog>.<schema>` in the workspace, probing each in parallel.
+
+    Phase 1 lists catalogs (minus `skip_catalogs`) and their schemas in parallel, dropping
```
xsh310: Can you update this docstring so that it's clearer, more concise, and easily understandable by human what this does. Focus on the high level functionality. 

Please also explain what probe and collect callables represents here in simple terms and what should be the shape of these callables.

src/ucode/databricks.py
```diff
@@ -2585,46 +2576,78 @@ def list_all_mcp_services(
...
             schemas, _ = result
             for schema in schemas:
-                schema_name = schema.get("name")
-                if (
-                    isinstance(schema_name, str)
-                    and schema_name
-                    and schema_name != "information_schema"
-                ):
-                    schema_refs.append(f"{catalog}.{schema_name}")
+                name = schema.get("name")
```
xsh310: why rename this variable to name from schema name?

src/ucode/databricks.py
xsh310: Updated in b1c7740: the docstring now leads with the high-level behavior (discover every user `<catalog>.<schema>` and probe each in parallel, with a deadline so a slow workspace returns partial results), and spells out the callables — `probe(catalog, schema) -> result` fetches one schema's data, and `collect(result, done, total)` handles each result as it lands (accumulate / dedup / stream), where `done`/`total` are the schema counts.

src/ucode/databricks.py
xsh310: No real reason — incidental churn from the extraction. Reverted to `schema_name` in b1c7740 to match the original.

src/ucode/databricks.py
```diff
@@ -2591,40 +2587,76 @@ def collect_schemas(result, catalog):
...
-            return [], "deadline exceeded while listing UC schemas"
-        return [], "no UC schemas found"
+            return "deadline exceeded while listing UC schemas"
+        return "no UC schemas found"
 
-    # Parallel per-schema mcp-services listing.
-    names: set[str] = set()
     schemas_total = len(schema_refs)
     schemas_done = 0
     probe_workers = max(1, min(_UC_FUNCTION_PROBE_WORKERS, schemas_total))
```
xsh310: Should we extract this worker count as a variable param as well, and also why is this named UC function prober worker here?

xsh310: Done in d96da01: renamed `_UC_FUNCTION_PROBE_WORKERS` → `_SCHEMA_PROBE_WORKERS` (it is just the concurrency cap for the parallel per-schema / per-catalog work — nothing UC-function-specific now that the walk is generic), and exposed it as a `max_workers` param on `walk_catalog_schemas` (default `_SCHEMA_PROBE_WORKERS`), mirroring `skip_catalogs`.

sunishsheth2009 (approved):

PR #578, opened 2026-09-11
Extract shared skill-download core and add selective download

xsh310:
## What

PR C of the `ug skill add` interactive-picker stack. It reshapes `skills_download.py` so the download mechanics become one reusable core, and adds the (still unwired) functions the picker will call in PR D. **No behavior changes on any existing path.**

Design doc: `ug_skill_add_download_interactive_picker_design.md` (PR table, row C).

## Changes

- **`SkillRef` carries its full location.** New `catalog`/`schema` fields plus an `fqn` property, parsed once from the API `name` in `_skill_ref`, so a ref is self-contained and can be downloaded on its own.
- **Extract `_download_refs(refs, roots) -> (written, total)`**, the shared core: collision dedup, overwrite prompt, parallel fetch, write. `should_download_skill` and `_reject_bundle_name_collisions` now read the source off the ref (`ref.fqn`) instead of an explicit `location` arg, which produces byte-identical messages and lets the core serve a flat, cross-schema set.
- **`_fetch_bundles` keys results by FQN** (a securable name is unique only within its schema) and takes its progress-bar label from the caller.
- **Rename `download_skills` to `download_skills_from_schema_locations`**, now calling `_download_refs` per location and keeping the exact per-location `Downloaded k/n from <location>` summary.
- **Add `get_skill(fqn)`** (GetSkill at `/2.1/unity-catalog/skills/<fqn>`, reusing `_skill_ref`; NOT_FOUND to `None`) and **`download_selected_skills(fqns, path)`** (one `_download_refs` pass over the flat selection). Both are dead code until PR D wires the picker.

## Deviations from the design doc

- `download_managed_skills_on_launch` is a **second caller of `_fetch_bundles`** the doc did not mention; it is updated to the new keying (behavior-preserving).
- The doc lists `should_download_skill` / `_reject_bundle_name_collisions` as reused unchanged, but making the core location-agnostic means they read the source off the ref. The emitted messages are identical (`<catalog>.<schema>.<securable>` equals `ref.fqn`).

## Testing

`uv run ruff check .`, `ruff format --check src tests`, and `uv run pytest` all pass (the 2 pre-existing e2e failures in `test_e2e_user_agent` and `test_claude_smart_routing_v2` are unrelated and also fail on the base branch). New unit tests cover `_download_refs` (including cross-schema dedup), `get_skill` (NOT_FOUND to `None`, unfinalized to `None`), and `download_selected_skills`.

Stacked on #574 (`skills-picker-walk`).


src/ucode/skills_download.py
```diff
@@ -301,32 +315,58 @@ def _fetch_bundles(
     return results
 
 
-def _reject_bundle_name_collisions(refs: list[SkillRef], *, location: str) -> list[SkillRef]:
+def _reject_bundle_name_collisions(refs: list[SkillRef]) -> list[SkillRef]:
```
xsh310, line 318: On second thought, why do we need this check given that we already have the prompt when we have write detection? Also given that now that we can have multiple locations, this would detect bundle collision across locations given the current download_skills logic right

xsh310: Not redundant with the overwrite prompt: that only catches a bundle dir already on disk, while this catches two skills in one batch claiming the same dir (neither on disk yet, since every `should_download_skill` decision runs before any write). But the handling is weak: it keeps whichever skill the listing returned first (arbitrary order) and drops the rest with just a warning, and since `total` is counted after the reject the summary can read `1/1` while a skill was silently dropped.

This helper is also used by the best effort launch path (which must never block), so the real fix belongs at the callers rather than the shared helper: make the drop deterministic and visible everywhere, and fail closed only on the explicit selection path (named or picked skills), where silently delivering a partial set is worst. Leaving it as is in this PR (a no behavior change refactor) and tracking the fix as a follow up.

src/ucode/skills_download.py
```diff
@@ -241,20 +255,18 @@ def existing_skill_on_disk(roots: list[Path], bundle_name: str) -> bool:
...
-    overwriting a skill already on disk (``location`` is the source
-    ``<catalog>.<schema>`` shown in that prompt), so a declined skill is never
-    fetched. Dedup keys on the bundle name, since that is the directory an agent
-    would load. Name validity is the server's job -- FinalizeSkill enforces the
-    Agent Skills naming rules on ``bundle_name`` before we ever see it -- so
-    ucode does not re-check it here.
+    overwriting a skill already on disk (naming the source by ``ref.fqn``), so a
+    declined skill is never fetched. Dedup keys on the bundle name, since that is
+    the directory an agent would load. Name validity is the server's job --
+    FinalizeSkill enforces the Agent Skills naming rules on ``bundle_name`` before
```
sunishsheth2009, line 265: **P1 (defense-in-depth):** `bundle_name` flows straight into filesystem paths — `write_skill` does `root / ref.bundle_name`, and `_skill_installs` records those same dirs, which `ug skill remove` later `shutil.rmtree`s. Today that's safe because, as this docstring says, FinalizeSkill enforces Agent Skills naming rules server-side (no `.`/`/`), so a legit `bundle_name` can't traverse. But it's a remote-provided value feeding a write **and delete** sink with zero client-side check — if the server guarantee ever regresses or is bypassed, we write/delete outside the skill dir. A one-line guard where the ref is built (reject `..`, `/`, or absolute paths, skip+warn) is cheap insurance for a delete sink. Not exploitable as-is — flagging for hardening.

sunishsheth2009 (approved):

src/ucode/skills_download.py
xsh310: Added a guard in `_skill_ref` that skips and warns on any `bundle_name` containing a separator, `..`, or an absolute path before it can name a directory, so the write and delete sinks no longer depend solely on FinalizeSkill. It's mirrored into `skills_api.py` for the upper-stack branches where `_skill_ref` now lives.

PR #579, opened 2026-09-12
Add interactive picker for `ug skill add` download

xsh310:
## What

PR D (final) of the `ug skill add` interactive-picker stack: it wires the feature. Running `ug skill add` with no `--mcp`, `--location`, or `--skills` on an interactive terminal now opens a picker of the workspace's skills to download, mirroring `ug mcp add`'s picker exactly.

Design doc: `ug_skill_add_download_interactive_picker_design.md` (PR table, row D). Builds on PR A (picker primitives), PR B (`walk_catalog_schemas`), and PR C (`download_selected_skills`).

## Changes

All in `skills_download.py` (discovery + picker glue) and `cli.py` (routing):

- **`list_all_skills`** — workspace-wide skill walk via `walk_catalog_schemas` (catalogs → schemas → skills in parallel under a wall-clock deadline), flattening and streaming each schema's finalized skills by FQN.
- **Picker glue** — `_skill_download_choice` (value = FQN, title flags on-disk bundles, which stay selectable so a re-download can update them), `_skills_download_background_loader` (runs `list_all_skills` on the token the command already fetched and streams discovered skills into the open picker as choices), and `prompt_for_skill_download_choices` (`scrolling_checkbox` with `loading_noun="skills"`).
- **`configure_skills_download_picker_command`** — opens the picker immediately, streams skills in behind it, then `download_selected_skills` + `register_schemaless_skills_connection` on save. Ctrl-C downloads nothing and leaves the connection untouched.
- **`skills_add`** — opens the picker when `--mcp`, `--location`, and `--skills` are all omitted on an interactive terminal (via `_stdin_is_interactive`). The typed `--location` / `--skills` paths are unchanged, and a non-interactive run (piped/CI) still errors with `--location is required`.
- **README** — documents the no-argument picker.

## Edge cases (handled by the reused primitives)

- **Empty / streaming:** the picker opens on an empty list and fills in via the background loader.
- **Deadline / partial results:** `walk_catalog_schemas` returns what it found once the budget is exceeded.
- **Permission / discovery failure:** discovery runs on the picker's daemon thread, so a 403 or transient failure just stops the stream and the picker stays open.
- **Ctrl-C:** exit 0, nothing downloaded, no registration.

## Testing

`uv run ruff check .`, `ruff format --check src tests`, `ty check src`, and `uv run pytest` all pass (the 2 pre-existing e2e failures in `test_e2e_user_agent` and `test_claude_smart_routing_v2` are unrelated and also fail on the base branch). New unit tests cover `list_all_skills` (flatten, stream, dedupe by FQN, progress, deadline, walk-failure reason), the picker glue (choice on-disk flag, background loader streaming, prompt selection and cancel), the picker command (download+register, Ctrl-C no-op), and CLI routing (interactive opens the picker, `--skills` stays on the typed path, non-interactive errors).

Stacked on #578.

https://github.com/user-attachments/assets/e5140d3c-cdf0-4746-b7cf-ffe93826cfe3


src/ucode/skills_download.py
```diff
@@ -505,3 +516,124 @@ def configure_skills_download_command(
...
+    if reason is not None:
+        return [], reason
+    if not by_fqn:
+        if time.monotonic() > deadline:
+            return [], "deadline exceeded while listing skills"
+        return [], "no skills found"
+    return sorted(by_fqn.values(), key=lambda ref: ref.fqn), None
+
+
+def discover_all_skills(
```
xsh310: Should we just inline this function? Seems to be a very thin wrapper

src/ucode/skills_download.py
```diff
@@ -505,3 +516,124 @@ def configure_skills_download_command(
...
+
+
+def _skill_download_choice(ref: SkillRef, roots: list[Path]) -> questionary.Choice:
+    """Picker row for one skill: value is its FQN, title flags an on-disk bundle.
+
+    On-disk skills stay selectable, since re-downloading is a legitimate update and
+    the existing overwrite prompt confirms it.
+    """
+    on_disk = " (on disk)" if existing_skill_on_disk(roots, ref.bundle_name) else ""
+    return questionary.Choice(title=f"{ref.fqn}{on_disk}", value=ref.fqn)
```
xsh310, line 575: qq: does this Choice support a subtitle? I'm wondering whether it could be possible to show a preview of the description in the Choice as well

src/ucode/skills_download.py
xsh310: Done, inlined it in da43d1c. It was a thin single use wrapper, and the loader now runs `list_all_skills` on the token the picker command already fetched, which also drops a redundant token lookup. I left the MCP path's `discover_all_mcp_service_names` as is, since it earns its keep there via a synchronous sibling (`discover_mcp_service_names`) that skills has no equivalent of.

src/ucode/skills_download.py
xsh310: Good idea, and it's feasible without any extra API calls. `questionary.Choice` does take a `description`, and `ListSkills` already returns each skill's description, so it arrives on the same streaming walk (no per skill `GetSkill`). Our `StreamingInquirerControl` also supports questionary's `show_description`; today `scrolling_checkbox` just hardcodes it off.

One caveat on rendering: our picker gives the choices window a fixed height equal to the row count, so questionary's inline description line would get clipped. The clean fit is a dedicated footer line showing the highlighted row's description (mirroring the existing search string / loading footers), which also handles descriptions being long (up to 1024 chars) by only previewing the pointed at row.

So the change is: capture `description` on `SkillRef` (defined back in PR C) + `_skill_ref`, pass it into the Choice, and add the footer preview to the picker. That's a modest scope bump on top of this wiring PR. Do you want it here, or as a follow up PR on top of the stack?

src/ucode/skills_download.py
```diff
@@ -505,3 +516,113 @@ def configure_skills_download_command(
...
+def _skills_download_background_loader(
+    workspace: str, token: str, roots: list[Path]
+) -> Callable[[Callable[[list[questionary.Choice]], None]], None]:
+    """A picker ``background_loader`` that streams the workspace-wide skill walk in as choices."""
+
+    def loader(append: Callable[[list[questionary.Choice]], None]) -> None:
+        def on_skills(refs: list[SkillRef]) -> None:
+            append([_skill_download_choice(ref, roots) for ref in refs])
+
+        list_all_skills(workspace, token, on_skills=on_skills)
```
sunishsheth2009, line 587: **P1:** same issue as the `--mcp` schema picker (#588): `list_all_skills` returns `(refs, reason)` and returns partial results when its 30s deadline is exceeded, but the loader drops the return. On a slow/large workspace the download picker silently shows a subset of skills with no indication the list is incomplete, so a user can download fewer skills than exist and never know. Capture `reason` and surface it (sentinel row in the picker, or a warning after it closes).

src/ucode/skills_download.py
```diff
@@ -586,21 +588,24 @@ def _skill_download_choice(ref: SkillRef, roots: list[Path]) -> questionary.Choi
...
 
-    def loader(append: Callable[[list[questionary.Choice]], None]) -> None:
+    def loader(append: Callable[[list[questionary.Choice]], None]) -> str | None:
         def on_skills(refs: list[SkillRef]) -> None:
             append([_skill_download_choice(ref, roots) for ref in refs])
 
-        list_all_skills(workspace, token, on_skills=on_skills)
+        found, reason = list_all_skills(workspace, token, on_skills=on_skills)
+        if reason == _SKILLS_WALK_TIMEOUT_REASON:
+            return f"⚠ Timed out after {int(_SKILLS_WALK_DEADLINE_SECONDS)}s, found {len(found)} skills"
```
xsh310, line 600: Can we maybe use this ⚠️ emoji?

src/ucode/skills_download.py
xsh310: Fixed: `list_all_skills` now reports the deadline even on partial results, and the picker footer swaps its loading count for `⚠️ Timed out after 30s, found N skills`, so a truncated list is never shown silently. The same fix is mirrored into `skills_api.py` for the branches above where `list_all_skills` now lives.

src/ucode/skills_download.py
xsh310: Done, switched to `⚠️`.

sunishsheth2009 (approved):

PR #580, opened 2026-09-12
use codex custom catalog for smart routing if exists 

lilly-luo:
this PR does the following: 

1/ makes it so that for smart routing, IF a custom catalog for codex is defined, we use the slugs in the catalog and send those models to smart routing instead of automatic model discovery. 
2/ consolidates the hierarchy of config files (managed -> cli -> personal) into 1 place 

i also confirmed that if a custom catalog is defined, then codex doesn't poll codex/v1/models. - [refresh strategy](https://github.com/openai/codex/blob/53ff712a48379ce8df605e292afd6046ca88ae9b/codex-rs/models-manager/src/manager.rs#L630-L640) is not used for the raw catalog. 

to test: 
https://github.com/user-attachments/assets/9427e919-82f1-4d4f-aea1-2aecedb5b0dd

i intentionally dropped glm-5-3-flash from the catalog. then i peeked at the payload we sent to the smart router `routes:select` and confirmed glm-5-3-flash was not sent.

### DIFF
--- a/src/ucode/codex_config.py
+++ b/src/ucode/codex_config.py
@@ -23,15 +23,22 @@ def codex_managed_config_path() -> Path | None:
 
 def codex_config_precedence_paths(
     managed_path: Path | None,
-    profile_path: Path = DEFAULT_CODEX_CONFIG_PATH,
+    profile_path: Path,
 ) -> tuple[Path, ...]:
     """Return Codex config paths in managed, profile, then user precedence."""
     config_home = os.environ.get("CODEX_HOME")
     if config_home:
         profile_path = Path(config_home).expanduser() / f"{CODEX_PROFILE_NAME}.config.toml"
+
+    # Highest precedence: machine-managed settings, normally /etc/codex/managed_config.toml.
+    managed_config = managed_path
+    # Middle precedence: Unity Gateway's ucode.config.toml layer passed to Codex via the CLI.
+    cli_config = profile_path
+    # Lowest precedence: $CODEX_HOME/config.toml, or ~/.codex/config.toml by default.
+    default_config = profile_path.parent / "config.toml"
     return tuple(
         path
-        for path in (managed_path, profile_path, profile_path.parent / "config.toml")
+        for path in (managed_config, cli_config, default_config)
         if path is not None
     )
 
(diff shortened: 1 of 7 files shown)


src/ucode/codex_config.py
```diff
@@ -1,12 +1,91 @@
...
+        return Path("/etc/codex/managed_config.toml")
+    return None
+
+
+def codex_config_precedence_paths(
+    managed_path: Path | None,
+    profile_path: Path,
+) -> tuple[Path, ...]:
+    """Return Codex config paths in managed, profile, then user precedence."""
+    config_home = os.environ.get("CODEX_HOME")
```
david-siqi-liu, line 31: just checking, this is OS specific right? I.e., different path for Windows

src/ucode/codex_config.py
```diff
@@ -1,12 +1,91 @@
...
+    # Highest precedence: machine-managed settings, normally /etc/codex/managed_config.toml.
+    managed_config = managed_path
+    # Middle precedence: Unity Gateway's ucode.config.toml layer passed to Codex via the CLI.
+    cli_config = profile_path
+    # Lowest precedence: $CODEX_HOME/config.toml, or ~/.codex/config.toml by default.
+    default_config = profile_path.parent / "config.toml"
+    return tuple(path for path in (managed_config, cli_config, default_config) if path is not None)
+
+
+def custom_catalog_models() -> list[str] | None:
```
david-siqi-liu, line 44: is there a similar function in https://github.com/openai/codex ? i'm guessing they also do some kind of parsing

david-siqi-liu (approved):
my high level comment is that a lot of function that seems o

src/ucode/codex_config.py
lilly-luo: CODEX_HOME is like ~/.codex which is same btw windows + mac. `codex_managed_config_path` is diff btw windows and mac. we currently don't really support windows......

src/ucode/codex_config.py
lilly-luo: there is `load_config_layers` however codex is all implemented in rust so we can't easily import the functionality

PR #581, opened 2026-09-12
Preview the highlighted skill's description in the picker

xsh310:
## What

A small UX follow-up on the `ug skill add` picker (PR #579): the picker now previews the highlighted skill's description in a footer, so a row is more than a bare FQN. Requested in review on #579.

## Changes

- **`SkillRef` gains a `description` field**, populated by `_skill_ref` from the `ListSkills` response, so it rides in on the same streaming walk with no per skill `GetSkill`.
- **`_skill_download_choice`** passes it through as the `Choice.description`.
- **`scrolling_checkbox` gains `show_description`** (default `False`, so the MCP picker is unchanged). When on, it renders the highlighted row's description in its own footer window, updating as you move the cursor.

## Why a footer rather than questionary's inline `show_description`

questionary's built-in `show_description` appends the description as an extra line inside the choices window. That window is sized to the row count in this picker, so the inline line would be clipped. A dedicated footer window (mirroring the existing search string and loading footers) sidesteps that, and showing only the highlighted row's description means a long one (UC caps skill descriptions at ~1024 chars) is confined to a bounded, 2 line wrapping footer via `_description_preview`, instead of bloating every row.

```
 › main.default.triage
   main.default.pii
   ml.prod.scoring
 Routes incoming support tickets to the right on call queue by severity.   ← footer: highlighted row
```

## Testing

`uv run ruff check .`, `ruff format --check src tests`, `ty check src`, and `uv run pytest` all pass (the 2 pre-existing e2e failures in `test_e2e_user_agent` and `test_claude_smart_routing_v2` are unrelated and also fail on the base branch). New unit tests cover `_description_preview` boundaries (short unchanged, at the limit unchanged, long clipped with an ellipsis), `_skill_ref` carrying the description, the Choice carrying it, and the picker being invoked with `show_description=True`.

Stacked on #579.

https://github.com/user-attachments/assets/11bc7368-f99d-47e5-ae76-a8bacace8791


src/ucode/ui/interactive_picker.py
```diff
@@ -210,6 +260,22 @@ def has_search_string() -> bool:
...
                     filter=Condition(lambda: control.error_message is not None),
                 ),
+                # Pinned at the bottom, one blank line below the other footers, so the
+                # highlighted row's description reads as a separate detail pane.
+                ConditionalContainer(
+                    HSplit(
+                        [
+                            Window(height=Dimension.exact(1)),
+                            Window(
+                                height=Dimension.exact(2),
```
sunishsheth2009, line 270: **P2 (UX):** the description footer window is fixed at `Dimension.exact(2)`. A UC description containing embedded newlines will have its later lines clipped even when it's well under the 240-char budget. UC descriptions are usually single-line, so minor — but consider collapsing whitespace/newlines into spaces before preview so the 2-line budget is spent on content rather than a hard-wrapped first line.

xsh310: Fixed: `_description_footer_tokens` now collapses whitespace runs (including newlines) to single spaces before previewing, so the two-line footer soft-wraps content instead of losing later lines to a description's own line breaks.

uv.lock
```diff
@@ -10,7 +10,7 @@ resolution-markers = [
 [[package]]
 name = "aiohappyeyeballs"
 version = "2.7.1"
-source = { registry = "https://pypi-proxy.cloud.databricks.com/simple/" }
+source = { registry = "https://pypi-proxy.cloud.databricks.com/simple" }
```
sunishsheth2009: Why did we change this? Can we undo this?

sunishsheth2009 (approved):

uv.lock
xsh310: removed

PR #583, opened 2026-09-12
[skills] Take cross-schema FQNs for named skill downloads

xsh310:
## What

Rework the "download these named skills" path so a selection is a set of fully-qualified `<catalog>.<schema>.<name>` names that may span any number of schemas, downloaded through `download_selected_skills` (the same flow the interactive picker uses). This applies to both:

- `ug skill add --skills <fqn>[,<fqn>...]`
- `ug configure skills --skill <fqn>[,<fqn>...]`

## Why

The old `--skills`/`--skill` was coupled to a single `--location`: bare leaf names required exactly one location, fully-qualified names had to all share one `<catalog>.<schema>`, and entries had to match `--location`. That could not express "download these specific skills from several schemas". Routing named downloads through the existing selected-download flow removes the limit.

## Behavior

- Each `--skills`/`--skill` entry must be a 3-part FQN; a bare or malformed name is now an error listing the offenders.
- `--skills`/`--skill` no longer takes `--location`; passing both is an error ("takes fully-qualified names; drop --location").
- Named downloads still register the schema-less skills MCP connection. `download_selected_skills` alone does not register, so a shared `configure_selected_skills_download_command` does the download then the register, mirroring the picker.
- An FQN that GetSkill cannot resolve (unknown or unfinalized) warns and is skipped per skill, not a hard error.
- `--skills ""` still downloads nothing.

Unchanged: whole-schema downloads via `--location`, `--mcp`, `--path`, and the no-arg interactive picker.

## Cleanup

Both CLI paths were the only callers of the subset filter on `download_skills_from_schema_locations` / `configure_skills_download_command`, so that now-dead `skills` parameter is removed; the schema-location path is whole-schema only.

## Stack

Part of gh stack #576, on top of #581 (skills-picker-download-preview).

## Testing

`uv run ruff check .`, `ruff format --check`, `ty check src`, and `uv run pytest` all pass (the 2 pre-existing `test_e2e_user_agent` / `test_claude_smart_routing_v2` failures also fail on the base). Rewrote `TestSkillsAddCommand` and the `configure skills --skill` cases, added `TestConfigureSelectedSkillsDownloadCommand`, and updated the README.


src/ucode/cli.py
```diff
@@ -350,6 +351,32 @@ def _parse_skill_locations(location: str | None) -> list[str]:
...
 
 
+def _is_qualified_skill_name(name: str) -> bool:
+    """True if `name` is a 3-part `<catalog>.<schema>.<name>` FQN with non-blank parts."""
+    parts = name.split(".")
+    return len(parts) == 3 and all(part and part == part.strip() for part in parts)
+
+
+def _download_selected_skills_route(
+    requested_skills: set[str], location: str | None, path: str | None, *, flag: str
```
xsh310: If the location here is only for throwing exception, why can't we check this in the main CLI function instead? I think we can probably inline this entire function

src/ucode/skills_download.py
```diff
@@ -505,20 +493,34 @@ def download_managed_skills_on_launch(
     return written
 
 
-def configure_skills_download_command(
-    locations: list[str], *, path: str | None, skills: set[str] | None = None
-) -> int:
+def configure_skills_download_command(locations: list[str], *, path: str | None) -> int:
```
xsh310: nit: configure_selected_skill_schemas_download_command

src/ucode/cli.py
xsh310: Done in fe09f96. Moved the `--location` rejection up into each command's guard block alongside the other mutual-exclusion checks, inlined the FQN validation and dispatch, and removed `_download_selected_skills_route`. Kept `_is_qualified_skill_name` as the shared predicate.

src/ucode/skills_download.py
xsh310: Renamed in fe09f96: the schema-download command is now `configure_location_skills_download_command`, so the name makes clear it downloads every skill in the selected `<catalog>.<schema>` locations. The by-name (cross-schema) sibling stays `configure_selected_skills_download_command`.

sunishsheth2009 (approved):

PR #584, opened 2026-09-12
[skills] Record downloaded-skill attribution in ~/.ucode/skills.json

xsh310:
## What

Track which on-disk skill directories ucode downloaded from Unity Catalog, in a
single manifest at `~/.ucode/skills.json`, so ucode can tell a downloaded skill
from a user-authored one and (in a later PR) remove downloads by their UC schema.

Skill attribution design doc:
https://doc-center-6051921418418893.staging.aws.databricksapps.com/d/hbzhsbid

## Why

The download path kept no record of what it wrote or where it came from, so there
is no safe way to remove a downloaded skill without risking a same-named skill the
user authored themselves. This is the foundation (PR 1 of the attribution work): it
populates the manifest but adds no user-facing command yet.

## Changes

- New `skills_state.py` owns the manifest and its read/write/query/reconcile logic.
  Records are keyed by `(metastore_id, fqn, base)`; query helpers are
  `list_downloaded`, `attribution_for_dir`, `records_for_schema`, and `forget`.
- `record_downloads` reconciles superseded records on write: it drops the stale
  record for a directory just overwritten (keeping the freshly written files) and
  deletes the orphaned old directory when the same skill's bundle name changed.
- `SkillRef` now also carries `metastore_id`, `skill_id`, and `uc_update_time`,
  read for free from the `ListSkills`/`GetSkill` response ucode already fetches.
- The three download flows (whole-schema, named FQNs, picker) and the managed-launch
  download record each skill after it reaches disk.

Kept separate from `state.json` so a state-version change and `ug revert` leave it
untouched.

## Testing

failures also fail on the base). New `tests/test_skills_state.py` covers the manifest
round-trip, corrupt/version handling, the queries, and all three reconciliation
cases; `tests/test_skills_download.py` gains coverage that a download records
attribution and that `SkillRef` captures the UC metadata.

## Stack

Part of gh stack #576, on top of #583 (`skills-picker-skills-fqn`).


src/ucode/skills_state.py
```diff
@@ -0,0 +1,171 @@
...
+    workspace: str
+    scope: str
+    base: str
+    dirs: tuple[str, ...]
+    metastore_id: str | None = None
+    skill_id: str | None = None
+    uc_update_time: str | None = None
+
+
+def _skills_path() -> Path:
```
xsh310: nit: can we call this _skills_state_path for clarity?

src/ucode/skills_state.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def _load() -> list[dict]:
+    try:
+        data = json.loads(_skills_path().read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError):
+        return []
+    if not isinstance(data, dict) or data.get("version") != SKILLS_STATE_VERSION:
+        return []
+    skills = data.get("skills")
```
xsh310: actually, can we call this top level field "skill_downloads" instead?

src/ucode/skills_state.py
```diff
@@ -0,0 +1,186 @@
...
+        if _claims_any(record, target):
+            continue
+        if _same_install(record, install):
+            _delete_dirs(record.get("dirs") or [])
+            continue
+        kept.append(record)
+    return kept
+
+
+def record_downloads(installs: list[SkillInstall]) -> None:
```
sunishsheth2009, line 144: **P2:** the manifest is a read-modify-write (`_load` → mutate → `_save`) with no locking. Two concurrent `ucode` invocations (e.g. a download in one shell, a remove in another) can interleave so the second clobbers the first's update — a just-recorded skill silently vanishes from attribution. `atomic_write_json` keeps the file well-formed but doesn't prevent the lost update. Rare in practice; worth a file lock or a re-read-before-write if concurrent runs are plausible.

xsh310: Leaving this as-is for now. In practice it needs two `ucode` runs mutating the manifest in the same sub-second window, which is unlikely for a single-user CLI that's normally run one command at a time. The failure is also bounded: `_save` uses `atomic_write_json` (temp file + `os.replace`), so every write lands as a complete, valid document and a reader never sees a half-written file. Worst case is one lost update (a dropped attribution record), never corruption, and the on-disk skill files are untouched. If concurrent runs turn out to be common, we can add an `fcntl.flock` guard around the load-modify-save (there's a precedent in `smart_routing/v2.py`).

sunishsheth2009 (approved):

PR #585, opened 2026-09-12
[skills] Support removing downloaded skills

xsh310:
## What

Wire the download side of `ug skill remove` (until now it errored with "not
supported yet"), mirroring `ug skill add`'s surface. Builds on #584, which records
what was downloaded from Unity Catalog.

## Why

With the attribution manifest in place, ucode can finally delete downloaded skills
safely: removal is driven entirely by attribution, so a same-named skill the user
authored themselves is never touched. That safety is the reason this could not ship
before the manifest.

## Behavior

- `ug skill remove --location <catalog.schema>` deletes every skill downloaded from
  that schema, across all download bases; `--path` limits it to one base.
- `ug skill remove --skills <fqn>[,<fqn>...]` deletes named fully-qualified skills
  that may span schemas (parity with `ug skill add --skills`).
- `ug skill remove` with none of those opens a picker over every downloaded skill,
  across all bases, each row labeled `<fqn>  (<scope>: <base>)`.
- `--mcp` removal is unchanged; `--agents` remains `--mcp`-only, and `--location` /
  `--path` / `--skills` are rejected with `--mcp`.

## Changes

- `skills_state.remove_downloads` deletes each record's directories, then forgets it;
  `records_for_fqns` looks records up by fully-qualified name.
- `remove_downloaded_skills_command` and the removal picker live in `skills_download`,
  next to the download flows they mirror.
- `ug skill remove` grows `--location`, `--path`, and `--skills`.

## Testing

failures also fail on the base). Added unit coverage for `remove_downloads` and
`records_for_fqns`, the by-location / by-name / picker removal paths (including that
a user-authored directory is left alone), and the CLI routing and argument validation.

https://github.com/user-attachments/assets/91c4e8db-8057-4d46-8e3b-6e36d562eb47

## Stack

Part of gh stack #576, on top of #584 (`skills-download-attribution`).


src/ucode/cli.py
```diff
@@ -1358,48 +1360,84 @@ def skills_add(
...
+    except (RuntimeError, ValueError) as exc:
         print_err(str(exc))
         raise typer.Exit(1) from None
     except KeyboardInterrupt:
         print_err("Interrupted.")
         raise typer.Exit(130) from None
 
 
+@skill_app.command("list")
+def skills_list() -> None:
```
xsh310: I think let's not add list in this PR for now

src/ucode/cli.py
```diff
@@ -1358,48 +1360,84 @@ def skills_add(
 
 @skill_app.command("remove")
 def skills_remove(
+    location: Annotated[
```
xsh310, line 1364: Should we also support --skills for parity with ug skill add?

README.md
```diff
@@ -350,6 +365,9 @@ The output looks like:
 | `ug skill add --skills main.default.my-skill` | Download named skills by fully-qualified name (comma-separated; may span schemas) |
 | `ug skill remove --mcp` | Remove skill schemas from the skills MCP connection (every agent) |
 | `ug skill remove --mcp --agents claude` | Remove skill schemas from specific agents only, keeping them on the rest |
+| `ug skill remove` | Pick from every downloaded skill (across all bases) and delete it from disk |
+| `ug skill remove --location main.default [--path <dir>]` | Delete every skill downloaded from a schema (all bases, or one under `<dir>`) |
+| `ug skill remove --skills main.default.my-skill` | Delete named downloaded skills by fully-qualified name (comma-separated; may span schemas) |
```
xsh310: --skill can only take --path, right?

src/ucode/skills_state.py
```diff
@@ -178,9 +178,29 @@ def records_for_schema(location: str, base: str | None = None) -> list[dict]:
...
     dropped = {_record_key(record) for record in records}
     _save([record for record in _load() if _record_key(record) not in dropped])
+
+
+def remove_downloads(records: list[dict]) -> None:
+    """Delete each record's on-disk directories, then drop it from the manifest."""
+    if not records:
+        return
+    for record in records:
+        _delete_dirs(record.get("dirs") or [])
```
sunishsheth2009, line 205: **P2:** `remove_downloads` deletes the on-disk dirs first, then `forget()` rewrites the manifest. If `_save` fails after deletion (disk full, etc.), the files are gone but their records remain → phantom entries pointing at deleted dirs on the next run. Also, `_delete_dirs` uses `shutil.rmtree(..., ignore_errors=True)`, so a dir that's a symlink is silently skipped yet still `forget()`-ten — the link lingers with no signal. Both are minor; consider forgetting-then-deleting, or at least noting undeletable paths.

xsh310: Fixed (b): `_delete_dirs` no longer swallows errors via `ignore_errors=True` — it unlinks symlinked dirs and collects any paths it couldn't remove, and `remove_downloads` warns listing them, so a remove never silently half-completes.

On (a), I kept the current delete-then-forget order deliberately: it's the more self-healing of the two. A phantom record left by a mid-op crash is still in the manifest, so the next `ug skill remove` re-targets it, `_delete_dirs` no-ops on the already-gone dirs, and `forget` cleans it up. Forget-then-delete would instead orphan untracked dirs on failure — the files linger with nothing tracking them, and remove can no longer target them, which doesn't self-correct. Both are rare, but delete-then-forget degrades better on retry.

sunishsheth2009 (approved):

PR #586, opened 2026-09-12
[skills] Record the workspace org id in downloaded-skill attribution

xsh310:
## What

Record the numeric workspace (org) id as `workspace_id` on each downloaded-skill
attribution record. Final PR of the attribution work, on top of #585.

## Why

The record already stores the workspace URL, but a URL can change (workspace
rename / vanity host). The numeric org id is a stable, rename-proof workspace
identifier, so attribution keeps both. It costs nothing extra to capture:
Databricks stamps every authenticated response with an `X-Databricks-Org-Id`
header, and the download already makes those calls.

## Changes

- `databricks.py`: the HTTP GET helpers capture `X-Databricks-Org-Id` into a
  session cache keyed by hostname (mirroring the existing model-service listing
  caches); `workspace_org_id(workspace)` reads it back, and a cache-clear hook is
  wired into the test isolation fixture. No extra request is made.
- `skills_state.py` / `skills_download.py`: `SkillInstall` carries `workspace_id`,
  set from `workspace_org_id` when a record is written. `metastore_id` remains the
  logical key; the field is omitted when no response has revealed the id yet.

## Testing

failures also fail on the base). Added coverage that the GET helper captures the
header (and tolerates its absence), and that a download records `workspace_id`
when the id is known and omits it when it is not.

## Stack

Part of gh stack #576, on top of #585 (`skills-download-remove`).


src/ucode/skills_state.py
```diff
@@ -38,6 +38,7 @@ class SkillInstall:
     base: str
     dirs: tuple[str, ...]
     metastore_id: str | None = None
+    workspace_id: str | None = None
```
sunishsheth2009, line 41: **P2:** `workspace_id` is now recorded, but the install identity key (`_record_key` → `(metastore_id, fqn, base)`) doesn't include it, and the `SkillInstall` docstring still describes the key without it. When `metastore_id` is None (older/edge records), two skills with the same `fqn` at the same `base` from *different* workspaces collide on reconcile and one overwrites the other. Either fold `workspace_id` into the identity key or update the docstring to state it's descriptive-only.

xsh310: Keeping `workspace_id` out of the identity key. A UC skill is identified by its metastore + fqn; `workspace_id` only records which workspace we downloaded through, and two workspaces attached to the same metastore see the same skill. Folding it into the key would over-disambiguate that case, splitting one shared-metastore skill into two install records pointing at the same on-disk copy. So the key stays `(metastore_id, fqn, base)` and `workspace_id` is descriptive-only.

On the `metastore_id is None` edge: it's narrow (new records always populate `metastore_id`, so it's limited to legacy/edge records) and only collides for two same-fqn skills at the same base from different workspaces. Not worth changing the key semantics for. If it ever surfaces, the clean fix is to fall back to `workspace_id` only when `metastore_id` is absent, which closes the edge without touching the common shared-metastore case.

sunishsheth2009 (approved):

PR #588, opened 2026-09-13
[skills] Add an interactive schema picker to `ug skill add --mcp`

xsh310:
## What

`ug skill add --mcp` with no `--location` errored asking for a typed schema. On an
interactive terminal it now opens a picker of the workspace's schemas that contain
skills, streamed in as discovery finds them, and adds the selected schemas to the
skills MCP connection's scope.

## Why

The download path already has an interactive picker; the MCP scope path was the only
add-side flow still requiring a typed `--location`. This makes the two symmetric.

## How

- The picker (in `mcp.py`, beside the other skill MCP-scope commands) reuses the download
  picker's machinery: the workspace walk (`list_all_skills`) and the streaming checkbox
  widget. `list_all_skills` probes one schema per call, so each streamed batch is one
  schema's skill set; the loader renders one row per schema with its exact skill count.
- Already-scoped schemas are flagged and stay selectable; re-selecting is a no-op, since
  adding is additive (removal is `ug skill remove --mcp`).
- To let both `mcp.py` (picker) and `skills_download.py` (download) use the workspace walk
  without a circular import, the UC skills read layer (`SkillRef`, the API client, and
  `list_all_skills`) moves into a new `skills_api.py` that depends only on `databricks.py`.
- `add_skills_command`'s apply body is extracted into `add_skill_locations_to_mcp`, shared
  by the command and the picker so each runs a single client setup.
- Non-interactive (no TTY) still errors asking for `--location`, so scripts fail loudly.

## Testing

`uv run ruff check`, `ruff format --check`, `ty check src`, and `uv run pytest` all pass
(the pre-existing `test_e2e_user_agent` / `test_claude_smart_routing_v2` failures also fail
on the base). The read-layer tests move to `test_skills_api.py`; added tests for the schema
picker, its command, `add_skill_locations_to_mcp` / `configured_skill_locations`, and the CLI
dispatch.

## Stack

Part of gh stack #576, on top of #586 (`skills-download-workspace-id`).

https://github.com/user-attachments/assets/bf080f49-2db7-4348-b416-72f76635d734

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


src/ucode/cli.py
```diff
@@ -1338,8 +1339,16 @@ def skills_add(
             return
         locations = _parse_skill_locations(location)
         if not locations:
-            if not mcp and _stdin_is_interactive():
-                configure_skills_download_picker_command(path=path)
+            if _stdin_is_interactive():
+                if mcp:
+                    scope = (
```
xsh310: nit: should we call it agents instead of scope so that it's less confusing?

src/ucode/skills_download.py
```diff
@@ -689,6 +694,76 @@ def configure_skills_download_picker_command(path: str | None = None) -> int:
...
+        style=picker_style(),
+        background_loader=background_loader,
+        loading_noun="skill schemas",
+    ).ask()
+    if selection is None:
+        return None
+    return [str(value) for value in selection]
+
+
+def configure_skills_mcp_picker_command(agents: set[str] | None = None) -> int:
```
xsh310: Should this mcp picker live in the skills_download.py? What's the best way to structure these util functions?

src/ucode/mcp.py
```diff
@@ -1975,6 +1975,28 @@ def _union_locations(base: list[str], new: list[str]) -> list[str]:
...
+    """Add ``locations`` to each client's skill scope, keeping any already configured."""
+    locations_by_client = _skill_locations_by_client_from_state(state)
+    for client in clients:
+        locations_by_client[client] = _union_locations(
+            locations_by_client.get(client, []), locations
+        )
+    _update_skills_mcp(state, workspace, profile, clients, locations_by_client)
+
+
+def configured_skill_scope(state: dict, clients: list[str]) -> set[str]:
```
xsh310: nit: can we call it configured_skill_schemas or configured_skill_locations instead? scope is a really overloaded term.

src/ucode/skills_download.py
```diff
@@ -689,6 +694,76 @@ def configure_skills_download_picker_command(path: str | None = None) -> int:
...
 
+def _skill_schema_choice(location: str, skill_count: int, in_scope: bool) -> questionary.Choice:
+    """Picker row for one schema: value is ``<catalog>.<schema>``, title carries the skill count.
+
+    An already-scoped schema is flagged and stays selectable; re-selecting it is a no-op, since
+    adding to the MCP scope is additive (removal is ``ug skill remove --mcp``).
+    """
+    noun = "skill" if skill_count == 1 else "skills"
+    scope_flag = "  (in scope)" if in_scope else ""
+    return questionary.Choice(
```
xsh310: Can we make this string more clear? Maybe "(already loaded in skill MCP)"?

src/ucode/mcp.py
```diff
@@ -1975,6 +1975,28 @@ def _union_locations(base: list[str], new: list[str]) -> list[str]:
     return merged
 
 
+def add_skill_scope(
```
xsh310: nit: maybe add_skill_locations_to_mcp

src/ucode/cli.py
xsh310: Renamed to `configured_agents` in both branches. `agents` itself is the CLI option name, so reusing it would shadow it; `configured_agents` distinguishes it from the raw `agents` / parsed `requested_agents`.

src/ucode/skills_download.py
xsh310: Good call. Moved the whole picker (choice builder, background loader, prompt, and `configure_skills_mcp_picker_command`) into `mcp.py`, next to the other skill MCP-scope commands (`add_skills_command`, `configure_skills_mcp_command`, `remove_skills_command`). `skills_download.py` now has zero net change. The one cross-module need, the discovery walk `list_all_skills`, is pulled in with a local import, since the import graph is one-way (skills_download imports mcp, not the reverse).

src/ucode/mcp.py
xsh310: Renamed to `configured_skill_locations`.

src/ucode/skills_download.py
xsh310: Changed to `(already in skill MCP)`. (This function now lives in mcp.py after the move above.)

src/ucode/mcp.py
xsh310: Renamed to `add_skill_locations_to_mcp`.

src/ucode/mcp.py
```diff
@@ -1983,12 +2006,77 @@ def add_skills_command(locations: list[str], agents: set[str] | None = None) ->
...
+    ``list_all_skills`` probes one schema per call, so each ``on_skills`` batch is that schema's
+    complete skill set: one row per schema, carrying its exact skill count.
+    """
+
+    def loader(append: Callable[[list[questionary.Choice]], None]) -> None:
+        def on_skills(refs: list[SkillRef]) -> None:
+            location = f"{refs[0].catalog}.{refs[0].schema}"
+            append([_skill_schema_choice(location, len(refs), location in in_scope)])
+
+        list_all_skills(workspace, token, on_skills=on_skills)
```
sunishsheth2009, line 2038: **P1:** `list_all_skills(...)` returns `(refs, reason)` and yields *partial* results once its 30s wall-clock deadline is hit (or a schema probe errors), but the loader discards the return. When the walk times out on a large/slow workspace, the picker simply stops adding rows and the "Loading more…" footer disappears — indistinguishable from a complete load. The user then scopes only the schemas that happened to arrive in time, believing they've seen everything.

Suggest capturing `reason` and surfacing it — e.g. append a disabled sentinel row (`⚠ schema discovery timed out — showing partial results`) or `print_warning` after the picker closes. Same pattern exists in the download picker's `_skills_download_background_loader` (#579).

xsh310: Fixed: the schema picker's loader now captures `list_all_skills`'s return and surfaces `⚠️ Timed out after 30s, found N skill schemas` in the footer, so a timed-out walk no longer silently shows a partial schema list. (`list_all_skills` already reports the deadline as of the download-picker fix lower in the stack.)

sunishsheth2009 (approved):

PR #590, opened 2026-09-13
[AIGTWY-4573] Parse the managed coding-agent config

david-siqi-liu:
Parse the managed `CodingAgentConfig` and add the version and watermark helpers on their own, ahead of any launch or configure change, so the data-shape change reviews in isolation. Nothing here runs on the launch path yet; the child PR wires it in.

Changes:
- Model the wire `CodingAgentConfig` as frozen dataclasses (`CodingAgentConfig`, `EnabledAgent`/`AgentConfig`, `AgentModels`, the `NamesOrLocation` selector, `SpendTiers`) parsed via `from_wire` and projected to the internal shape via `to_internal`, so the interface shape is legible instead of scattered dict reads. The model source is the server's oneof: exactly one of `model_provider_service` / `unity_catalog_location` / `model_services`.
- `mcp_servers` and `skills` each parse to a names-or-location selector: an absolute `names` list of UC FQNs, or a `unity_catalog_location` (a `catalog.schema`) whose contents are discovered, exactly one.
- Greenfield, so no back-compat: the deprecated wire fields the proto reserves are not read (`custom_headers`, the `model_config` oneof, `tracing_config`, `tags`), and neither are the reserved top-level `display_name` or the removed tracing table.
- `managed_config`: add `managed_config_is_newer` / `managed_update_time`, a `spec_version` forward-compat gate (a newer or malformed `spec_version` keeps the last-known-good cache), and the `UCODE_MANAGED_CONFIG_STUB` local-file read for pre-server testing.
- `state`: add the applied-version watermark (`get`/`set_applied_managed_update_time`) the launch gate reads.
- `managed_setup`: serialize and validate the same shape (names are 3-part FQNs, location a 2-part `catalog.schema`, exactly one set), so serialize then normalize round-trips.
- `databricks`: the managed-config update mask sends the mutable, non-reserved paths (`default_agent`, `enabled_agents`, `mcp_servers`, `skills`, `spend_tiers`); rename the user-facing "Unity AI Gateway" strings to "Unity Gateway".


src/ucode/managed_config.py
```diff
@@ -36,7 +36,7 @@
 )
 from ucode.ui import console, print_warning
 
-MANAGED_STATE_PATH = config_io.APP_DIR / "managed-state.json"
+MANAGED_CONFIG_PATH = config_io.APP_DIR / "managed-config.json"
```
lilly-luo, line 41: will this break any current ug users

src/ucode/managed_config.py
```diff
@@ -55,18 +55,13 @@
...
-    "MCP_SERVER_TYPE_VECTOR_SEARCH": "vector-search",
-    "MCP_SERVER_TYPE_UC_FUNCTIONS": "uc-functions",
-    "MCP_SERVER_TYPE_DATABRICKS_APP": "app",
-    "MCP_SERVER_TYPE_DATABRICKS_SQL": "sql",
+_AGENT_ENUM_PREFIX = "CODING_AGENT_"
+AGENT_NAME_TO_TOOL: dict[str, str] = {
+    enum[len(_AGENT_ENUM_PREFIX) :].lower(): tool for enum, tool in AGENT_ENUM_TO_TOOL.items()
 }
 
+MAX_SPEC_VERSION = 1
```
lilly-luo, line 65: any interest in passing this in as an env var and defaulting to 1? will allow users to migrate more easily without needing to update their client version

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    config_in = _as_dict(config)
+    agent_config: dict = {}
+    headers = _clean_str_dict(config_in.get("http_headers")) or _clean_str_dict(
+        config_in.get("custom_headers")
+    )
+    if headers:
+        agent_config["custom_headers"] = headers
+    tracing_table = _tracing_table(config_in.get("tracing")) or _tracing_table(
+        config_in.get("tracing_config")
+    )
```
lilly-luo: ? do we have to support both keys?

lilly-luo: rm legacy stuff, we can just create brand new

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    if tracing_table:
+        agent_config["tracing_table"] = tracing_table
+    model_config = _normalize_agent_models(config_in) or _normalize_model_config(
+        config_in.get("model_config")
+    )
```
lilly-luo: is this also 2 params for legacy handling? if yes, can you either: 

1/ split out the legacy handling into a separate file or func so it's clear what's legacy and what isn't 
2/ remove the legacy stuff since it's not clear to me that anyone actually used the legacy and it will create confusion in ug codebase

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    """Map an agent reference to a ucode tool name, accepting either spelling.
+
+    The server may send agent references as either proto enum (``CODING_AGENT_CLAUDE_CODE``) or
+    by name (``claude_code``). Both resolve to the same tool, or None when this build doesn't
+    know the agent.
+    """
+    name = _str(key)
+    if name is None:
+        return None
+    return AGENT_ENUM_TO_TOOL.get(name) or AGENT_NAME_TO_TOOL.get(name)
```
lilly-luo, line 397: same comment about legacy handling

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    overall_default = _str(default_models.get("default_model"))
+    if overall_default:
+        result["default_model"] = overall_default
+    models_dict = _as_dict(agent.get("models"))
+    provider = _str(models_dict.get("model_provider_service"))
+    if provider:
+        result["model_provider_service"] = provider
+    location = _str(models_dict.get("unity_catalog_location"))
+    if location:
+        result["model_service_location"] = location
```
lilly-luo: should u throw if there's a model provider service AND unity catalog location specified? that won't work right? 

or if not throw, at least do elif so the intent is clear

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    model_services = _str_list(models_dict.get("model_services"))
+    if model_services:
+        result["names"] = model_services
+    slots = {
+        slot: model
+        for slot in (
+            "default_opus_model",
+            "default_sonnet_model",
+            "default_haiku_model",
+            "default_fable_model",
```
lilly-luo: any way to make this dynamic so it will automatically support a new family?

lilly-luo: any way to make this dynamic so it will automatically support a new family?

src/ucode/managed_config.py
```diff
@@ -147,51 +142,128 @@ def _normalize_model_config(model_config: object) -> dict | None:
...
+    both are carried when present and downstream resolves whichever it finds. Returns None when
+    neither is set."""
+    value_dict = _as_dict(value)
+    result: dict = {}
+    names = _str_list(value_dict.get("names"))
+    if names:
+        result["names"] = names
+    location = _str(value_dict.get("unity_catalog_location"))
+    if location:
+        result["location"] = location
```
lilly-luo: so i can set both names and unity_catalog_location? but unity_catalog_location will be ignored if i set specific model names? if yes can we do an elif so the intent is clear in the code

src/ucode/managed_config.py
```diff
@@ -239,33 +313,90 @@ def normalize_managed_config(raw: dict) -> dict:
...
+    skills = _normalize_uc_names_or_location(raw.get("skills"))
+    if skills:
+        result["skills"] = skills
     tracing_table = _tracing_table(raw.get("tracing"))
     if tracing_table:
         result["tracing_table"] = tracing_table
-    budget_policy = _normalize_budget_policy(raw.get("budget_policy"))
-    if budget_policy is not None:
-        result["budget_policy"] = budget_policy
+    spend_tiers = _normalize_budget_policy(raw.get("spend_tiers") or raw.get("budget_policy"))
```
lilly-luo: maybe just remove the budget_policy part ?

src/ucode/managed_config.py
```diff
@@ -239,33 +313,90 @@ def normalize_managed_config(raw: dict) -> dict:
...
+    Either way the result keys by ucode tool name, dropping agents this build doesn't recognize.
+    """
+    enabled_agents: dict[str, dict] = {}
+    if isinstance(raw_agents, list):
+        for entry in raw_agents:
+            normalized = _normalize_enabled_agent(entry)
+            if normalized is not None:
+                tool, agent_config = normalized
+                enabled_agents[tool] = agent_config
+    elif isinstance(raw_agents, dict):
```
lilly-luo: is this also for legacy? if yes, poss to simplify ?

src/ucode/managed_export.py
```diff
@@ -18,11 +18,28 @@
...
+    "workspace_id",
+    "create_time",
+    "update_time",
+    "retrieved_time",
+    "created_user_id",
+    "updated_user_id",
+    # Stripped so the explicit EXPORT_SPEC_VERSION envelope below always wins over any
+    # spec_version the raw config happens to carry.
+    "spec_version",
+)
```
lilly-luo, line 42: can we also validate this server side?

src/ucode/managed_setup.py
```diff
@@ -182,36 +169,50 @@ def claude_model_slots(models: list[str]) -> dict[str, str]:
...
     """
-    body: dict = {}
+    models_obj: dict = {}
     mps = model_config.get("model_provider_service")
     if isinstance(mps, str) and mps:
-        body["model_provider_service"] = mps
+        models_obj["model_provider_service"] = mps
+    location = model_config.get("model_service_location")
+    if isinstance(location, str) and location:
+        models_obj["unity_catalog_location"] = location
```
lilly-luo, line 187: elif?

lilly-luo (commented):
could you add a @dataclass that represents the overall API? it's difficult to grok the shape right now. the dataclass will also help with request / response validation, it's easy to make an error and use the wrong interface

src/ucode/managed_setup.py
```diff
@@ -222,19 +223,15 @@ def _enabled_agent_payload(tool: str, agent_config: dict) -> dict:
     if isinstance(headers, dict):
         clean = {k: v for k, v in headers.items() if isinstance(k, str) and isinstance(v, str)}
         if clean:
-            config["custom_headers"] = clean
+            config["http_headers"] = clean
```
lilly-luo, line 221: make config a dataclass so we can access via config.http_headers instead

src/ucode/managed_config.py
david-siqi-liu: To be clear this is the maximum version that the currently installed unity-gateway distribution supports. Basically if the GET sees a higher version, then that means the currently installed unity-gateway is out of date and the user must `ug upgrade`, otherwise it wouldn't understand the new spec

src/ucode/managed_config.py
david-siqi-liu: removed

src/ucode/managed_config.py
david-siqi-liu: probably ya, but if they run `ug configure` again, they'll get it working again

src/ucode/managed_config.py
david-siqi-liu: removed

src/ucode/managed_config.py
david-siqi-liu: api does validation, changed to elif chain

src/ucode/managed_config.py
lilly-luo: this only affects ppl who were using the coding agent config. no one in prod was using it yet, so this is fine

src/ucode/managed_config.py
```diff
@@ -55,18 +57,284 @@
...
+class SpendTiers:
+    """Spend-based routing tiers (the wire ``spend_tiers`` / proto ``SpendTiers``)."""
+
+    budget_id: str | None = None
+    tiers: list[SpendTier] | None = None
+
+    @classmethod
+    def from_wire(cls, value: object) -> SpendTiers | None:
+        """Parse the wire ``spend_tiers``."""
+        bp = _as_dict(value)
```
lilly-luo: nit bp -> budget_policy

lilly-luo (approved):

PR #591, opened 2026-09-13
[AIGTWY-4573] Resolve managed static model lists and discovery locations

david-siqi-liu:
Add the shared `managed_resolve` helpers that both agent pickers need, a static model allow-list and a discovery location, before the agent-specific consumers so the shared logic reviews once.

**Changes**
- `managed_resolve`: add `managed_static_models` and `managed_model_service_location`; populate `claude_static_models`, `codex_static_models`, and the per-agent model_service_location into state; extend `managed_supplies_models` so claude counts a discovery location (skips ug discovery) but codex does not (codex has no reader, so discovery still runs).

**Validate**
- Focused: `uv run pytest tests/test_managed_resolve.py`.
- Scope check: the diff against the base touches only managed_resolve.py and its test.


david-siqi-liu:
Combined into #594 (the shared managed_resolve helpers live there now, alongside the picker/catalog/headers). Closing.

PR #592, opened 2026-09-13
[AIGTWY-4573] Curate Claude's /model picker from the managed config

david-siqi-liu:
Curate Claude's `/model` picker from the managed config: a static list writes `availableModels`, `enforceAvailableModels`, and `modelPicker`; a discovery location turns on gateway discovery instead. Ownership markers let ug prune only what it wrote.

**Changes**
- `claude`: render the picker from `claude_static_models`, or enable discovery from `claude_model_service_location` (both skipped under a provider or relayed launch); on a static-to-discovery transition, prune only ug-owned picker keys and leave an admin-authored picker intact. Markers clear only when the OS-managed file was actually written, so a suppressed write cannot orphan them.

**Validate**
- Focused: `uv run pytest tests/test_agent_claude.py`.
- Covered by tests: a static list produces the picker keys; a discovery location sets the discovery env and writes no picker; a transition prunes only ug-owned keys; a suppressed write preserves the markers.


david-siqi-liu:
Combined into #594, which now carries the per-agent managed consumption (Claude picker, Codex catalog, and http_headers) as one PR. Closing.

PR #593, opened 2026-09-13
[AIGTWY-4573] Apply the managed config at launch and version-gate it

david-siqi-liu:
With the config parsed (the parent PR), apply the managed configuration to every enabled agent at launch, but only when it changed since the last apply (an update_time watermark) or on `--refresh`, so an unchanged launch never re-writes the OS-managed files or prompts for a password. Also wire the managed MCP servers and skills, drop the post-configure probe validation, and rename user-facing strings.

Changes:
- Gate the launch apply on update_time. `refresh_managed_config` always fetches; a launch re-applies the config (the only step that writes the OS-managed files and can prompt for a password) only when update_time is newer than the applied watermark, or on `ug <agent> --refresh`, and records the new watermark. An unchanged launch skips the apply and never prompts. `ug configure` always applies. The per-tool launch write is wrapped in `suppressed_managed_writes()` so the apply-all owns the single OS write.
- Apply the managed config's mcp_servers and skills to each enabled agent. Both are UC-services-only selectors: register the absolute names (as UC mcp-services) or download the named skills, plus everything discovered under a unity_catalog_location (a schema's mcp-services, or its skills), unioned. Nothing on the launch path turned the workspace's published servers or skills into agent registrations before.
- `ug configure` wires the managed MCP servers and skills at configure time (not only at launch), and suppresses the interactive MCP/skills/plugins prompt when the managed config already defines them (and never installs under `--dry-run`).
- Stop `ug configure` from validating agents with a probe message: remove the post-configure validation and the `--skip-validate` flag (the probe reverted a freshly-written config on failure, and the gateway enforces model access regardless). `validate_tool` stays for headless and e2e callers.
- When a config we previously applied is gone (deleted, or the feature turned off), the launch does not auto-revert; it points the developer to `ug revert`, but only when the config is definitively gone (NOT_FOUND or feature off), not on a transient fetch failure. Keyed on the applied watermark.
- Drop the bare-ug launch banner and redundant "Starting <agent>" line; rename "CLI Managed Configuration" to "managed configuration".


src/ucode/agents/__init__.py
```diff
@@ -472,24 +480,35 @@ def launch(
     _MODULES[tool].launch(state, tool_args, options=options)
 
 
-def check_gateway_endpoint(state: dict, tool: str) -> bool:
-    """V2-only: a tool is available iff we discovered models for it."""
+def check_gateway_endpoint(state: dict, tool: str, managed: dict | None = None) -> bool:
+    """V2-only: a tool is available iff we discovered models for it or the managed config supplies them.
```
lilly-luo: what does v2 only mean here. should we remove it. i hope it's not aigw v2...

src/ucode/cli.py
```diff
@@ -765,7 +746,6 @@ def configure_workspace_command(
     *,
     prompt_optional_updates: bool = True,
     use_pat: bool = False,
-    skip_validate: bool = False,
```
lilly-luo, line 770: yay

src/ucode/cli.py
```diff
@@ -807,21 +795,6 @@ def configure_workspace_command(
...
-        if ok:
-            print_success(f"{spec['display']} is working")
-        else:
-            print_err(f"{spec['display']}: {provider_permission_error(tool, state, err)}")
-            managed = bool(state.get("managed_configs", {}).get(tool))
-            restore_file(spec["config_path"], spec["backup_path"], managed)
-            available_tools = [t for t in (state.get("available_tools") or []) if t != tool]
-            state["available_tools"] = available_tools
-            save_state(state)
-            raise RuntimeError(f"{spec['display']} validation failed — config reverted.")
```
lilly-luo, line 826: yay

src/ucode/cli.py
```diff
@@ -2110,24 +2152,44 @@ def _launch_tool(
...
         )
-        # An admin-published managed config wins over the developer's own settings. Layered on after
-        # `configure_shared_state`, whose returned state it overrides, and before the provider and
-        # model are settled below — the two state files are never merged on disk.
-        # Bare `ucode` already read one to choose the agent; refetching would double the round trip.
+        # Re-apply (and re-write the OS-managed files, the only step that can prompt for a password)
+        # only when the config changed since last applied here, or on `--refresh`. The launched
+        # tool's own write below is suppressed, so this apply-all owns the OS write.
+        applied_ut = get_applied_managed_update_time(existing)
+        if managed is not None and (refresh or managed_config_is_newer(managed, applied_ut)):
```
lilly-luo: can we check not just the timestamp but also that the workspace url is the same? if the workspace url is diff, we should always reapply 

in the case below 
```
time = t
workspace A -> updated at t-1 --> no managed config, apply for the first time at time t
workspace B -> updated at t-2 --> managed config but it's not newer than t. so it doesn't get updated 
``` 

we currently don't reapply. but we should

src/ucode/cli.py
```diff
@@ -2233,18 +2312,21 @@ def _launch_tool(
...
-            relayed=relayed,
-            route_root_model=route_root_model,
-            # Claude's explicit model is launch-scoped and is passed through LaunchOptions below.
-            custom_model=None,
-            coding_agent_config_defaults=coding_agent_config_defaults,
-        )
+        # The OS-managed (sudo) write is owned by the version-gated apply-all above and by `ug
+        # configure`; a launch only refreshes the user-level config and computes launch params, so
+        # suppress the OS write here to keep an unchanged launch prompt-free.
+        with suppressed_managed_writes():
```
lilly-luo, line 2353: does this mean that we ONLY write to managed settings during `ug configure`? i think this is what i was worried about re the autoconfigure for `ug claude`

src/ucode/cli.py
```diff
@@ -880,23 +876,47 @@ def configure_workspace_command(
...
+        state = configure_selected_tools(
+            state, picked, install_ai_tools=not is_dry_run(), managed=managed
+        )
+    # `ug configure` is the explicit sync: record the applied managed configuration watermark so
+    # later launches skip re-applying (and re-prompting for the OS write) until the admin next edits,
+    # and wire the config's MCP servers/skills now instead of waiting for the first launch.
+    if managed is not None:
+        state = set_applied_managed_update_time(state, managed_update_time(managed))
+        save_state(state)
+        _apply_managed_mcp_and_skills(managed, picked, state)
```
lilly-luo, line 917: what happens if i only configure 1 agent? `ug configure --agents claude`. that means that the timestamp is workspace wide but my codex is still using an old workspace config version?

src/ucode/cli.py
david-siqi-liu: we also write to managed settings during `ug`/`ug claude`/`ug codex` if drift has been detected

src/ucode/cli.py
```diff
@@ -2869,15 +2988,6 @@ def configure(
...
     ] = None,
-    skip_validate: Annotated[
-        bool,
-        typer.Option(
-            "--skip-validate",
-            help="Skip the post-configure validation step that sends a quick test "
-            "message through each agent. Config files are still written with the "
-            "freshly discovered models.",
-        ),
-    ] = False,
```
lilly-luo, line 2880: im worried that clients will use --skip-validate and break. can we mark as deprecated instead but leave the `--skip-validate` cmd so that `No such option: --skip-validate` doesn't throw? it can just do nothing + we can have a deprecation warning

src/ucode/mcp.py
```diff
@@ -1167,82 +1167,56 @@ def _is_app_mcp_server(server: dict) -> bool:
...
-        url_path = "vector-search" if mcp_type == "vector-search" else "functions"
-        name_prefix = (
-            "databricks-vector-search" if mcp_type == "vector-search" else "databricks-functions"
-        )
-        entry_name = _catalog_schema_server_name(name_prefix, catalog, schema, set())
-        return entry_name, f"{workspace}/api/2.0/mcp/{url_path}/{catalog}/{schema}"
-    return None
+    parts = name.split(".")
+    if len(parts) != 3 or not all(parts):
+        return None
```
lilly-luo, line 1181: can you use `_is_uc_fqn` instead and make it public 

https://github.com/databricks/unity-gateway/blob/ce887ad073354d7ab47b105723f4434a3f12076b/src/ucode/managed_setup.py#L267

src/ucode/cli.py
lilly-luo: is that in this PR or a diff PR? it looks like in this PR we never write to managed settings during `configure_tool`?

src/ucode/cli.py
```diff
@@ -2131,24 +2186,43 @@ def _launch_tool(
...
+        # always reconcile the launched tool without a prompt, so a ucode-side change to config
+        # generation (not just an admin edit that bumps update_time) lands on the next launch. The
+        # bulk re-apply to the OTHER enabled agents stays gated on the update_time watermark.
+        applied_ut = get_applied_managed_update_time(existing)
+        if managed is not None:
+            config_changed = refresh or managed_config_is_newer(managed, applied_ut)
+            if not refresh and applied_ut is not None and config_changed:
+                print_note("The managed configuration was updated; re-applying it.")
+            enabled = managed_enabled_tools(managed)
+            to_apply = enabled if config_changed else [tool] if tool in enabled else []
```
lilly-luo, line 2200: is this possible because we're doing all the applies for all tools at once? 

1/ admin sets up claude code 
2/ user uses claude code 
3/ admins sets up codex incorrectly
4/ user goes to run claude code -> is now blocked because codex configuration is wrong

src/ucode/cli.py
```diff
@@ -2111,8 +2163,11 @@ def _launch_tool(
         # Bare `ucode` already fetched one to choose the agent; refetching would double the
         # control-plane round trip and any fallback warning it printed.
         coding_agent_config_feature_disabled = False
+        managed_config_definitively_absent = False
         if managed is None:
-            managed, coding_agent_config_feature_disabled = _fetch_managed_config(state)
+            managed, coding_agent_config_feature_disabled, managed_config_definitively_absent = (
+                _fetch_managed_config(state)
+            )
```
rohita5l, line 2133: Can we fetch the managed config before the auto-configure path? `_launch_tool` has already called `_auto_configure_tool` at this point, and that helper runs discovery followed by `configure_single_tool(tool, state)` without `managed`. On a first launch where the managed policy is the only model source, the unmanaged availability check can fail before we ever reach this fetch. When discovery does succeed, we first write personal config and then configure the same agent again with the managed policy.

Please resolve workspace/auth, fetch the policy once, and pass it into bootstrap/configuration before checking model availability. The first-launch regression test should exercise the real auto-configure helper; currently it is mocked out.

This ordering is inherited from the base branch; it is an integration gap in the intended first-use managed flow, rather than a new ordering regression introduced by this diff.

src/ucode/cli.py
```diff
@@ -2256,19 +2347,23 @@ def _launch_tool(
...
-            route_root_model=route_root_model,
-            # Claude's explicit model is launch-scoped and is passed through LaunchOptions below.
-            custom_model=None,
-            coding_agent_config_defaults=coding_agent_config_defaults,
-            parent_schema=parent_schema,
-        )
+        # The OS-managed (sudo) write is owned by the version-gated apply-all above and by `ug
+        # configure`; a launch only refreshes the user-level config and computes launch params, so
+        # suppress the OS write here to keep an unchanged launch prompt-free.
+        with suppressed_managed_writes():
```
rohita5l, line 2316: This context currently suppresses the OS settings write even when the workspace has never published managed config. The apply block above is skipped in that case, so an existing user can retain stale highest-precedence Claude/Codex settings after switching workspace/provider. That changes the normal launch behavior for users who do not use this feature. The outer agent checks still execute, but the low-level reconcile returns `unchanged` without applying the desired contents.

Can we restrict suppression to current or previously applied workspace policies? The prior-watermark case preserves the explicit `ug revert` behavior after policy removal. This suggestion is self-contained; the `nullcontext` import can also move to the module imports.

```suggestion
        # Preserve normal reconciliation for workspaces that have never had a managed policy.
        # A previously applied policy still requires an explicit `ug revert` when removed.
        from contextlib import nullcontext

        write_context = (
            suppressed_managed_writes()
            if managed is not None or applied_ut is not None
            else nullcontext()
        )
        with write_context:
```

src/ucode/cli.py
```diff
@@ -2131,24 +2186,43 @@ def _launch_tool(
...
         )
-        # An admin-published managed config wins over the developer's own settings. Layered on after
-        # `configure_shared_state`, whose returned state it overrides, and before the provider and
-        # model are settled below — the two state files are never merged on disk.
-        # Bare `ucode` already read one to choose the agent; refetching would double the round trip.
+        # Apply the managed policy to the OS-managed files (the only step that can prompt for a
+        # password). The write content-diffs, so an unchanged file is a silent no-op — that lets us
+        # always reconcile the launched tool without a prompt, so a ucode-side change to config
+        # generation (not just an admin edit that bumps update_time) lands on the next launch. The
+        # bulk re-apply to the OTHER enabled agents stays gated on the update_time watermark.
```
rohita5l, line 2156: Can we make the unchanged managed launch finish after resolving/applying the selected agent once? At present `to_apply = [tool]` still runs `configure_selected_tools` on every launch, and the same tool runs through `configure_tool` again below. The OS file content comparison does not make the entire call a no-op: user config writes, state saves, and provider/model resolution can still repeat. The MCP apply helper already skips unchanged registration entries, but its caller still saves state and prints registration messages; UC location and skill schema listings also sit outside the version gate.

I would separate resolving the effective launch configuration from writing it, then render/reconcile each target once. Keep a version or fingerprint for the applied output, including the workspace and renderer version, so client upgrades and workspace switches still trigger reconciliation. For MCP/skills, define a refresh/cache policy separately since a UC location's contents can change without the policy timestamp changing. A one-line change to `to_apply` alone would miss these lifecycle cases.

src/ucode/cli.py
```diff
@@ -2131,24 +2186,43 @@ def _launch_tool(
...
+            if not refresh and applied_ut is not None and config_changed:
+                print_note("The managed configuration was updated; re-applying it.")
+            enabled = managed_enabled_tools(managed)
+            to_apply = enabled if config_changed else [tool] if tool in enabled else []
+            if to_apply:
+                state = configure_selected_tools(
+                    state, to_apply, install_ai_tools=False, managed=managed
+                )
+            if config_changed:
+                state = set_applied_managed_update_time(state, managed_update_time(managed))
```
rohita5l, line 2169: Confirming Lilly's concern: an update to any enabled agent currently makes the requested launch configure all of them, and `_configure_one` errors abort the entire launch. For example, an unusable Codex provider can prevent `ug claude` from starting even when Claude's own configuration is valid.

Can launch synchronize just the requested agent and record successful application per agent? Explicit configure can still eagerly apply the enabled agents. Because the actual agent/OS files are shared across workspaces, the applied record also needs to identify the workspace currently represented in those files; a historical per-workspace timestamp alone does not establish that they are current. This would also remove the need to configure unselected agents merely to justify a workspace-wide watermark.

src/ucode/cli.py
```diff
@@ -2854,15 +2973,6 @@ def configure(
...
-    skip_validate: Annotated[
-        bool,
-        typer.Option(
-            "--skip-validate",
-            help="Skip the post-configure validation step that sends a quick test "
-            "message through each agent. Config files are still written with the "
-            "freshly discovered models.",
-        ),
-    ] = False,
     skip_unavailable: Annotated[
```
rohita5l, line 2948: Agree with Lilly that we should retain `--skip-validate` as a deprecated no-op. This is already failing the Codex smoke journey: the existing `ug configure ... --skip-validate` invocation exits with `No such option: --skip-validate` before workspace setup. This affects users regardless of whether they have managed config.

The suggestion restores CLI compatibility without restoring the probe. If removing probe validation for unmanaged users is intentional, can that behavior change be split from managed-config application?

```suggestion
    skip_validate: Annotated[
        bool,
        typer.Option(
            "--skip-validate",
            hidden=True,
            help="Deprecated compatibility option; configure no longer sends validation prompts.",
        ),
    ] = False,
    skip_unavailable: Annotated[
```

src/ucode/managed_config.py
```diff
@@ -688,27 +691,27 @@ def refresh_managed_config(state: dict) -> ManagedConfigResult:
...
         fallback = _persisted_fallback(workspace, reason, refused=_is_permission_denied(reason))
-        return ManagedConfigResult(fallback, False)
+        return ManagedConfigResult(fallback, False, False)
     if raw is None:
         # Record that this workspace has no config, rather than leaving an earlier one on disk:
         # the file doubles as the fallback above, so a removed policy would otherwise come back
         # into force after the next transient outage.
         save_managed_state(workspace, {})
-        return ManagedConfigResult(None, False)
+        return ManagedConfigResult(None, False, True)
```
rohita5l, line 711: Small simplification for the no-policy path: we currently create/rewrite `managed-config.json` even when this workspace has never had a policy. We only need an empty tombstone when clearing a nonempty cached policy for this same workspace. Otherwise this adds a disk write and a possible write failure to a feature the user does not use. The workspace-filtered cache reader already gives us the guard below; transient-failure fallback remains unchanged.

```suggestion
    if reason is not None:
        if _is_feature_disabled(reason):
            if load_managed_configuration(workspace):
                save_managed_state(workspace, {})
            return ManagedConfigResult(None, True, True)
        fallback = _persisted_fallback(workspace, reason, refused=_is_permission_denied(reason))
        return ManagedConfigResult(fallback, False, False)
    if raw is None:
        # Clear a previously cached policy so a later outage cannot revive it.
        if load_managed_configuration(workspace):
            save_managed_state(workspace, {})
        return ManagedConfigResult(None, False, True)
```

rohita5l (commented):
Reviewed against two requirements: users without workspace-managed config should retain the existing behavior, and users with managed config should fetch and apply it with minimal setup work.

The small fixes below include applicable code suggestions. Bootstrap ordering and the duplicated/all-agent apply paths need coordinated changes. For the state lifecycle, keeping the raw policy cache separate from personal settings makes sense; please keep the applied-output bookkeeping explicit as well, including which workspace/agent the shared files currently represent. `state['managed_configs']` already means ug-owned config keys, so it should not be used as evidence that a workspace-admin policy was applied.

PR #596, opened 2026-09-13
Consume managed mcp_servers and skills from their UC FQN

david-siqi-liu:
Managed MCP servers and skills never reach the agent. The config stores `mcp_servers.names` and `skills.names` as 3-part UC FQNs (`catalog.schema.name`) — what `serialize_managed_config` writes, what the mcp-services/skills APIs return, and what the coding-agent-config validator (AIGTWY-4604) enforces — but the client's apply paths expected different shapes and silently skipped every entry.

**Changes**
- `mcp.py::managed_mcp_server_entry`: read an `mcp-service` name as a dotted UC FQN; register under the dot-free slug the interactive path uses (dots to dashes) and point the URL at the dotted name, matching `_resolve_mcp_selection`. It previously split on `-` and required the dash form, so a dotted FQN returned `None` and was skipped.
- `skills_download.py::download_managed_skills_on_launch`: read each `skills.names` entry as a 3-part skill FQN (`catalog.schema.skill`) and download that one skill, instead of requiring a 2-part `catalog.schema` and downloading the whole schema. A non-FQN or unknown skill warns and is skipped.
- `cli.py`: update the `_managed_skill_locations` docstring.
- Tests updated to the FQN forms.

**Validate**
- `uv run pytest tests/test_mcp.py tests/test_skills_download.py` passes.
- Behavior: a managed config naming `system.ai.web_search` (mcp-service) now registers `system-ai-web_search` at `/ai-gateway/mcp-services/system.ai.web_search`; a config naming `main.default.triage` (skill) downloads just that skill. Both were previously skipped with a warning.


david-siqi-liu:
Folded into #593 (the launch-apply PR), which is where consuming the managed config's mcp_servers/skills belongs. Closing this standalone PR.

PR #597, opened 2026-09-13
[AIGTWY-4573] Apply managed models and headers to Claude's config

david-siqi-liu:
Resolve the managed config's Claude inputs into Claude Code's own config shape: the model allow-list or discovery location into the `/model` picker, and the custom `http_headers` into Claude's outbound requests. Scoped to Claude; Codex and Gemini per-agent model/header handling is a follow-up so this stays a small, single-agent change.

Changes:
- managed_resolve: add `managed_static_models`, `managed_model_service_location`, and `managed_custom_headers`; extend `managed_state_overrides` to layer `claude_static_models`, the model_service_location, and `claude_custom_headers` into state for Claude.
- claude: a static list becomes `availableModels` + `enforceAvailableModels` + a `modelPicker` (exactly the named ids); a discovery location turns on gateway discovery instead. Ownership markers prune only ug-owned picker keys on a static-to-discovery transition, and clear only when the OS-managed file was written. Managed `http_headers` append to `ANTHROPIC_CUSTOM_HEADERS`, ug's fixed headers winning on conflict; values that would break the newline-delimited format are dropped.
- databricks: shared `extra_custom_headers` helper drops reserved (ug-owned) and delimiter-unsafe headers so ug's headers always win.
- README: document the Claude picker and the ucode-settings.json managed file.


src/ucode/agents/claude.py
```diff
@@ -457,9 +472,26 @@ def render_overlay(
...
+        overlay["enforceAvailableModels"] = True
+        overlay["modelPicker"] = {
+            "replaceBuiltInOptions": True,
+            "options": [{"model": m, "label": _picker_label(m)} for m in static_models],
+        }
+        keys += [[key] for key in CLAUDE_MANAGED_PICKER_KEYS]
+    elif model_service_location and not provider and not relayed:
+        env["CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY"] = "1"
+        keys.append(["env", "CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY"])
+
```
lilly-luo, line 486: we should sync w @andy-xu-db and make sure that model_service_location gets set + propagated correctly in discovery header

src/ucode/agents/claude.py
```diff
@@ -802,26 +912,29 @@ def _reconcile_managed_settings(
     compose: Callable[[dict], dict],
     owned_paths: list[list[str]],
     relayed: bool,
-) -> None:
+) -> bool:
```
lilly-luo, line 915: nit it's a bit unclear what the bool means from the method name `_reconcile_managed_settings`. maybe make it a dataclass 

```
@dataclass 
class ManagedSettingsReconciliationResult
    written: bool 
````

PR #604, opened 2026-09-14
Add an isolated installed-product test harness

rohita5l:
Adds a separate black-box harness that installs a fresh ug wheel or exact release and pinned real agents into isolated environments. It provides bounded subprocesses, fresh homes, real PTY control, transcript/routing evidence, cleanup, and reproducible dependency artifacts.

Includes three credential-free installation checks, standalone pytest configuration, contract checks, editing instructions, and Docker/local reproduction guidance. Live journey definitions are in the next layer; only the three installation cases are claimed here.

Validation: 559 applicable agent/CLI and integration-contract unit tests pass. Ruff and formatting pass. No live gateway pass is claimed.

Layer 2 of 4, based on the production-fixes PR. Merge the stack from bottom to top after the final layer's full CI passes.

PR #605, opened 2026-09-14
Add complete Claude and Codex user journeys

rohita5l:
Defines 45 installed-product journeys for real Claude and Codex, including ten interactive TUI cases: Hosted/MPS configuration, first-prompt and subagent routing, explicit-model bypass, headless prompts, command forwarding, app-server initialization, repeated configuration, revert, and rejected credentials.

Every journey has visible setup and assertions. Completed file tasks, native transcripts, protocol responses, and exit statuses establish success; no mocks, seeded ug/onboarding state, skipped model tasks, or weaker assertions are introduced. Coverage and reproduction documentation describe these cases.

Validation: real pytest collection finds all 45 live cases, with ten TUI cases. Integration contract checks, Ruff, and formatting pass. Live execution is provided by the final stack layer's workflow.

Layer 3 of 4, based on the shared harness. Merge the stack from bottom to top after the final layer's full CI passes.


tests/integration/test_smart_routing_codex.py
lilly-luo, line 1: ill extend this to add the autoconfigure case as well

tests/integration/README.md
```diff
@@ -1,85 +1,208 @@
...
+  --claude-version 2.1.268 \
+  --codex-version 0.154.0 \
+  --profile YOUR_PROFILE
+```
+
+`checkout` builds a wheel and installs it with fresh consumer dependency
+resolution. **It does not use `uv.lock`.** This exercises the install path that
+caught the tomlkit discrepancy in #496. To reproduce a user's release, pass its
+exact distribution version instead, e.g. `--ug-version 0.1.0+f7b4b97`. Use
+`--index-url` for the Python index that contains that release and `--npm-registry`
```
lilly-luo: nit default-index instead of index-url

lilly-luo (approved):

tests/integration/test_smart_routing_codex.py
rohita5l: Acknowledged—this is planned follow-up coverage rather than a requested change. This PR keeps the existing explicit-configure first-prompt journey focused as reviewed.

tests/integration/README.md
rohita5l: Updated the expanded integration README to use `--default-index` while rebasing onto the merged #604 harness in 75853fa.

PR #549, opened 2026-09-10
[AIGTWY-4565] Use the MPS model catalog in Codex

andy-xu-db:
## Summary

- Fetch the selected MPS catalog from `/ai-gateway/codex/v1/models` before launch.
- Send the provider header only in the launch config; persistent and managed Codex config stays MPS-neutral.
- Pass a workspace/provider-scoped atomic catalog through `model_catalog_json`, replacing bundled defaults in `/model`.
- Use the active standard or custom OAuth token for discovery and runtime.
- Fail clearly for invalid, empty, unwritable, or administrator-overridden catalogs.

Server counterpart: https://github.com/databricks-eng/universe/pull/2586683

## Usage

Codex does not use an enable-model-discovery flag:

```bash
uv run ug codex \
  --provider <catalog>.<schema>.<model-provider-service>
```

## Testing

- Focused: 599 passed
- Full suite: 2196 passed, 37 skipped; 2 unrelated installed-binary E2E failures
- Ruff and diff checks passed
- LiteSwap `/model`: exactly `o4-mini`, `o3`, and `o1` for the explicit test MPS
- Allow-all `/model`: exactly 31 compatible models, with no bundled or unsupported extras
- Selected `o3` and `o4-mini`; both returned the expected E2E response through LiteSwap

Codex /model showing only MPS models

PR #611, opened 2026-09-14
Restore Codex provider launch without model discovery

andy-xu-db:
## Summary

- Restore `ug codex --provider <MPS>` when `/ai-gateway/codex/v1/models` is not enabled for the workspace.
- Keep the MPS routing header and launch without a model-catalog override, matching the behavior before #549.
- Preserve authoritative discovery when available and keep auth, permission, invalid-catalog, and other discovery failures fatal.

## Testing

- `uv run pytest tests/test_agent_codex.py tests/test_databricks.py -q` — 353 passed
- `uv run ruff check .`
- Full suite — 2105 passed, 42 skipped, 1 unrelated installed-Codex capture-server failure that reproduces in isolation.

**E2E test**
If codex/v1/models is not available in the workspace, fallback to skip model discovery:
Screenshot 2026-09-14 at 3 54 12 PM


src/ucode/agents/codex.py
```diff
@@ -696,10 +697,18 @@ def launch(
...
-        catalog_path = _model_catalog_path(workspace, provider)
-        _write_model_catalog(catalog_path, catalog)
-        profile_doc["model_catalog_json"] = str(catalog_path)
+        try:
+            catalog = fetch_codex_mps_model_catalog(workspace, token, provider)
+        except CodexMpsModelCatalogUnavailable:
+            print_warning_err(
+                "Codex model discovery is not enabled for this workspace; "
+                "launching without an MPS model catalog. Codex will use its own model selection, "
+                "and the /model picker may include models that are unavailable through this provider."
```
lilly-luo: this is a bit confusing - users didn't see this before and now are seeing this error. there's also not really any action the user can take right because we need to enable this on the server side. 

let's hide this error from the user

andy-xu-db: removed the error message

lilly-luo (approved):

PR #613, opened 2026-09-14
[AIGTWY-4564] Refresh Codex parent model catalogs

andy-xu-db:
## Summary
- fetch a fresh Codex catalog on every `ucode codex --parent` launch
- scope generated catalogs by workspace and parent schema
- pass the catalog through launch-scoped `model_catalog_json` while retaining the parent header for inference
- avoid stale models from the shared five-minute Codex cache

## Testing
- `uv run ruff check .`
- changed-file Ruff format check
- `uv run pytest -q tests/test_agent_codex.py tests/test_databricks.py tests/test_cli.py` (676 passed)
- `uv run pytest -q --ignore=tests/test_e2e_user_agent.py` (2111 passed, 42 skipped)

**E2E Test**
Test shows that cache is cleared and /model is updated when a new parent schema is set
Screenshot 2026-09-14 at 3 37 04 PM
Screenshot 2026-09-14 at 3 37 32 PM


src/ucode/agents/codex.py
```diff
@@ -703,13 +717,32 @@ def launch(
...
+            else:
+                assert parent_schema is not None
+                catalog = _fetch_codex_model_catalog(
+                    workspace,
+                    token,
+                    headers={MODEL_SERVICE_PARENT_SCHEMA_HEADER: parent_schema},
+                    identifier=parent_schema,
+                    kind="Parent schema",
+                )
+                catalog_scope = f"parent:{parent_schema}"
```
lilly-luo: maybe 

```
if provider is not None:
    ....
elif parent_schema is not None:
    ....
else:
    raise Exception(...)
```

src/ucode/cli.py
```diff
@@ -2331,8 +2331,11 @@ def _launch_tool(
                     state["_claude_launch_model"] = launch_model
             if provider:
                 state["_claude_launch_provider"] = provider
-        elif tool == "codex" and provider:
-            state["_codex_launch_provider"] = provider
+        elif tool == "codex":
+            if provider:
+                state["_codex_launch_provider"] = provider
+            elif parent_schema:
+                state["_codex_launch_parent_schema"] = parent_schema
```
lilly-luo, line 2338: this was the main change?

src/ucode/databricks.py
```diff
@@ -3031,22 +3031,29 @@ def build_tool_base_url(tool: str, workspace: str) -> str:
     raise RuntimeError(f"Unsupported tool '{tool}'.")
 
 
-def fetch_codex_mps_model_catalog(workspace: str, token: str, provider: str) -> dict:
+def _fetch_codex_model_catalog(
```
lilly-luo, line 3043: basically the only 2 options that can be passed here are the below right? 
1/ `{MODEL_PROVIDER_SERVICE_HEADER: provider}` 
2/ `{MODEL_SERVICE_PARENT_SCHEMA_HEADER: parent_schema}` 

rn the format makes it hard to deduce and see that there's only 2 possibilites. can you be a bit more explicit about it? you can do something like: 

```
from enum import Enum


class CodexCatalogSource(Enum):
    PROVIDER = (MODEL_PROVIDER_SERVICE_HEADER, "Provider")
    PARENT_SCHEMA = (MODEL_SERVICE_PARENT_SCHEMA_HEADER, "Parent schema")


def _fetch_codex_model_catalog(
    workspace: str,
    token: str,
    *,
    source: CodexCatalogSource,
    identifier: str,
):
    header_name, kind = source.value
    headers = {header_name: identifier}

    # Existing fetch/error handling uses headers, identifier, and kind.
```

lilly-luo (approved):

src/ucode/cli.py
andy-xu-db: yes, so we ensure a cache refresh everytime parent schema is specified

src/ucode/agents/codex.py
andy-xu-db: updated to make options more clear

src/ucode/databricks.py
andy-xu-db: updated to make options more clear

PR #614, opened 2026-09-14
fix: register existing MCPs for agents

larsmoan:
treated a server configured for another agent as already configured, so Claude never received the registration.

Merge existing client bindings with the requested agents and offer missing per-agent registrations.


larsmoan:
Addressed both review findings. The additive merge preserves nameless saved entries and leaves skill MCP client bindings owned by configure skills. Added focused regression tests.

larsmoan:
Addressed the follow-up review. Change accounting uses the complete original server map, while the filtered map only controls picker visibility. Added command-level coverage for adding an existing MCP server to a new agent.

PR #615, opened 2026-09-14
fix: avoid sudo after local ucode updates

larsmoan:
Build metadata in Claude's managed User-Agent changed after every local ucode update, which made each launch rewrite root-owned settings and request sudo.

Use the stable release version in the managed header so local build changes do not rewrite the file.


larsmoan:
Addressed the review finding. Claude uses the shared release-version formatter, and the end-to-end header expectation applies the same rule.

PR #619, opened 2026-09-14
Add experimental `ug claude-cowork` for Claude Desktop via AI Gateway

masonc08:
## What & why

Claude Desktop and its Cowork mode can route through Databricks AI Gateway, but unlike Claude Code (a CLI `ug` spawns, whose lifetime bounds the refresh proxy) Desktop is a standalone GUI with a static config file. Writing tokens into that file leaves them to go stale — the Databricks OAuth token expires in ~1h. So `ug claude-cowork` keeps the refresh-proxy model `ug claude` already uses.

`ug claude-cowork --provider <catalog.schema.name>`:
1. **Establishes both auth sessions** — the Databricks OAuth session (so the proxy can mint swap credentials) and the Anthropic subscription OAuth via `claude setup-token` (long-lived, relayed by the proxy). `CLAUDE_CODE_OAUTH_TOKEN` is honored to skip the browser flow.
2. **Starts the loopback refresh proxy**, which live-refreshes the Databricks credential into the `X-Databricks-AI-Gateway-Token` swap header and stamps the fixed Anthropic `Authorization` + the `Databricks-Model-Provider-Service` routing header on every request. The Desktop client's own credential/auth-scheme is irrelevant: the proxy drops it and owns the upstream auth.
3. **Writes a Claude Desktop gateway config** (`Claude-3p/configLibrary`, one stable entry per workspace) pointing at that proxy — holding **no real tokens**, so nothing on disk can go stale.
4. **Stays foreground** to keep the proxy alive while Desktop is used.

The proxy gains two backward-compatible knobs to support a non-Claude-Code client:
- `extra_headers` — fixed headers the proxy owns and stamps per request (the injected `Authorization` + the MPS routing header).
- `strip_client_headers` — client headers to drop, so a stray `x-api-key` can't reach Anthropic alongside the injected `Authorization`.

## Status / caveats

**Experimental.** The Desktop gateway-config schema (`Claude-3p` `configLibrary`) is not a public contract; the keys here are the observed-working set and may drift across Desktop releases. Discovery relies on `--model` populating `inferenceModels` so Desktop skips its `GET /v1/models` probe, which the relayed path does not serve yet (a companion gateway route change is in flight).

Follow-ups: promote the foreground proxy to a background daemon so Desktop can run independently; refresh the Anthropic OAuth in-proxy; drop the manual `--model` list once relayed `/v1/models` ships.

## How do you know it works?

https://github.com/user-attachments/assets/fa2b838e-ea01-4465-bbee-98124d6de8bf

- `uv run ruff check` / `ruff format` clean.
- New tests: config rendering (points at proxy, no real token on disk), per-workspace config path (macOS/Windows, unsupported-OS raise), `setup-token`/env auth resolution, and the two proxy knobs (`extra_headers` overwrite + `strip_client_headers`).
- Full non-e2e suite green (`uv run pytest -k "not e2e"`) except one pre-existing PTY test (`test_restores_model_captured_immediately_before_switch`) that fails identically on clean `main`.
- `ug claude-cowork --help` registers and parses.


lilly-luo:
can you include a via of it working or what ug claude-cowork will launch?

PR #621, opened 2026-09-14
Pin Codex to the MPS's primary model on provider launch

masonc08:
## Summary
`ug codex --provider <MPS>` writes the MPS's models into Codex's picker (`model_catalog_json`) but left the active model to Codex's own bundled default (e.g. `gpt-5.6-sol`). An MPS allowlist doesn't route that slug, so the first request returned `403 PERMISSION_DENIED: Model '...' is not in the allowed models list`.

When MPS model discovery succeeds, pin the active `model` to the MPS's primary (first) catalog target — unless the user passed `-m`/`--model` or a managed default already applies. Orthogonal to #611: only the discovery-success branch changes; the 404 fallback path stays untouched.

## Changes
- `agents/codex.py`: pin `model` from the fetched catalog in `launch()`; add `_tool_args_select_model()` so an explicit `-m`/`--model` still wins.
- `codex_config.py`: extract `catalog_slugs(catalog)`, shared by the file reader and the launch path.
- Tests: pins first catalog model; respects user `-m`/`--model`; keeps a managed default; existing catalog test asserts the pin.

## Testing
- `uv run --frozen pytest tests/test_agent_codex.py tests/test_codex_config.py tests/test_databricks.py tests/test_lint.py` — 365 passed
- `uv run ruff format --check src/ tests/` and `uv run ruff check src/ tests/` — clean


src/ucode/codex_config.py
```diff
@@ -87,6 +86,11 @@ def _catalog_slugs(path: Path) -> list[str]:
     return slugs
 
 
+def _catalog_slugs(path: Path) -> list[str]:
+    """Extract deduplicated model slugs from a Codex custom catalog JSON file."""
+    return catalog_slugs(read_json_safe(path))
```
lilly-luo: private func call a public func?

src/ucode/agents/codex.py
```diff
@@ -713,6 +718,14 @@ def launch(
             catalog_path = _model_catalog_path(workspace, provider)
             _write_model_catalog(catalog_path, catalog)
             profile_doc["model_catalog_json"] = str(catalog_path)
+            # Codex otherwise boots on its bundled default model (e.g. gpt-5.6-sol),
+            # which an MPS's allowlist doesn't route, so the first request 403s. Pin
+            # the MPS's primary (first) target unless the user chose a model or a
+            # managed default already applies.
+            if not profile_doc.get("model") and not _tool_args_select_model(tool_args):
```
lilly-luo, line 758: where is this custom catalog coming from that has the models?

lilly-luo (approved):

PR #623, opened 2026-09-14
ucode MCP cleanup: dedup cursor mcp.json writers (follow-up to #560)

sunishsheth2009:
## What
Cleanup follow-up to #560. Two independent tidy-ups:

1. **`agents/cursor.py` dedup.** `write_mcp_server_config` (stdio proxy) and `write_http_mcp_server_config` (OAuth HTTP) were byte-for-byte identical except the entry they store, so the read → merge → write logic is factored into a single `_upsert_mcp_server(name, entry)` helper and both writers are thin wrappers.

2. **Move the Claude MCP-registration helpers into `agents/claude.py`.** `add_claude_mcp_server`, `add_claude_http_mcp_server`, and `remove_claude_mcp_server` lived in `ucode.mcp` even though every other agent's registration lives in its own `agents/<agent>.py`. They now sit next to the Claude agent, so `mcp.py` dispatches uniformly (`claude.X`, like `cursor.X` / `opencode.X`).
   - Shared scope constants (`MCP_USER_SCOPE`, `MCP_CLEANUP_SCOPES`) move to the leaf `ucode.constants` so both `mcp` and `agents.claude` import them with **no import cycle** (verified both import orders load clean).
   - `_is_missing_mcp_server_output` stays in `mcp.py` (shared by the codex/gemini removers); `claude.remove_claude_mcp_server` imports it lazily to avoid the `mcp -> agents.claude` load cycle — the same lazy-import pattern `claude.py` already used for these helpers.
   - `claude.py`'s web_search register/unregister now call the local helpers directly.
   - The corresponding unit tests move to `tests/test_agent_claude.py`.

## Why
Removes duplicated JSON-merge logic and puts each agent's MCP-registration code in that agent's module, so `mcp.py` is orchestration-only.

## Testing
`uv run pytest tests/test_mcp.py tests/test_agent_claude.py tests/test_agent_cursor.py tests/test_mcp_oauth.py` (330 passed); full suite green except 7 pre-existing failures unrelated to this change (smart-routing v2 / e2e-user-agent, fail on master too). ruff format/check + ty clean. Behavior-identical.


xsh310 (approved):
LGTM

PR #626, opened 2026-09-15
Retry per-model e2e launches on 429 and skip when saturated

xsh310:
## What changed, and why?

**Change:** The per-model launch e2e tests (codex, claude, opencode, copilot, pi, gemini) now retry a launch with exponential backoff when it is rate-limited (429), and a suite skips instead of failing when every remaining failure is a 429.

**Why:** Each per-model test fires a real inference request at the shared test workspace's ai-gateway, which enforces a per-workspace rate limit. When several CI jobs run their per-model suites at the same time, the combined load trips 429s. Codex's own retry budget expires in about three seconds, too fast for the limit window to reset under that load, so the codex suite fails while the others pass by luck of timing (see run 34801560526, `TestCodexLaunch::test_launch_codex_per_model`). This makes CI flaky whenever multiple PRs run at once.

Details:
- `_run_agent_retrying_rate_limits(...)` wraps `_run_agent` and retries only rate-limit outcomes (`429` / `too many requests` / `exceeded retry limit`) with backoff plus jitter (base 5s, up to 3 attempts). A clean exit or any non-rate-limit failure returns immediately, and `TimeoutExpired` still propagates so each suite keeps its existing timeout handling.
- `_skip_if_all_rate_limited(...)` runs just before each `assert not failures`. If, after retries, every remaining failure is a 429, it skips with a clear message rather than failing, so a saturated shared workspace does not red the CI of unrelated PRs. A single genuine (non-429) failure still fails the assert.

Genuine, non-rate-limit failures continue to fail immediately with no retry, so this does not mask real regressions.

## How do you know it works?

**Testing:**
- `ruff check` and `ruff format --check` on `src/` and `tests/`: clean.
- `pytest --collect-only tests/test_e2e.py`: 39 tests collect.
- A standalone logic harness exercised the helpers: success returns on the first call; two 429s then success takes three calls; a persistent 429 stops after three calls and returns the throttled result; a non-429 failure is not retried; all-429 failures skip; a mixed 429-plus-real failure does not skip; an empty failure list does not skip.

The e2e suite itself only runs against a live workspace (it skips without `UCODE_TEST_WORKSPACE`), so it is exercised by the `e2e` CI job on this PR.


tests/test_e2e.py
```diff
@@ -87,6 +89,49 @@ def _run_agent(
...
+_RATE_LIMIT_MARKERS = ("429", "too many requests", "exceeded retry limit")
+
+
+def _looks_rate_limited(text: str) -> bool:
+    lowered = text.lower()
+    return any(marker in lowered for marker in _RATE_LIMIT_MARKERS)
+
+
+def _run_agent_retrying_rate_limits(
+    cmd: list[str], env: dict | None = None, timeout: int = 60, attempts: int = 3
```
xsh310: Can you make default attempt 5?

xsh310: Done, default is now 5 (af9a3de). Also tightened the comments/docstrings I'd added in the same push.

PR #628, opened 2026-09-15
Fix Codex per-model e2e selection and skip Astra

lilly-luo:
## Summary
- Exclude Astra from the Codex per-model e2e test: Astra has limited allowance in production and will hit 429s if tested.
- By default, ug uses the default model of the harness. Instead, pin Codex to use the specified model in each test iteration.

## Validation
- Ruff and git diff --check passed.
- Live workspace e2e was not rerun.


tests/test_e2e.py
```diff
@@ -490,6 +489,9 @@ def test_launch_codex_per_model(self, tmp_path, monkeypatch, e2e_state, e2e_work
                 codex.write_tool_config(state, model)
 
             cmd = codex.validate_cmd("codex")
+            # By default, ug uses the default model of the harness.
+            # Instead, pin Codex to use the specified model.
+            cmd[1:1] = ["--model", codex.codex_model_id(model)]
```
lilly-luo, line 490: this was bug, we just kept calling the default model on the harness which is astra

PR #631, opened 2026-09-15
Support OpenAI models in Bedrock MPS for Codex

andy-xu-db:
## Summary

- allow Codex to use Bedrock MPS entries that expose OpenAI targets
- keep incompatible Bedrock services filtered out with an actionable error

Fixes the AIGTWY-4565 bugbash failure for main.max-rozen.bedrock.

## Testing

- `ug codex --provider main.max-rozen.bedrock` reaches Codex and `/model` shows only `openai.gpt-oss-20b-1:0`
- focused unit tests: 462 passed
- full suite excluding machine-config-sensitive user-agent probes: 2188 passed, 42 skipped
- ruff check and format check passed


lilly-luo (approved):

lilly-luo:
can u see if there's a way to add this into rohit's new suite of tests?

PR #635, opened 2026-09-15
Rename distribution to unity-gateway and generated helpers to ug (Phase 3)

rohita5l:
Installing from the renamed repository still registers a `ucode` Python distribution. This completes Phase 3 by registering `unity-gateway`, while retaining both `ug` and `ucode` console commands bound to `ucode.cli:main`.

- Read CLI/telemetry versions from the renamed distribution and default source-checkout upgrades to its name.
- Update the installation runner's package lookup and dependency snapshots; archived legacy `ucode` wheels remain supported.
- Document automatic migration, the manual fallback, and recovery after a failed installation. Python imports, saved configuration paths, and telemetry naming stay compatible.
- Update only the root package name in `uv.lock`; dependency versions are unchanged.
- Generate `ug` authentication helpers, MCP bridges/web-search commands, and routing hooks across agent integrations, while keeping existing `ucode` helpers supported. No deprecation warnings are added in this PR.
- Cover canonical helper generation and legacy hook replacement in unit/component tests. Add installed-command protocol checks for both aliases and execute each generated auth helper in the live Claude/Codex Hosted configure journeys.
- Preserve spaces in the integration runner's wheel/constraint paths using file URIs; verified with a real fresh installation under a path containing spaces.

Validation:

- 377 focused tests passed across CLI, entry points, telemetry, doctor, and integration contracts. Ran with the ambient `ENABLE_SMART_ROUTING_V2` variable removed from the test process.
- 1,339 affected unit/component tests passed for helper generation, authentication, MCP, agent configuration/routing, CLI, and entry points; 10 integration-contract/evidence tests passed. Used a short macOS pytest temp path and removed the ambient routing override.
- All 5 credential-free installation tests passed against a freshly built wheel in Python 3.12, installed under a path containing spaces. Both aliases emit exact token-only auth output (normal/forced refresh) and preserve MCP JSON-RPC stdout.
- Isolated `uv tool install .` succeeded, registered `unity-gateway`, and installed both commands; both `--version` commands matched its metadata.
- Ruff lint/format, focused production/new-test type checking, and `git diff --check` passed. An expanded type check also reports 10 diagnostics on unchanged lines in the existing runner/shared protocol harness; those are outside this change.
- All 18 CI checks passed on final head `0d2a536`: the full unit suite, installation checks, gateway API tests, all six agent launch suites, and Claude/Codex smoke and full live integration journeys (including the new generated-helper assertions). [CI run](https://github.com/databricks/unity-gateway/actions/runs/34977667948). Workspace/model e2e tests were not run locally.

Migration notes: installations with the Phase 1 upgrade logic can run `ucode upgrade` or `ug upgrade`. `uv tool upgrade ucode` cannot cross the rename. Older installations need the documented uninstall/install pair. If the replacement installation fails after uninstalling `ucode`, recovery remains manual; this PR does not add automatic rollback.


src/ucode/telemetry.py
```diff
@@ -20,7 +20,7 @@
 @cache
 def ucode_version() -> str:
```
lilly-luo: nit rename this func? ucode_version -> ug_version

lilly-luo (approved):

PR #637, opened 2026-09-15
Remove agent validation; keep --skip-validate as a deprecated no-op

david-siqi-liu:
ug's post-configure probe sent a live test message through each agent and reverted a freshly written config when the probe failed. The AI Gateway already enforces model access at launch, so the probe added latency and could wrongly revert a working config. This removes the probe while keeping the `--skip-validate` flag accepted, so existing scripts and CI that pass it keep working.

Changes:
- Remove the probe: validate_tool, validate_all_tools, provider_permission_error, and the per-agent validate_env / skip_validation hooks.
- Keep `ug configure --skip-validate` as a hidden, accepted no-op; configure no longer validates.
- Keep per-agent validate_cmd: the e2e and unit tests reuse it as the smoke-launch command builder.
- Update tests and docs to match.
- Fix a TUI-timing race in the integration harness (`wait_for`) that the faster configure exit exposed: accept a satisfied predicate on an already-finished process instead of failing on the exit. Without this, test_ug_configure_codex_openai_mps flakes because configure now completes and exits right after the final provider pick.

This supersedes the validation-removal approach in #593, which removed the flag outright and failed e2e because journeys still pass --skip-validate.


tests/test_e2e.py
```diff
@@ -865,64 +861,6 @@ def test_mps_is_listed_usable_and_nonrelayed(self, e2e_workspace, e2e_token):
         assert not entry.get("relayed")
 
 
-class TestGeminiLaunch:
-    """Run gemini against every available gemini model."""
-
-    def test_launch_gemini_per_model(
-        self, tmp_path, monkeypatch, e2e_state, e2e_workspace, e2e_token
```
rohita5l, line 872: Non-blocking: could we retain a real Gemini CLI launch test while removing the production probe? Deleting this class leaves the Gemini launch-test shard running only TestGeminiFreshInstall (config-file assertions) and TestGeminiAuthRecovery (a direct gateway HTTP request); neither launches Gemini. That means a broken Gemini CLI invocation can still leave this shard green.

Suggested change: keep the test's isolation and supported-version guard, but replace validate_tool() with _run_agent(gemini.validate_cmd("gemini"), env=gemini.build_runtime_env(e2e_workspace, model, e2e_token)), asserting a successful exit and nonempty response. This preserves actual CLI smoke coverage without bringing validation back into configure.

rohita5l (commented):
The production validation removal looks good, including retaining --skip-validate as an accepted no-op. The latest commit addresses the fast-exit terminal race seen in the earlier Codex CI run; the new CI run is still in progress. One non-blocking test-coverage comment below.

rohita5l (approved):
Leaving nit comments inline

PR #647, opened 2026-09-15
[AIGTWY-4573] Managed CLI wiring: configure and launch through the managed flow

david-siqi-liu:
Wire the managed config into `ug configure` and the `ug <agent>` launch path.

- Fetch and cache, then branch managed vs personal
- At configure time, apply to all enabled agents (#593 concern 4)
- At launch, version-gate reuse while reconciling the launched agent's config every launch via content-diff (#593 concern 3)
- Add the `definitively_absent` field to the launch-refresh result (`ManagedConfigResult`) so the launch path can distinguish a definitive removal from a transient fetch failure; this field and its tests live here because the launch path is its only consumer

Configure and launch land together because they share the `_fetch_managed_config` helper (defined among the launch functions, called by configure) and the applied-update-time watermark plumbing; splitting them would force a forward-reference or an unused import into an intermediate PR and redden its CI. Not in scope here, tracked as follow-ups: `--workspaces` to a single workspace, and the deeper launch redesign (per-agent fingerprint watermark, no cross-agent abort, true no-op) so this PR stays behavior-preserving.

Integration coverage (existing): this layer wires the flow the whole suite drives. Configure smoke `test_ug_configure_{claude,codex}_databricks`; launch lifecycle `test_ug_configure_{claude,codex}_repeat_and_revert` (repeat apply + version-gate + double revert = the concern-3/4 behavior) and `*_rejects_invalid_credentials`; headless launch `test_ug_{claude,codex}_headless_*`; command forwarding `test_ug_{claude,codex}_*_help` and `test_ug_codex_app_reports_unknown_argument`; protocol `test_ug_codex_app_server_client_initializes`. These pass `--workspaces` and `--skip-validate` (no-op), both preserved here. Unit: tests/test_cli.py, tests/test_managed_config.py. No integration test added or modified (behavior-preserving).


david-siqi-liu:
Closing: re-scoping into a smaller stack focused on making static model lists work for Claude Code and Codex. The rest (tracing, UC-location discovery, MCP, skills, spend defaults, smart routing) follows separately.

PR #616, opened 2026-09-14
Add managed OTLP trace export for Codex

tt-le:
Stacked on #599 (base `tien/managed-otel-traces`) — this diff is just the one Codex commit.

## What
Extend the OTel tracing rail to **Codex**. When the managed `CodingAgentConfig` enables tracing for the codex agent (`AgentConfig.tracing_config.enabled`), `ug codex` exports OTLP trace spans to the AI Gateway ingest endpoint.

## How
- **managed_resolve**: `OTEL_TRACING_TOOLS = ("claude", "codex")` — the per-agent flag now sets `codex_otel_tracing` in state (same rail as Claude).
- **codex**: at launch, `codex.py` mints a fresh Databricks bearer (reusing the `OAUTH_TOKEN` mint) and injects the `[otel.trace_exporter.otlp-http]` block — `endpoint` (full `/v1/traces` path, used verbatim), `protocol = "binary"` (http/protobuf), and `Authorization: Bearer` — as `-c`/`--config` overrides. Nothing is persisted; the token is minted per launch.

## Auth model (verified against openai/codex source)
Codex keeps inference and telemetry auth **fully separate** (`model-provider-info/src/lib.rs` vs `otel/src/config.rs`):
- Inference: `[model_providers.<n>.auth]` command-backed bearer (refreshing).
- OTel: static `[otel...headers]` map only — **no headers-helper, no refresh, no reuse of the inference cred**.
Both use the same Databricks workspace bearer for our gateway; they're just configured in two places.

## Known limitation (stopgap)
Because Codex has no `otelHeadersHelper` equivalent, the bearer is **pinned statically per launch** — it goes stale after ~1h, and (via `-c`) is visible in the process command line. This is an interim manual-testing path. The durable fix is a **local refresh proxy** (the OTLP analog of `ucode mcp-proxy` / `gateway_proxy.py`): point `trace_exporter` at a loopback URL and refresh the bearer server-side, keeping it fresh and out of argv. Planned as a follow-up.

## Validate
- `uv run ruff check .` + format clean; `tests/test_agent_codex.py` + `tests/test_managed_resolve.py` pass (199 focused). Added: codex in the opt-in list, and launch injects/omits the `[otel]` `-c` args by the `codex_otel_tracing` flag.
- End-to-end: a per-agent `tracing_config.enabled` codex stub resolves to `codex_otel_tracing` and produces the correct `--config otel={...}` launch arg (endpoint + `binary` + `Bearer`).
[...]

PR #649, opened 2026-09-15
[AIGTWY-4573] Managed OTLP trace export for Codex

david-siqi-liu:
Extend the managed OTLP tracing rail to Codex. Re-implements the tracing work from #616 (Tien Le, credited as co-author) on top of this stack so #616 becomes obsolete.

- Per-agent tracing flag now covers Codex (`OTEL_TRACING_TOOLS = ("claude", "codex")`)
- At launch, Codex gets the `[otel.trace_exporter.otlp-http]` exporter injected via `-c` overrides pointing at the gateway's traces endpoint

Stopgap (carried from #616): Codex has no headers helper, so the bearer is pinned per launch, goes stale after about an hour, and is visible in argv. The durable fix is a loopback OTLP refresh proxy, tracked as a follow-up.

Integration coverage: documented gap, same as the Claude tracing PR. Live coverage requires admin-enabling tracing on the shared e2e workspace, which the suite's rules forbid faking, so it is not added here. Covered by unit tests: tests/test_agent_codex.py (otlp `-c` block), tests/test_managed_resolve.py. A live tracing CUJ is a scoped follow-up.


david-siqi-liu:
Closing: re-scoping into a smaller stack focused on making static model lists work for Claude Code and Codex. The rest (tracing, UC-location discovery, MCP, skills, spend defaults, smart routing) follows separately.

PR #653, opened 2026-09-15
Make `ug configure` take a single --workspace/--profile

david-siqi-liu:
`ug configure` previously accepted comma-separated `--workspaces` and `--profiles` lists. This collapses each to a single value, since a configure run targets one workspace at a time.

- Renames `--workspaces` to `--workspace` (one workspace URL) and `--profiles` to `--profile` (one Databricks CLI profile).
- Each flag rejects a comma-separated value with an actionable error.
- Keeps `--workspaces` and `--profiles` as hidden, deprecated aliases that accept a single value, so existing scripts keep working (matching the `--skip-validate` / `--skip-upgrade` precedent).
- Updates help text, error messages, docstrings, README, and tests. Confirmed no other command exposed `--workspaces` or `--profiles`.
- Internal state plumbing still accepts a list, so each parser wraps its single entry; retiring the now-unused multi-workspace path is left as a follow-up.


src/ucode/cli.py
```diff
@@ -2869,16 +2872,23 @@ def configure(
         install_databricks_cli()
         if agent is not None and agents is not None:
             raise RuntimeError("Use either --agent or --agents, not both.")
-        if workspaces is not None and profiles is not None:
-            raise RuntimeError("Use either --workspaces or --profiles, not both.")
-        if use_pat and profiles is None:
+        # --workspaces / --profiles are deprecated aliases of the singular flags.
+        if workspace is not None and workspaces is not None:
+            raise RuntimeError("Use either --workspace or --workspaces, not both.")
```
lilly-luo, line 2877: can u assert that the length of workspaces is 1 and otherwise, throw an error?

david-siqi-liu: updated to
```
if len(workspaces) != 1:
    raise RuntimeError(f"Expected exactly one workspace, got {len(workspaces)}.")
```

lilly-luo (approved):

PR #654, opened 2026-09-15
ug mcp login: per-service sign-in status + login for agents' MCP services

sunishsheth2009:
> All dependencies are now merged: `databricks/cli` `auth login --resource` (databricks/cli#6763), plus #717/#731/#679. This branch is rebased on `main` and ready to review. The service-name filter flag is `--names` (matching `ug mcp add`).

## What
Adds `ug mcp login` — sign in to the connection-backed AI Gateway MCP services the coding agents are configured to use. Shows which are already **signed in** vs. **need a per-user connection sign-in**, and runs the sign-in for the ones you pick (interactive picker) or name with `--services` / scope with `--agents`.

## Why
Non-native agents (codex, gemini, opencode) reach connection-backed MCP services (e.g. `system.ai.github`) via `ug mcp-proxy`, which vends tools only after a per-user connection login. Rather than block inside an agent-spawned proxy (fragile), this is an explicit, decoupled login the user runs in their own shell. It also works for Claude/Cursor — the credential is per-user and **shared across all agents**, so signing in once unblocks the service everywhere.

## Reuses the shared enumeration (incl. #717's OS-managed files)
- `list_mcp_command`'s developer + workspace-managed server merge is extracted into a shared `configured_mcp_servers_by_name(state, agents)`; `ug mcp list`, `ug status`, and `ug mcp login` all use it, so they agree on the configured-server set.
- **Managed-file coverage:** #717 delivers Claude/Codex managed servers into the agents' OS-managed files (not `state`). `configured_mcp_servers_by_name` now also reads those files (`claude.read_managed_mcp_urls()` / `codex.read_managed_mcp_urls()`), so `ug mcp login` sees managed mcp-services regardless of how they were delivered — the cross-PR gap is closed here instead of left as a follow-up.

## How it works (existing UC REST APIs — no unshipped deps)
- `GET /mcp-services/<full-name>` → id + `source_connection` (name + `securable_kind`)
- `GET /connections/<conn>/user-credentials/<user>?dependent.mcp_service.id=<id>` → `provisioning_info.state` (`ACTIVE` = signed in; HTTP 404 = needs login)

Works for any connection-backed MCP service — **not just `system.ai.*`** (source connection resolved from the service's own config).

## UX (symmetric with `ug mcp add`)
- bare `ug mcp login` → status table + picker (needs-sign-in pre-checked)
- `--services system.ai.github[,github]` → targeted, non-interactive
- `--agents claude,codex` → scope to those agents' services

Status renders with the same rich `Table` + `status_badge` styling as `ug mcp list`. Sign-in is `databricks auth login --resource <mcp-url>` (databricks/cli#6621), with a clear "upgrade your CLI" error if `--resource` is missing.

## Testing
`tests/test_mcp_login.py` — status classification, the `--resource` invocation incl. the old-CLI clear error, `--services`/`--agents` scoping, workspace-managed inclusion, and **OS-managed-file inclusion**. `tests/test_mcp.py` — `configured_mcp_servers_by_name` merge/scope + the managed-file read (`TestConfiguredMcpServersByName`). `ruff` + `ty` clean; the full mcp/login/cli/agent suite (876) is green.

## Example output

`ug mcp login` (real output) — the status table, then an interactive checklist (needs-sign-in pre-checked) to pick which to sign in to:

```
╭───────────╮
│ MCP login │
╰───────────╯
  Workspace: https://eng-ml-inference.staging.cloud.databricks.com

Connection-backed MCP services
MCP SERVICE       AGENTS         SIGN-IN
system.ai.github  claude, codex  needs sign-in

? Sign in to MCP services (space to toggle, enter to confirm):
 ❯ ◉ system.ai.github  (needs sign-in)
```

Non-interactive form, `ug mcp login --names system.ai.github`, signs in to the named service(s) directly (no picker), and prints `signed in` on success.


xsh310 (commented):
A few correctness notes inline on the new `ug mcp login` code (the changes on top of #557).

src/ucode/mcp_login.py
```diff
@@ -0,0 +1,268 @@
...
+            print_note("No matching MCP services to sign in to.")
+            return 0
+
+    try:
+        token = get_databricks_token(workspace, profile)
+    except Exception as exc:  # noqa: BLE001 - surface auth trouble as guidance
+        raise RuntimeError(
+            f"Could not get a Databricks token for {workspace}: {exc}. Run `ug configure` first."
+        ) from exc
+    user_identity = (_scim_me(workspace, token) or {}).get("userName")
```
xsh310: `userName` is the user's email, but the connection user-credentials API keys on the numeric workspace user id. The webapp uses `String(confx('userId', -1))` and the `connection_user_credentials` proto specifies `UserId`, so a service the user has already signed into 404s here and gets reported as `needs sign-in`. Suggested fix: use `.get("id")` instead of `.get("userName")`.

src/ucode/mcp_login.py
```diff
@@ -0,0 +1,268 @@
...
+            for full, _url, _clients, _managed in entries
+        }
+
+    url_by_full = {full: url for full, url, _clients, _managed in entries}
+    login_needing = [full for full in url_by_full if status_by_full[full] == STATUS_NEEDS_LOGIN]
+
+    # Non-interactive (--names): sign in to every targeted service that needs it.
+    # Interactive: render the per-service status (same Table style as `ug mcp list`), then a picker.
+    if names is not None:
+        targets = [(f, url_by_full[f]) for f in url_by_full if status_by_full[f] != STATUS_NO_LOGIN]
```
xsh310: This keeps already authenticated services (only `NO_LOGIN` is filtered out), so `ug mcp login --names github` re-runs `databricks auth login` and blocks on a browser even when github is already signed in, which is rough for the non-interactive path this flag serves. The comment just above says "every targeted service that needs it", which doesn't match. Suggested fix: filter to `status_by_full[f] in (STATUS_NEEDS_LOGIN, STATUS_UNKNOWN)`.

src/ucode/mcp_login.py
```diff
@@ -0,0 +1,268 @@
...
+    print_section("MCP login")
+    print_kv("Workspace", workspace)
+    with spinner("Checking sign-in status..."):
+        status_by_full = {
+            full: mcp_service_login_status(workspace, token, full, user_identity)
+            for full, _url, _clients, _managed in entries
+        }
+
+    url_by_full = {full: url for full, url, _clients, _managed in entries}
+    login_needing = [full for full in url_by_full if status_by_full[full] == STATUS_NEEDS_LOGIN]
```
xsh310: `login_needing` counts only `NEEDS_LOGIN`, so if the UC API is unreachable and every status comes back `UNKNOWN`, this list is empty, the command prints "All configured MCP services are already signed in", and the picker is skipped, misleading the user. Suggested fix: include `STATUS_UNKNOWN` here, or gate the success message on there being no `UNKNOWN` status either.

src/ucode/mcp_login.py
```diff
@@ -0,0 +1,268 @@
...
+        table.add_column("AGENTS")
+        table.add_column("SIGN-IN")
+        for full, _url, clients, managed in entries:
+            name = full + (" [magenta](managed)[/magenta]" if managed else "")
+            table.add_row(name, ", ".join(clients), _login_status_markup(status_by_full[full]))
+        console.print(table)
+        if not login_needing:
+            print_success("All configured MCP services are already signed in.")
+            return 0
+        selected = _prompt_login_selection(
```
xsh310: `_prompt_login_selection` receives every entry and pre-checks anything that isn't `AUTHENTICATED` (`checked=status != STATUS_AUTHENTICATED`, line 136), so a `NO_LOGIN` service shows up pre-selected and confirming it calls `run_connection_login` on a connection that needs no login. The `label` dict also has no `NO_LOGIN` key, so it renders the raw `no_login_needed` string. Suggested fix: exclude `NO_LOGIN` rows before passing them to the picker.

xsh310 (approved):

src/ucode/mcp_login.py
sunishsheth2009: Good catch — fixed in the latest push. Now resolves the numeric workspace user id via `_scim_me(...).get("id")` (was `userName`/email), so an already-signed-in service no longer 404s and mis-reports as needs-sign-in. Added `test_credential_lookup_keys_on_numeric_user_id_not_email` asserting the cred URL keys on the id, not the email.

src/ucode/mcp_login.py
sunishsheth2009: Fixed. The `--names` path now targets only `STATUS_NEEDS_LOGIN`/`STATUS_UNKNOWN` (via the shared `pending` list), so an already-signed-in service is skipped instead of re-running the browser login. Added `test_services_skips_already_signed_in`.

src/ucode/mcp_login.py
sunishsheth2009: Fixed. Replaced `login_needing` with `pending` = NEEDS_LOGIN or UNKNOWN, and gated the 'all signed in' message + picker on it — so an unreachable UC API no longer masquerades as 'already signed in'. Added `test_unknown_status_is_not_reported_as_all_signed_in`.

src/ucode/mcp_login.py
sunishsheth2009: Fixed. NO_LOGIN rows are filtered out before `_prompt_login_selection`, so they're no longer pre-checked (and can't trigger a pointless sign-in) and the raw `no_login_needed` label can't render. Added `test_picker_excludes_no_login_services`.

PR #656, opened 2026-09-15
[AIGTWY-4573] Managed static model list for Claude Code

david-siqi-liu:

When a workspace's managed coding-agent config sets a static model list for Claude Code
(`enabled_agents[claude].config.models.model_services`), `ug claude`'s `/model` picker now shows
exactly that admin-curated list, in the admin's order, instead of the discovered workspace models.
This lets an admin pin which models developers may pick.

The change is small and additive on top of main's existing managed launch flow (which already
fetches the config and calls `resolve_state`):
- `managed_resolve.py`: `managed_static_models(managed, tool)` reads the `model_services` list, and
  `managed_state_overrides` surfaces it as `<tool>_static_models`.
- `agents/claude.py`: `render_overlay` writes the list into Claude's `availableModels` and
  `modelPicker`, ownership-safe so removing the list prunes only ucode's own keys.

This is the static-LIST path only. It is distinct from the family defaults
(`ANTHROPIC_DEFAULT_*_MODEL`) and from `unity_catalog_location` discovery, both of which are unchanged.

## How do you know it works?

**Before**

**After**

image

image

Unit tests cover `managed_static_models` and the picker rendering, including ownership-safe pruning
when the list is removed. This PR also adds the Claude assertion to the managed-workspace
integration case (`tests/integration/test_ug_configure_managed.py`): configured against the managed
e2e workspace, Claude's `availableModels`/`modelPicker` equal the admin's list and Claude boots to a
real gateway prompt. ty, ruff, and the offline suite pass.


lilly-luo (approved):

lilly-luo (commented):
ug codex <workspace 1>, modelPicker 
ug codex <workspace 2> -> what happens? modelPicker should get wiped out

PR #657, opened 2026-09-15
[AIGTWY-4573] Managed static model list for Codex

david-siqi-liu:

Extends the static model list to Codex: when the managed config sets
`enabled_agents[codex].config.models.model_services`, `ug codex`'s `/model` shows exactly that list.
Codex has no picker allow-list like Claude's, so the list is delivered as a `model_catalog_json`
built offline from the harness's bundled model metadata (validated against the installed Codex
binary), replacing main's prior rejection of a managed catalog.

Re-implements Lilly Luo's #612 approach; she is credited as co-author.

Load-bearing files:
- `agents/codex_catalog.py` (new): builds and validates the catalog offline from the admin's list.
- `agents/codex.py`: writes `model_catalog_json` for the static-list case. The MPS/provider path and
  its model pin are unchanged, and no `-c model=` is pinned for the static-list case so the user can
  still pick from `/model`.
- `managed_resolve.py`: surfaces the list as `codex_static_models`, using the shared
  `managed_static_models` helper introduced in the Claude static-list PR.

## How do you know it works?

Unit tests cover the offline catalog build (metadata preservation, non-GPT defaults, validation) and
the launch wiring. This PR also completes the managed-workspace integration case
(`tests/integration/test_ug_configure_managed.py`) with the Codex assertion: configured against the
managed e2e workspace, Codex's generated model catalog lists exactly the admin's `model_services`
and Codex boots to a real gateway prompt. ty, ruff, and the offline suite pass with the stack applied.


lilly-luo:
lgtm! 
Screenshot 2026-09-15 at 9 42 34 PM

lilly-luo (approved):

PR #659, opened 2026-09-15
[AIGTWY-4573] Apply managed config to enabled agents at configure

david-siqi-liu:

When the workspace publishes a managed coding-agent config, `ug configure` now applies it to every
enabled, available agent instead of prompting the developer to pick agents: the admin dictates the
setup, so the personal agent selector should not appear. Each enabled agent is resolved with the
managed overlay and configured right away, so its MCP, skills, and model wiring land at configure
time rather than only at first launch. With no managed config, the existing personal configure flow
is unchanged for existing users.

- `cli.py` `configure_workspace_command`: after the shared workspace/auth setup, fetch the managed
  config; if one is published, configure each enabled agent that is available on the workspace
  (`resolve_state` + `configure_selected_tools`), confirm, and return; otherwise fall through to the
  unchanged personal flow.
- The managed configure output now reads top-down: it announces the managed config up front (so the
  skipped agent selector is expected), configures each enabled agent naming the file(s) it wrote
  (`configured_paths`, dispatched per agent), then shows the summary box.
- Managed-workspace integration harness this stack builds on: a `managed` pytest marker,
  service-principal token minting in `scripts/run_integration.py`, and a per-agent `Managed config`
  CI job (fed by the `E2E_ADMIN_*` secrets) added to the required integration gate.

This is the base of the static-model stack: it establishes the "is there a managed config?" branch
that the Claude and Codex model-list PRs build on.

## How do you know it works?

**Non-enabled workspace**

image

**Enabled workspace**

image

A unit test asserts `ug configure` on a managed workspace configures every enabled agent, does not
prompt for tools, and returns 0; the existing personal-flow configure tests pass unchanged (no
regression for existing users). This PR also adds the configure-level managed-workspace integration
case (`tests/integration/test_ug_configure_managed.py`): against the managed e2e workspace,
`ug configure` shows no agent selector and prints the managed confirmation. The Claude and Codex
model-list assertions are added by the PRs that introduce that rendering (#656, #657). ty, ruff, and
the offline suite are green.


src/ucode/cli.py
```diff
@@ -781,6 +781,22 @@ def configure_workspace_command(
     state = states[0]
     save_state(state)
 
+    # If the workspace publishes a managed config, the admin dictates the setup, so apply it to every
+    # enabled agent available here rather than prompting the developer to pick. Each agent is resolved
+    # with the managed overlay and configured now (so MCP/skills/model wiring lands at configure time,
+    # not only at launch). Launch still refreshes the launched agent.
+    managed, _ = refresh_managed_config(state)
+    if managed is not None:
```
lilly-luo, line 805: wont break existing users

lilly-luo (approved):

PR #661, opened 2026-09-15
Pin Codex integration tests to a non-Astra model

xsh310:
## What changed, and why?

**Change:** Pin every real inference Codex launch in the integration suite to `system.ai.gpt-5-4-nano`, and make the `model_for_explicit_case` harness helper skip Astra (preferring `system.ai.gpt-5-4-nano`) during model discovery.

**Why:** Astra is Codex's current default model but has a small production allowance and returns HTTP 429 under CI load. Tests that fell back to that default (`test_ug_configure_codex_repeat_and_revert`, the three `test_ug_codex_headless_prompt_*` tests) or discovered Astra first (`test_ug_codex_headless_explicit_model_bypasses_routing`) were failing on 429 rather than on any real regression. This centralizes the model on `system.ai.gpt-5-4-nano`, following the earlier fixes in #628 (`test_e2e.py`) and #633 (`test_ug_configure_codex.py`). The `--help` and parser tests in `test_ug_codex_commands.py` are left untouched since they never request a model.

## How do you know it works?

**Testing:** `ruff check` and `ruff format --check` pass on the three changed files. The touched tests are `pytest.mark.live` and only run against a workspace in the CI integration job, so they are exercised there; the change removes the Astra dependency that was producing the 429 failures.


tests/integration/test_ug_codex_headless.py
```diff
@@ -27,7 +27,15 @@ def test_ug_codex_headless_prompt_argument(live_session, workspace):
...
 
     result = session.run(
-        "codex", "--", "exec", "--skip-git-repo-check", "--json", task.prompt, timeout=180
+        "codex",
+        "--",
+        "exec",
+        "--skip-git-repo-check",
+        "--json",
+        "--model",
+        "system.ai.gpt-5-4-nano",
```
lilly-luo: can u define system.ai.gpt-5-4-nano in a const (maybe a constants file until utils) and then import it in all of these places?

lilly-luo (approved):

tests/integration/test_ug_codex_headless.py
xsh310: Done. Moved it to `CODEX_TEST_MODEL` in `tests/integration/utils/constants.py` and imported it in every place that pinned the literal, including the existing pin in `test_ug_configure_codex.py`.

PR #663, opened 2026-09-15
[AIGTWY-4573] Integration coverage for the managed static model list

david-siqi-liu:

Adds the live integration coverage that proves the managed static model list works end to end,
the one path the offline suite cannot exercise. `test_ug_configure_managed.py` runs against a
dedicated managed e2e workspace that publishes a CodingAgentConfig, and asserts the whole stack:
`ug configure` applies the admin config to every enabled agent with no agent selector, Claude's
generated `availableModels`/`modelPicker` and Codex's generated model catalog list exactly the
admin's static `model_services`, and each agent boots to a real gateway prompt rather than its
own login flow.

Load-bearing pieces for a reviewer:
- `tests/integration/test_ug_configure_managed.py`: the two cases (one per agent, marker
  `managed`, not `live`, so the existing `live` lanes never pick them up). Expected model ids
  mirror the workspace's published config.
- `scripts/run_integration.py`: the managed workspace authenticates as a service principal, so
  the runner mints a short-lived token from the standard `DATABRICKS_CLIENT_ID` /
  `DATABRICKS_CLIENT_SECRET` (M2M tokens expire hourly, so nothing long-lived is stored).
- `.github/workflows/integration.yml`: a new per-agent `Managed config` job supplies those from
  the `E2E_ADMIN_*` secrets and is added to the required `All integration tests` gate.
- `tests/test_integration_contract.py`: the one-CI-agent invariant now also covers `managed`.

## How do you know it works?

The transformation was verified against the real workspace config: fetch plus normalize plus
`resolve_state` plus `render_overlay` produce exactly the admin's lists (Claude opus/sonnet/haiku,
Codex gpt-5-6-sol), and the runner's token minting authenticates as the workspace service
principal. The two cases then run in CI against the managed workspace. Prerequisite: the managed
workspace and the `E2E_ADMIN_WORKSPACE` / `E2E_ADMIN_SP_CLIENT_ID` / `E2E_ADMIN_SP_CLIENT_SECRET`
secrets, which are configured on this repository. Offline unit, ruff, and ty are green.


david-siqi-liu:
Superseded: redistributed the managed integration coverage into the stack instead of a separate top PR. The harness (managed marker, service-principal token minting in the runner, the Managed config CI job) and the no-selector case now live in #659, the Claude static-list assertion in #656, and the Codex static-list assertion in #657, so each test lands in the PR whose behavior it validates.

PR #664, opened 2026-09-16
[AIGTWY-4690] Make Databricks AI Tools install opt-in, not default

david-siqi-liu:
`ug configure` auto-installed Databricks AI Tools for the configured agents (opt-out), which slowed managed launches and could fail (AIGTWY-4690). This makes the install off by default, with two rules:

- Workspace has an admin-managed config: skip the install entirely, even with `--enable`.
- Otherwise: opt in via `--enable-databricks-ai-tools` (the flag pair is kept, not deprecated).

The launch path never installs, and the interactive "Configure MCP servers, skills, and plugins?" prompt is unchanged.


src/ucode/agents/__init__.py
```diff
@@ -89,7 +90,14 @@ def install_databricks_ai_tools_for_agents(tools: list[str], state: dict) -> Non
...
     Gemini and Pi have no ``aitools`` support and are dropped.
     """
-    if state.get("databricks_ai_tools_enabled", True) is False:
+    if not state.get("databricks_ai_tools_enabled"):
+        return
+    # An admin's managed config governs the workspace, so ucode does not
+    # self-install AI Tools under one (may become a managed-config option later).
+    if refresh_managed_config(state).manifest is not None:
+        print_note(
+            "Skipping Databricks AI Tools: your workspace's managed config governs agent setup."
```
rohita5l: Can we avoid printing this ?

rohita5l (approved):

PR #669, opened 2026-09-16
[AIGTWY-4573] Register managed mcp_servers at ug configure

david-siqi-liu:
## Problem

Managed MCP-server registration was a silent no-op. #304 added `apply_managed_mcp_servers`, which read the managed config's `mcp_servers` as a list of `{name, type}` entries, but #590 reshaped `mcp_servers` into the `NamesOrLocation` selector (a `{names, unity_catalog_location}` dict). The list check never matched, so a workspace-published MCP server was never registered with any agent. Registration also ran at launch, for the single launching tool only.

## Change

Register the managed config's `mcp_servers` during `ug configure`, for every enabled MCP-client agent that is installed and configured. Both selector shapes are supported:

- `names`: full `<catalog>.<schema>.<service>` service names, registered exactly.
- `unity_catalog_location`: a `<catalog>.<schema>`, registering every MCP service discovered under it.

Resolution reuses the existing discovery and apply helpers (`setup_mcp_clients`, `_resolve_location_mcp_servers`, `apply_mcp_server_changes`). The resolved set is reconciled against the managed set a prior configure registered (persisted under `managed_mcp_servers`), so a server the admin later drops from the config is unregistered rather than left behind. Even when no MCP client is installed, any prior managed set is still unregistered. The managed set is tracked separately from the developer's own `mcp_servers`, so reconciling never touches a server they configured themselves.

Two correctness points fall out of registering at configure time:

- The per-agent configure loop resolves each agent from a fresh copy of the state, and `configure_selected_tools` persists `available_tools` from that copy. The accumulated set is now carried forward across the loop, so registration sees every enabled agent instead of only the last one configured.
- A managed `names` entry that is not a full three-part FQN (bare, or over-qualified like `a.b.c.d`) is skipped with a warning rather than failing the whole registration, so one malformed entry never blocks the valid ones.

Retires the stale `apply_managed_mcp_servers` and `managed_mcp_server_entry`, and the launch-time registration path.

## Follow-ups (out of scope here)

Deferred so this PR stays focused on the no-op fix; each is a narrow reconcile-semantics edge that predates this change in shape:

- Dropping an agent from the managed enabled set while a server is still wanted for other agents does not unregister that server from the dropped agent (reconcile removes a server only when its whole name disappears from the desired set).
- If the developer independently configured a server whose entry name collides with a managed one, managed reconcile treats the shared physical registration as managed-owned.

## Testing

image

image

image

Local `uv run ruff check`, `ruff format --check`, and `ty check src/` all pass clean. Focused unit tests pass (`tests/test_mcp.py` and `tests/test_cli.py`, 474 passed), covering names resolution, location resolution, strict FQN rejection, reconcile-on-removal, the empty selector no-op, the no-installed-client removal path, the CLI wiring and scope, and a regression test that configure accumulates `available_tools` for every enabled agent. The broader unit suite passes locally aside from environment-dependent tests (package-install and telemetry entry points, and live-workspace e2e) that fail identically on the base branch.


src/ucode/mcp.py
```diff
@@ -735,104 +735,91 @@ def _is_app_mcp_server(server: dict) -> bool:
...
+    names = [n for n in (selector.get("names") or []) if isinstance(n, str) and n]
+    if not names:
+        return []
+    malformed = sorted(
+        n for n in names if n.count(".") != 2 or not all(part for part in n.split("."))
+    )
+    if malformed:
+        raise RuntimeError(
+            "Managed mcp_servers names must be full `<catalog>.<schema>.<service>` names "
+            f"(got: {', '.join(malformed)})."
```
lilly-luo: do we need to hard fail on boot up for malformed mcp servers? why don't we just skip them ? this means a dev needs to reach out to admin to unblock themselves for an admin misconfiguration

src/ucode/mcp.py
```diff
@@ -735,104 +735,91 @@ def _is_app_mcp_server(server: dict) -> bool:
...
         )
-        entry_name = _catalog_schema_server_name(name_prefix, catalog, schema, set())
-        return entry_name, f"{workspace}/api/2.0/mcp/{url_path}/{catalog}/{schema}"
-    return None
+    by_schema: dict[str, set[str]] = {}
+    for name in names:
+        by_schema.setdefault(".".join(name.split(".")[:2]), set()).add(name)
```
lilly-luo, line 773: can you put in comment an example of what by_schema looks like? 

```
by_schema = {
    "system.ai": {"system.ai.slack", "system.ai.github"},
    "main.default": {"main.default.custom"},
}
```

src/ucode/cli.py
```diff
@@ -811,7 +811,15 @@ def configure_workspace_command(
                     [tool_name],
                     install_ai_tools=not is_dry_run(),
                 )
+                # Each iteration resolves from `state` and persists a copy, so carry the
+                # accumulated available_tools forward — otherwise the last agent's save drops
+                # the earlier ones, and the MCP reconcile below only sees that final agent.
+                state["available_tools"] = configured.get("available_tools") or state.get(
+                    "available_tools"
+                )
```
lilly-luo, line 815: ic this refers to claude/codex as the "tools"

lilly-luo (approved):

PR #691, opened 2026-09-16
[AIGTWY-4758] Reconcile managed-config skills at `ug configure`

david-siqi-liu:
## Problem

Managed coding-agent config can declare a `skills` selector (`{names | unity_catalog_location}`), but `ug` never consumed it. `cli` read `skills.names` as if they were `<catalog>.<schema>` locations and ignored `unity_catalog_location`, so both valid selector shapes downloaded nothing: a silent no-op, exactly like the managed `mcp_servers` no-op that #669 fixed.

## Change

Mirror how managed `mcp_servers` are handled: at `ug configure` (alongside MCP registration, off the launch hot path), resolve the `NamesOrLocation` selector correctly and reconcile the workspace's skills to disk.

- `skills_download.py`: `_resolve_managed_skills` resolves the selector into the refs to download plus the desired FQN set (a `unity_catalog_location` lists every finalized skill under that `<catalog>.<schema>`; `names` are full `<catalog>.<schema>.<name>` FQNs; malformed entries are skipped with a warning). `reconcile_managed_skills(managed)` downloads desired skills not already on disk and removes managed skills the config no longer lists. Removal is skipped when the desired set can't be determined (a listing failure), so a transient error never deletes a skill.
- Managed attribution: managed installs are recorded `scope="managed"` in `~/.ucode/skills.json`, so `ug skills remove` never drops an admin skill and a developer's own downloads stay distinct. Re-downloading a managed skill via `ug skills add` transfers it back to the developer.
- `cli.py`: `_configure_managed_skills` mirrors `_configure_managed_mcp_servers`, running on both the managed branch and the no-managed cleanup path (so switching to a workspace with fewer or no skills reconciles the old ones away). Skills land in both `.claude/skills` and `.agents/skills`, so Claude Code and Codex both pick them up.

## Follow-ups

In-place version refresh: a republished skill already on disk is not re-fetched yet. Doing it correctly needs bundle-replacement semantics (the writer adds files but does not delete ones a newer version drops), not just a `uc_update_time` compare, so it is deferred.

## Testing

**Set up**
image

**Claude**

image

image

**Codex**

image

image

Unit: selector resolution (names / location / precedence / malformed / unresolvable / failures / empty / duplicate-FQN dedup), reconcile removal (dropped skill removed, empty selector clears all, listing failure never removes, configured-but-unfetchable kept), `scope="managed"` attribution and `ug skills remove` exclusion, configure wiring and order, and launch making no skill calls.

E2E: the `managed_fixture` integration suite (`tests/integration/test_ug_configure_managed_skills.py`) injects an admin config via `UCODE_MANAGED_CONFIG_STUB`, runs `ug configure`, and launches the real agent. Two cases assert the downloaded skill appears in the agent's `/skills` view for Claude (via the `names` selector) and Codex (via the `unity_catalog_location` selector).

A third case, `test_managed_skills_reconcile_lifecycle`, exercises the managed-vs-personal behavior end to end across four `ug configure` runs: (1) no skills configured downloads nothing; (2) a `unity_catalog_location` config downloads the workspace's skills to disk; (3) a developer's own hand-authored skill (no attribution record) sits alongside them; (4) switching back to a no-skills config removes only the managed skills and leaves the developer's own untouched. That proves the reconcile is attribution-driven: it removes the admin skills the config drops while never deleting a skill ug did not download. Runs in the existing non-blocking managed lanes.

Managed-vs-personal ownership is also covered at the unit level: `scope="managed"` attribution, `ug skills remove` refusing to drop a managed skill, and reconcile keeping a developer's own `scope="user"` install.

`uv run ruff check .` / `ruff format --check` clean; focused unit suites pass locally; the new suite collects.

Part of epic AIGTWY-4329.


src/ucode/skills_download.py
```diff
@@ -292,45 +292,87 @@ def download_selected_skills(workspace: str, token: str, fqns: list[str], path:
...
+    Resolves the ``NamesOrLocation`` selector (see :func:`_resolve_managed_skill_refs`), then writes
+    only skills not already on disk -- no overwrite prompt, so ``ug configure`` never blocks on input
+    and a developer's own same-named skill is never clobbered. The bundles land in both
+    ``.claude/skills`` and ``.agents/skills``, so Claude Code and Codex both pick them up.
+    Best-effort per skill; returns the bundle names newly written.
+    """
+    selector = selector if isinstance(selector, dict) else {}
+    refs = _reject_bundle_name_collisions(_resolve_managed_skill_refs(workspace, token, selector))
+    roots = skill_dir_roots(path)
+    missing = [ref for ref in refs if not existing_skill_on_disk(roots, ref.bundle_name)]
```
xsh310, line 393: This skips any skill whose bundle directory already exists rather than overwriting it, which correctly protects a developer's own skill of the same name. Because there's no `uc_update_time` comparison, though, an admin publishing a new version of a managed skill that's already on disk won't reach developers until they remove the local copy. Is that the intended behavior, or should managed skills refresh in place?

src/ucode/skills_download.py
```diff
@@ -292,45 +292,87 @@ def download_selected_skills(workspace: str, token: str, fqns: list[str], path:
...
-            written.append(ref.bundle_name)
-        record_downloads(_skill_installs(installed, roots, path, workspace))
+        except OSError as exc:
+            # A disk failure on one skill (permissions, full disk) must not strand the rest or
+            # abort configure; the contract is best-effort per skill.
+            print_warning(f"Skipping `{ref.fqn}`: {exc}.")
+            continue
+        installed.append(ref)
+        written.append(ref.bundle_name)
+    record_downloads(_skill_installs(installed, roots, path, workspace))
```
xsh310: These installs are recorded with `scope="user"` in the shared `~/.ucode/skills.json`, so a managed skill can't be told apart from one a developer downloaded themselves. That blocks reconcile-on-removal (the MCP mirror keeps its managed set separate in `managed_mcp_servers`) and lets `ug skill remove` drop an admin skill. Suggest recording managed installs in a separate manifest that mirrors this schema so managed vs personal stay distinguishable.

src/ucode/skills_download.py
david-siqi-liu: Split this into a follow-up. ug configure now reconciles managed skills (it removes ones dropped from the config) and keeps the additive "only write whats missing" behavior, so a republished version does not refresh in place yet. Doing that right needs bundle-replacement semantics, not just a uc_update_time compare: write_skill adds files but never deletes ones a newer version dropped, so a timestamp check alone would leave stale files behind. Tracking it separately.

src/ucode/skills_download.py
david-siqi-liu: Done. Managed installs are now recorded with scope "managed" in skills.json, so ug skills remove leaves them alone and ug configure reconciles them (removing ones the config drops). Kept it in the one manifest rather than a separate file, since a skill is owned one way at a time; a developer can take over a managed skill by re-downloading it with ug skills add, which retags it as their own.

src/ucode/skills_download.py
david-siqi-liu: @xsh310 i think automatic "upgrading" is a nice follow up. for now, we can probably focus on just addition

src/ucode/skills_download.py
david-siqi-liu: @xsh310 i added a lifecycle test
(1) no skills configured downloads nothing; (2) a unity_catalog_location config downloads the workspace's skills to disk; (3) a developer's own hand-authored skill (no attribution record) sits alongside them; (4) switching back to a no-skills config removes only the managed skills and leaves the developer's own untouchedd

xsh310 (approved):

PR #695, opened 2026-09-16
Use Databricks CLI for gated custom OAuth

lilly-luo:
## Summary
- retrieve custom OAuth tokens through `databricks auth token` when `ENABLE_CUSTOM_OAUTH_FROM_CLI=1`
- pass the configured custom client ID through `DATABRICKS_CLIENT_ID` and preserve `--force-refresh`
- require Databricks CLI v1.17.0 only for the gated path; keep the existing SDK flow unchanged otherwise

## Testing
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_custom_oauth.py -q`
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check src/ucode/custom_oauth.py tests/test_custom_oauth.py`

## Version note
PR databricks/cli#6594 is included in Databricks CLI v1.17.0. The requested v0.17.0 predates this feature, so this change uses v1.17.0 as the feature-specific minimum.

```
  "apiKeyHelper": "/tmp/unity-gateway-custom-oauth/.venv/bin/ug auth-token --profile devtools-prod --client-id 4f7af981-8977-4042-b90e-a5dea5b62e19",
```


src/ucode/custom_oauth.py
```diff
@@ -112,17 +131,61 @@ def _custom_oauth_lock(cache_dir: Path, redirect_url: str) -> Iterator[None]:
...
 
 
+def _get_custom_client_token_from_cli(
+    workspace: str,
+    client_id: str,
+    *,
+    profile: str | None = None,
+    force_refresh: bool = False,
+) -> str:
+    ensure_databricks_cli_version(CUSTOM_OAUTH_CLI_MIN_VERSION)
```
rohita5l, line 147: **P1:** `ensure_databricks_cli_version()` may automatically upgrade an older CLI and emits its upgrade warning through the normal UI console. `ug auth-token` is a machine-readable helper whose stdout must contain only the bearer token, so on CLI <1.17 this can prepend status output and make Claude/Codex consume an invalid credential. Please enforce this version during configuration, or make the token path fail on stderr without running the installer.

src/ucode/custom_oauth.py
```diff
@@ -112,17 +131,61 @@ def _custom_oauth_lock(cache_dir: Path, redirect_url: str) -> Iterator[None]:
...
+    env = os.environ.copy()
+    env["DATABRICKS_CLIENT_ID"] = client_id
+    args = ["databricks", "auth", "token"]
+    args.extend(["--profile", profile] if profile else ["--host", workspace])
```
rohita5l: **P1:** This selects `--profile` instead of `--host` whenever a profile is present. `auth_token_cmd` fills an omitted `--profile` from saved UG state even when the caller explicitly passed `--host`, so `ug auth-token --host <workspace-B> --client-id ug-client` can silently run against an `ug-profile` belonging to workspace A. Please preserve the explicit host and validate it against the inherited profile, or inherit a profile only when no host was supplied.

src/ucode/custom_oauth.py
```diff
@@ -71,12 +79,19 @@ def create_custom_oauth_config(
...
+    workspace: str,
+    config: CustomOAuthConfig,
+    profile: str | None = None,
+) -> list[str]:
     normalized = create_custom_oauth_config(
         config["client_id"], config["scopes"], config["redirect_url"]
     )
+    if os.environ.get("ENABLE_CUSTOM_OAUTH_FROM_CLI") == "1":
+        target = ["--profile", profile] if profile else ["--host", workspace.rstrip("/")]
+        return [ug_binary(), "auth-token", *target, "--client-id", normalized["client_id"]]
```
rohita5l: **P2:** The generated helper drops `--scopes` and `--redirect-url` based on the current process's `ENABLE_CUSTOM_OAUTH_FROM_CLI` value, but the flag is not persisted into the command. If a later Claude/Codex process does not inherit that flag, the same helper switches to the SDK path and fails immediately because scopes are missing. Please keep the complete arguments stable and choose the backend when the helper executes.

src/ucode/agents/codex.py
```diff
@@ -186,7 +186,7 @@ def _provider_block(
     custom_oauth: CustomOAuthConfig | None = None,
 ) -> dict:
     if custom_oauth:
-        auth_argv = build_custom_auth_token_argv(workspace, custom_oauth)
+        auth_argv = build_custom_auth_token_argv(workspace, custom_oauth, databricks_profile)
```
rohita5l: **P2:** This forwards `ug-profile` into the generated Codex auth command, but Codex also performs a separate initial token fetch in `_launch_token()`, and that call still invokes `get_custom_client_token()` without `state["profile"]`. With multiple profiles for the same host, startup can fail or resolve a different profile before Codex reaches this configured auth command. Please forward the saved profile in that path too.

src/ucode/custom_oauth.py
```diff
@@ -71,12 +79,19 @@ def create_custom_oauth_config(
...
+def build_custom_auth_token_argv(
+    workspace: str,
+    config: CustomOAuthConfig,
+    profile: str | None = None,
+) -> list[str]:
     normalized = create_custom_oauth_config(
         config["client_id"], config["scopes"], config["redirect_url"]
     )
+    if os.environ.get("ENABLE_CUSTOM_OAUTH_FROM_CLI") == "1":
+        target = ["--profile", profile] if profile else ["--host", workspace.rstrip("/")]
```
rohita5l: **P2:** This emits a profile-only helper (`ug auth-token --profile ug-profile --client-id ug-client`), but `auth_token_cmd` requires a workspace from `--host` or UG state before it invokes the Databricks CLI. A valid CLI profile therefore cannot run this persisted helper after UG state is missing or reset. Please retain `--host` in the generated command or resolve the workspace from the specified profile.

rohita5l (approved):
Approving but please see the P1/P2

src/ucode/custom_oauth.py
lilly-luo: updated to use `databricks_cli_version` instead which does not automatically upgrade

src/ucode/custom_oauth.py
lilly-luo: updated from profile > host to host > profile

src/ucode/custom_oauth.py
lilly-luo: updated to make the arguments stable - ie even if ENABLE_CUSTOM_OAUTH_FROM_CLI = 1, i still pipe --scopes and --redirect uri. this means that subsequent launches of codex/cc without ENABLE_CUSTOM_OAUTH_FROM_CLI explicitly set will still work. this will help in case we need to roll this back

src/ucode/agents/codex.py
lilly-luo: piped profile to codex

src/ucode/custom_oauth.py
lilly-luo: piped host

PR #696, opened 2026-09-16
[AIGTWY-4756] Cache managed config reads for 5 min between launches

david-siqi-liu:
Unity Gateway fetched the workspace managed coding-agent config from the control plane on every non-dry-run launch, even when it had just been read. This reuses a recent read so back-to-back launches stop re-fetching, while `ug configure` always reads fresh.

- Add force_refresh to refresh_managed_config: launches reuse a persisted read younger than a 5-minute TTL. `ug configure` reads fresh instead, both the main path and `ug configure --agent` (whose only managed read is the AI-tools install gate), so it never decides from a since-changed config.
- Persist retrieved_at and an outcome (published, no config, or feature-disabled) as local wrapper metadata, leaving the raw gateway config verbatim for `ug export`.
- Invalidate the cache on a missing, unparseable, or future-dated timestamp, a wrong workspace, or an age at or over the TTL. A failed or auth-failed refresh never advances the timestamp, so a stale read cannot look fresh, and it keeps falling back to the last known-good config.
- Harden read_json_safe against non-UTF-8 bytes so a corrupted cache file read on every launch reads as absent instead of crashing the launch.

Successful no-config and feature-disabled reads are cached for the TTL, so after an admin enables the feature or publishes a config a launch may take up to five minutes to pick it up; `ug configure` reads fresh and is unaffected. Dynamic budget recommendation reads stay per-launch, unchanged.


src/ucode/managed_config.py
```diff
@@ -681,15 +705,44 @@ def managed_state_workspace() -> str | None:
     return workspace if isinstance(workspace, str) and workspace else None
 
 
-def refresh_managed_config(state: dict) -> ManagedConfigResult:
-    """Fetch the workspace's managed config fresh and persist it as a :class:`ManagedConfigResult`.
+def _cached_result_if_fresh(workspace: str) -> ManagedConfigResult | None:
```
lilly-luo, line 708: do u want to check spec version

lilly-luo (approved):

lilly-luo (approved):

src/ucode/managed_config.py
david-siqi-liu: config only gets cached if it passes the spec_version gate in `get_managed_config`, so at this point, we know that the config must have been OK at one point. although, i suppose a user may _downgrade_ ug, at which point the cached one may no longer be valid, but i think that's rare

PR #700, opened 2026-09-17
Polish managed configure output

rohita5l:
## Summary
- use `UG_WORKSPACE` as the fallback for `ug configure` while preserving explicit option precedence
- simplify managed configuration and launch output for a cleaner CLI experience
- keep budget-only managed configs on the normal agent-selection path
- replace Claude web-search discovery text with a generic preparation spinner

## Testing
- `uv run pytest tests/test_cli.py -q` (324 passed)
- `uv run ruff check src/ucode/cli.py tests/test_cli.py`
- `uv run ruff format --check src/ucode/cli.py tests/test_cli.py`
- `git diff --check`


src/ucode/cli.py
```diff
@@ -584,7 +556,7 @@ def configure_shared_state(
         # search (claude only) still needs one Responses-capable model, so fetch
         # just that with a single call.
         if want_claude:
-            with spinner("Fetching web search model..."):
+            with spinner("Preparing Claude Code..."):
                 ws_models, _ = discover_codex_models(workspace, token)
```
rohita5l, line 568: didnt want to say fetching web search model

src/ucode/cli.py
```diff
@@ -584,7 +564,7 @@ def configure_shared_state(
...
```
lilly-luo: why are we preparing cc but discovering codex models....

lilly-luo (approved):

src/ucode/cli.py
```diff
@@ -810,10 +790,13 @@ def configure_workspace_command(
     # A published managed config means the admin dictates the setup: apply it to every enabled agent
     # now rather than prompting the developer to pick.
     managed, _ = refresh_managed_config(state)
-    if managed is not None:
-        _announce_managed_config(managed)
-        for tool_name in managed_enabled_tools(managed):
+    managed_tools = managed_enabled_tools(managed) if managed is not None else []
```
david-siqi-liu, line 793: we should really rename `tool` to `agent`. IMO `tool` is super confusing

src/ucode/cli.py
```diff
@@ -2357,8 +2344,6 @@ def _launch_tool(
             ctx.args = ["--model", relayed_forward_model, *ctx.args]
             forwarded_model = relayed_forward_model
         print_section(_launch_title(tool))
-        if managed is not None:
```
david-siqi-liu, line 2360: why don't we want to print this?

rohita5l: its too verbose

src/ucode/cli.py
rohita5l: makes sense but this doesnt affect end cli output which im trying to clean up in the pr

PR #694, opened 2026-09-16
Derive UG_ENABLE_MODEL_DISCOVERY from managed MPS

andy-xu-db:
## Stack
- **#694 Shared policy**
  - #701 Claude
    - #702 Codex

## Summary
- Set `UG_ENABLE_MODEL_DISCOVERY=1` for the selected agent's managed MPS.
- Set it to `0` otherwise.
- Reject source flags when managed MPS owns routing.
- Scope the variable to the child launch.
- Add no managed-config field.

## Reference
- Follows #668's per-agent managed-policy and launch-scoped environment pattern.
- Reuses `models.model_provider_service`.

## Test
- 444 focused tests passed.
- Ruff and format checks passed.

Testing using this workspace config:
Screenshot 2026-09-17 at 10 42 08 AM

PR #701, opened 2026-09-17
Enable managed model discovery in Claude

andy-xu-db:
## Stack
- #694 Shared policy
  - **#701 Claude**
    - #702 Codex

## Summary
- Map managed discovery to Claude's launch-scoped gateway variable.
- Replace built-in `/model` rows with explicit MPS targets from the existing provider lookup.
- Keep native catalogs for relayed and `allow_all_targets` providers.
- Remove stale ug-owned picker settings when explicit MPS targets no longer apply.

## Test
- 658 focused tests passed.
- Combined stack: 2,383 tests passed, 41 skipped.
- Ruff and format checks passed.

Tested against a workspace-managed Claude MPS.

Managed Claude MPS

https://github.com/user-attachments/assets/97470f54-8240-4595-a3ea-67e3e1d36525

https://github.com/user-attachments/assets/07f00f03-1024-4e1e-b1ad-f6d8020d52a8


lilly-luo:
can u add video of e2e? define in managed config, then load up the models in cc via /model

PR #703, opened 2026-09-17
[AIGTWY-4759] Add fixture-driven managed-config integration tests

david-siqi-liu:
Stacked on #699 (parent). Adds a `managed_fixture` integration layer that injects the admin CodingAgentConfig via the built-in `UCODE_MANAGED_CONFIG_STUB` hook, so the real ug and agent CLIs are exercised across managed-config shapes the live ca-central workspace does not publish. Auth, normalization, config writers, and the agent binaries stay real; the fetch/wire contract stays covered by the un-stubbed `managed` tests.

Seed batch (CUJ table to follow as it expands):
- enabled_agents is an allowlist (single-agent config configures only that agent)
- tracing on/off writes/omits the OTLP telemetry env
- default_agent-not-in-enabled_agents is rejected with a validation error

Policy: sanctioned the input-only stub in tests/AGENTS.md and tests/CLAUDE.md; runs in the existing managed lanes via a `(managed or managed_fixture)` marker filter.


david-siqi-liu:
Rolled into #699 as a single combined PR (both-agents fix + managed tests + the managed_fixture layer).

PR #709, opened 2026-09-17
[AIGTWY-4710] Keep managed models across config transitions AIGTWY-4786

david-siqi-liu:
Keeps a coding agent's managed model settings in sync with the managed config across workspace transitions (AIGTWY-4710). Switching between managed configs reconciles each agent's model settings to the current config, and configuring a workspace with no managed config clears ug's managed model settings so an unmanaged workspace never enforces a stale list inherited from a previous managed workspace.

Lifecycle exercised, per agent (functional in-process tests plus end-to-end integration journeys):
- no managed config: no picker (Claude) / no catalog (Codex).
- static config A: the picker/catalog lists exactly A's models.
- static config B: reconciles to B; models A listed that B omits are pruned.
- model discovery via a Model Provider Service: the picker/catalog is cleared, since the `Databricks-Model-Provider-Service` header routes and there is no static list.
- no managed config again: the picker/catalog is cleared, not inherited.

Changes:
- Claude clears the managed picker (`availableModels`, `enforceAvailableModels`, `modelPicker`) whenever ug is not writing a static list this run (an unmanaged workspace or a provider switch). It always clears the picker in its own `ucode-settings.json`. In the OS `managed-settings.json` the picker is shared with the administrator, so ug treats it as one unit and reverts it to its pre-ucode baseline only when the live picker is unchanged since ug last wrote it (a three-way merge of baseline, ug's last write, and the live file). That way ug removes only what it introduced and never an administrator's own picker, a picker they added after ug cleared its own, or one they later edited (the enforce flag included).
- Codex clears its generated catalog and `model_catalog_json` pointer on an unmanaged configure or a provider switch, so an unmanaged workspace uses its own discovery. A managed static config writes them; the launch-time `model` pin follows Codex's existing clear-when-unmanaged behavior.
- Prettifies the Claude `/model` picker labels: the enforced allow-list still shows one row per model (it can include multiple versions per family and non-Claude models), but each label is now title-cased with a dotted version and known acronyms upper-cased (`system.ai.claude-haiku-4-5` shows as `Claude Haiku 4.5`, `system.ai.glm-5-3` as `GLM 5.3`) instead of the raw id. The routable id still shows on the right of each row.
- The `UCODE_MANAGED_CONFIG_STUB` test hook accepts an explicit JSON `null` to reproduce a workspace with no managed config, distinct from an unset or unreadable stub (which still falls through to the real HTTP read).
- Fixes a pre-existing bug where a managed config's per-family default models were dropped. `managed_resolve._manifest_models` read a stale `models` key instead of the normalized `default_models_by_model_family`, so `managed_claude_family_models` returned nothing and no `ANTHROPIC_DEFAULT_<family>_MODEL` was written (only the overall default and the picker survived, on separate paths). Unit fixtures that hand-built the stale shape masked it; the fix corrects them and adds a regression test that builds through `normalize_managed_config` so the shape can't drift again.
- Reconciles the per-family default env keys the same three-way way as the picker so a workspace switch does not carry stale defaults over. Each `ANTHROPIC_DEFAULT_<family>_MODEL` resolves as config, then an administrator-authored value (one that differs from ug's last write, preserved verbatim), then discovery; a value ug itself wrote is re-derived from the new workspace rather than preserved, and any family default whose model is not in the enforced `availableModels` list is dropped (e.g. a discovered `fable` default when the admin's list omits fable).
- Adds functional lifecycle tests (both agents) covering the full sequence above, including the per-family default env keys landing (opus and sonnet carry ug's `[1m]` context tag), per-agent `managed_fixture` integration journeys running it through `ug configure` against the managed e2e workspace, and coverage-matrix updates. The integration MPS states point at real provider services on that workspace: `main.default.ci_e2e_anthropic_mps` (Claude) and `main.default.ci_e2e_openai_mps` (Codex).

What Claude Code writes for models, by managed model source:
- static (`model_services`): sets `availableModels`, `enforceAvailableModels`, and `modelPicker` in `~/.claude/ucode-settings.json` and the OS `managed-settings.json`. The managed default is applied at launch as `env.ANTHROPIC_MODEL`.
- model discovery via MPS (`model_provider_service`): writes no model list and clears ug's static picker; the `Databricks-Model-Provider-Service` header routes and Claude shows its own model names. A Bedrock-backed provider pins `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL` from the manifest's family slots; an Anthropic provider pins nothing.
- model discovery via `unity_catalog_location`: not consumed on main (schema only; no fields written).

What Codex writes for models, by managed model source:
- static (`model_services`): writes the catalog file `~/.ucode/codex-model-catalog.json` and points `ucode.config.toml`'s `model_catalog_json` at it. The managed default is pinned as `model` in `ucode.config.toml`.
- model discovery via MPS (`model_provider_service`): writes no catalog and clears any static catalog; the `Databricks-Model-Provider-Service` header routes and Codex discovers models at launch.
- model discovery via `unity_catalog_location`: not consumed on main (schema only; no fields written).

Known limitations (out of scope here):
- Clearing on an unmanaged configure also resets in the rare case where the same workspace's managed config genuinely disappeared, since ug does not distinguish that from an intentional switch to an unmanaged workspace.
- On a non-interactive configure (no sudo, `managed_writes_allowed()` False), ug cannot rewrite the root-owned OS `managed-settings.json`, so a stale picker there is not cleared in that path; the private `ucode-settings.json` still is. The interactive flow clears both.
- If ug's managed-file snapshots (its pre-ucode baseline or last write) are unreadable, ug leaves the managed-file picker in place rather than risk clobbering an admin's entry.


src/ucode/agents/claude.py
```diff
@@ -757,7 +757,14 @@ def write_tool_config(
 
     # V2 installs routing hooks in a transient per-launch settings file. Persistent settings must
     # contain no ucode routing hooks; surgically strip legacy ones while preserving user hooks.
-    def _compose(base: dict, *, enforce_model_default_hierarchy: bool) -> dict:
+    def _compose(
+        base: dict,
+        *,
+        enforce_model_default_hierarchy: bool,
+        private: bool,
```
lilly-luo: private -> is_personal_settings_file? or is_managed_settings file and do the inverse? `private` is a bit unclear

src/ucode/agents/claude.py
```diff
@@ -757,7 +757,14 @@ def write_tool_config(
...
+        original: dict | None,
```
lilly-luo: nit original = original_settings

src/ucode/agents/claude.py
```diff
@@ -757,7 +757,14 @@ def write_tool_config(
...
+        last_applied: dict | None,
```
lilly-luo: nit last_applied = last_applied_settings

src/ucode/managed_files.py
```diff
@@ -168,6 +168,36 @@ def _print_managed_write_permission(display: str) -> None:
     _managed_write_notice_shown = True
 
 
+def managed_file_snapshots(tool: str, parser: ManagedParser) -> tuple[dict | None, dict | None]:
```
lilly-luo: can you return a dataclass instead of a tuple? the 2 types are the same and it's easy to mix up the order 

```
@dataclass 
class ManagedFileSnapshots 
     original_before_ug: dict | None 
     last_applied_by_ug: dict | None 
```

src/ucode/managed_files.py
```diff
@@ -168,6 +168,36 @@ def _print_managed_write_permission(display: str) -> None:
...
+    """Return ``(original, last_applied)`` parsed snapshots of ``tool``'s managed file.
+
+    ``original`` is the pre-ucode baseline; ``last_applied`` is what ucode last wrote. A caller
+    reverting a value ucode owns compares the live value against ``last_applied`` (unchanged since
+    ucode wrote it) and restores ``original`` (the value before ucode, if any) as a three-way merge,
```
lilly-luo: a three way merge??? 😱

src/ucode/agents/claude.py
```diff
@@ -816,19 +823,57 @@ def _compose(base: dict, *, enforce_model_default_hierarchy: bool) -> dict:
...
+    )
     write_json_file(
         CLAUDE_SETTINGS_PATH,
-        _compose(read_json_safe(CLAUDE_SETTINGS_PATH), enforce_model_default_hierarchy=False),
+        _compose(
+            read_json_safe(CLAUDE_SETTINGS_PATH),
+            enforce_model_default_hierarchy=False,
+            private=True,
+            original=None,
+            last_applied=None,
```
lilly-luo: add comment about why both of these are None

src/ucode/agents/claude.py
lilly-luo: nit rm private, original, last_applied and instead pass in `managed_settings_snapshots: ManagedFileSnapshots | None`. if it's none, you can infer private = true

lilly-luo:
picker + available model switching is working well for cc. however it doesn't look like the cc family defaults are getting propagated. see https://eng-ml-agent-platform.staging.cloud.databricks.com/ .

is this a known limitation ?

Screenshot 2026-09-17 at 9 54 59 PM
Screenshot 2026-09-17 at 9 55 36 PM

i see this in my managed settings but it doesn't match what's in the config, except for the default. moreover, i didn't define FABLE in the managed config  

```
    "ANTHROPIC_DEFAULT_OPUS_MODEL": "system.ai.claude-opus-4-8[1m]",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "system.ai.claude-sonnet-4-6[1m]",
    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "system.ai.claude-haiku-4-5",
    "ANTHROPIC_DEFAULT_FABLE_MODEL": "system.ai.claude-fable-5-1",
    "ANTHROPIC_MODEL": "system.ai.claude-sonnet-5"
```

lilly-luo (approved):

lilly-luo:
lgtm! 

Screenshot 2026-09-17 at 11 17 38 PM

Screenshot 2026-09-17 at 11 18 06 PM

PR #717, opened 2026-09-17
[AIGTWY-4784] Write managed MCP servers to agents' OS-managed files

david-siqi-liu:
Admin-managed MCP servers were registered into the developer's own user-scope agent config (~/.claude.json, ~/.codex/config.toml), commingling managed and personal servers and drifting when the developer edited their config. This delivers them through each agent's additive OS-managed file instead, so the managed reconcile never touches personal servers.

- Claude: write the resolved set to `managedMcpServers` in the managed-settings.json ug already owns (an HTTP + OAuth entry). Requires Claude Code 2.1.259+.
- Codex: write it to `[mcp_servers]` in /etc/codex/managed_config.toml (the `ug mcp-proxy` stdio command); Codex merges it with the developer's own servers.
- The managed file is the source of truth: each `ug configure` wipes and rewrites ug's entries from the freshly resolved set, so a dropped server disappears and a switch to an unmanaged workspace clears them.
- Fall back to the existing user-scope registration when the managed file cannot be used (non-interactive run, Claude < 2.1.259, PAT auth, or no published OAuth client) and for the other MCP clients (Cursor/Gemini/OpenCode/Copilot), which keep today's behavior.
- `ug mcp list` reads the managed files to keep tagging managed servers; `ug revert` restores them while preserving unrelated admin policy.

Testing: unit tests for the writers, reconcile routing (managed-file vs user-scope, PAT/version fallback, failed-write fallback, wipe-on-switch), and `ug mcp list` tagging. Integration `managed_fixture` CUJs drive `ug configure` through a PTY so ug runs its sudo OS-managed reconcile, and assert the server lands in the managed file with the personal config untouched.


src/ucode/mcp.py
```diff
@@ -800,46 +800,153 @@ def _resolve_managed_mcp_servers(
...
+    would register at user scope. Only servers that name ``agent`` in their ``clients`` are included.
+    """
+    entries: dict[str, dict] = {}
+    for server in servers:
+        name = _server_name(server)
+        url = server.get("url")
+        if not name or not isinstance(url, str) or agent not in (server.get("clients") or []):
+            continue
+        if agent == "claude":
+            entries[name] = claude.managed_mcp_entry(url)
```
sunishsheth2009, line 847: Future-proofing (not a bug today): this writes a direct-HTTP `claude-code` OAuth entry for every managed server URL, whereas the user-scope path (`configure_client_mcp_server`) only goes native when `AIGW_MCP_SERVICES_PATH in url` and otherwise uses the stdio proxy. Today it's equivalent because managed selectors (`_resolve_managed_mcp_servers`) only ever resolve to connection-backed mcp-services FQNs. But if managed MCP ever gains V2 selectors (genie / vector-search / uc-functions / external / app — already selectable in interactive `ug mcp add`), Claude would get a broken native entry instead of the proxy. Consider guarding this Claude branch on `AIGW_MCP_SERVICES_PATH in url` (falling non-mcp-services servers back to the proxy/user-scope) so the two delivery paths stay identical. Codex is unaffected — it's always the proxy command.

src/ucode/mcp.py
```diff
@@ -2027,8 +2134,15 @@ def _collect(server: dict, *, managed: bool) -> None:
 
     for server in state.get("mcp_servers") or []:
         _collect(server, managed=False)
+    # Managed servers also live in the OS-managed files (source of truth), not just fallback state.
     for server in state.get("managed_mcp_servers") or []:
         _collect(server, managed=True)
+    if agents is None or "claude" in agents:
+        for name, url in claude.read_managed_mcp_urls().items():
```
sunishsheth2009: Cross-PR coordination (not a regression on main): `list_mcp_command` and `ug status` now read the OS-managed files as the source of truth — great. But the shared `configured_mcp_servers_by_name` (used by the upcoming `ug mcp login`, PR #654) still reads only `state["mcp_servers"]` + `state["managed_mcp_servers"]`, which no longer contains managed servers delivered via the managed file. Once both land, `ug mcp login` won't list managed mcp-services (low urgency — Claude drives its own OAuth and Codex's proxy triggers login on demand, but pre-provisioning via `ug mcp login` would miss them). I'll fold the managed-file read into the shared enumeration when I rebase #654 — flagging so we keep them in sync.

src/ucode/mcp.py
```diff
@@ -800,46 +800,153 @@ def _resolve_managed_mcp_servers(
...
         )
         working = _resolve_managed_mcp_servers(selector, workspace, profile, clients)
     else:
-        # No enabled MCP client is installed and configured, so there is nothing to register — but
-        # still unregister any servers a prior configure registered (removal is a local no-auth op).
+        # No MCP client configured: nothing to register, but still clear what a prior configure left.
         workspace = state.get("workspace")
         if not workspace:
-            raise RuntimeError("Workspace is not configured. Run `ucode configure` first.")
+            return []
```
sunishsheth2009, line 896: Minor behavior change worth confirming is intentional: this previously `raise`d `"Workspace is not configured..."` when no MCP client was configured and no workspace was set; it now returns `[]` (no-op). This is more forgiving (managed-MCP reconcile stays best-effort and doesn't crash `ug configure`), which seems right — just calling it out.

src/ucode/mcp.py
```diff
@@ -800,46 +800,153 @@ def _resolve_managed_mcp_servers(
...
         profile, clients, working = state.get("profile"), [], []
+    use_pat = bool(state.get("use_pat"))
+
+    eligible = set()
+    if "claude" in scope and claude.managed_mcp_uses_managed_file(workspace, use_pat=use_pat):
```
sunishsheth2009, line 901: Minor: `managed_mcp_uses_managed_file` calls `oauth_client_available(workspace, ...)` (a network probe) on every `ug configure` where Claude is in scope, even when there are no managed servers this run (needed so a switch clears stale managed-file entries). It's cached, so negligible — noting in case the probe cost ever matters.

sunishsheth2009 (approved):

PR #718, opened 2026-09-17
Route Codex OTLP traces through the loopback refresh proxy

tt-le:
## What & why

**Stacked on #719.**

Codex's OTLP exporter accepts only a static header map, so a bearer placed directly in its launch configuration cannot refresh during a long-running session. This change adds the OTLP-specific proxy entry point and routes Codex trace export through it.

Codex exports without credentials to `http://127.0.0.1:<port>/v1/traces`; `start_otel_proxy(...)` injects a current Databricks token and forwards the request to `/ai-gateway/otel/v1/traces`.

- `_otel_token_provider` delegates to Codex's existing `_launch_token` path, preserving the same principal used for inference (custom OAuth when configured, otherwise the selected CLI profile).
- The proxy binds an OS-assigned port (`0`), and Codex receives the actual bound port in its launch overlay.
- Codex explicitly owns the lifecycle: start the proxy and server thread, run Codex as a child, then stop the cache/server/client in `finally`.
- With tracing disabled, the existing exec-based launch path is unchanged.

## Scope

- Covers `ug codex` sessions, not Codex Desktop.
- Traces are attributed to the launching user's credential rather than a shared service principal.

## Testing

- Unit coverage for default-profile and custom-OAuth token selection, forced-refresh propagation, credential-free OTLP configuration, endpoint injection, and launch teardown.
- Real-socket coverage for OTLP forwarding and token injection.
- Focused agent/proxy suite, Ruff, formatting, and `ty` are green.

Sample trace: https://eng-ml-inference.staging.cloud.databricks.com/ml/ai-gateway/governance/traces?o=1653573648247579&selectedTraceId=747dd9a9ce40969547f1882d6cacd27e
Screenshot 2026-09-18 at 1 00 39 AM


src/ucode/agents/codex.py
```diff
@@ -828,25 +836,74 @@ def _reject_managed_model_catalog() -> None:
...
+    The proxy must outlive the launch, so Codex runs as a child rather than
+    exec-replacing this process; mirrors Claude's relayed launch. The proxy binds an
+    OS-assigned port and tears everything down when Codex exits (or fails to spawn).
+    """
+    server, cache, client = gateway_proxy.start_otel_proxy(
+        workspace, _otel_token_provider(state, workspace)
+    )
+    server_thread = threading.Thread(target=server.serve_forever, daemon=True)
+    server_thread.start()
+    endpoint = f"http://{LOOPBACK_HOST}:{server.server_address[1]}/v1/traces"
```
david-siqi-liu, line 877: have we confirmed that `LOOPBACK_HOST` and `server.server_address[1]` both get reset/removed upon workspace switching?

src/ucode/agents/codex.py
```diff
@@ -828,25 +836,74 @@ def _reject_managed_model_catalog() -> None:
...
     }
 
 
+def _otel_token_provider(state: dict, workspace: str) -> Callable[[bool], str]:
+    """The proxy's token_provider: mints from the same source Codex uses for inference
+    (via _launch_token) so exported traces are attributed to the same principal."""
+    return lambda force: _launch_token(state, workspace, force_refresh=force)
+
+
+def _launch_codex_with_otel_proxy(
```
david-siqi-liu, line 860: awful lot of work to get otel started, codex doesn't have a builtin solution for this?

src/ucode/gateway_proxy.py
```diff
@@ -497,3 +497,19 @@ def start_relay_proxy(
...
+
+def start_otel_proxy(
+    workspace: str,
+    token_provider: Callable[[bool], str],
+) -> tuple[ThreadingHTTPServer, TokenCache, httpx.Client]:
+    """Start the Codex OTLP proxy on an OS-assigned port."""
+    return _start_proxy(
+        workspace,
+        token_provider,
+        0,
```
david-siqi-liu, line 510: what is this 0?

david-siqi-liu (approved):

src/ucode/agents/codex.py
tt-le: no, I'm gonna cut a PR to them for this. CC has the `otelHeaderHelper` similar to the `apiKeyHelper` that both support but codex doesn't have this

src/ucode/gateway_proxy.py
tt-le: port number 0, the OS assigns a port when you pass that in

src/ucode/agents/codex.py
tt-le: Let me confirm, but I would assume that this is already the case because this mirrors what we do in CC

tt-le: This server is torn down when we exit codex, and `server.server_address[1]` is just whatever is free at launch time from the OS.

PR #723, opened 2026-09-18
Add `ug skills list` command

xsh310:
## What changed and why

Adds `ug skills list`, a read only command that prints every skill configured for the developer's coding agents and how each one reaches them. Until now there was no single view across the two independent ways a skill gets configured:

* **local**: downloaded to the shared `.claude/skills` and `.agents/skills` directories, so every agent sees it. Tracked in the local `skills.json` manifest.
* **mcp**: a `<catalog>.<schema>` scoped into the skills MCP connection for specific agents. Stored per client in `state.json`.

The command merges the two sources by fully qualified name and renders a table:

```
Configured Skills

NAME                     LOCATION         VIA          AGENTS
ai-ops-confluence-docs   xsh.eng-ai-ops   local        all
ai-ops-confluence-docs   xsh.foo-bar      mcp          claude,codex
triage                   xsh.foo-bar      local,mcp    all

• Use `ug skills add` / `ug skills remove` to change the skills ug configures.
```

* `NAME` is the skill's securable name; `LOCATION` is its `<catalog>.<schema>`.
* `VIA` is `local`, `mcp`, or `local,mcp` when a skill is configured both ways.
* `AGENTS` is `all` for a `local` skill (every agent reads it), or the agents a schema is scoped to for an `mcp` skill (e.g. `claude,codex`).
* Rows are sorted by location, then name, so a schema's skills group together.

Schemas scoped into the MCP connection are expanded to their individual skill names through the UC skills API. A schema whose listing fails still appears as one placeholder row plus a warning, so a network error never hides the local rows. The command touches the network only when there is at least one MCP scope to expand.

## How do you know it works?

* Unit tests in `tests/test_skills_list.py` cover a local only skill, an mcp only schema scoped across multiple agents, a skill configured both ways, sort order (location then name), deduping repeated installs of one skill, the empty state, and a failed schema listing (placeholder plus warning).
* `uv run ruff check` is clean and the full `uv run pytest` suite passes (the workspace only e2e tests skip); `ug skills list --help` renders.

image


src/ucode/mcp.py
```diff
@@ -2111,6 +2111,18 @@ def skill_locations_for_client(entry: dict | None, client: str) -> list[str]:
     return _skill_locations_by_client(entry).get(client, [])
 
 
+def configured_skill_scopes(state: dict) -> tuple[str, dict[str, list[str]]] | None:
```
xsh310: Can you use a clearer function name? term "scope" has been overloaded

src/ucode/skills_download.py
```diff
@@ -518,3 +530,102 @@ def remove_downloaded_skills_command(
...
+        if reason:
+            print_warning(f"Could not list skills in `{location}`: {reason}.")
+            by_fqn[f"{location}.*"] = (f"(skills in {location})", location, agents)
+            continue
+        for ref in refs:
+            by_fqn[ref.fqn] = (ref.bundle_name, f"{ref.catalog}.{ref.schema}", agents)
+    return by_fqn
+
+
+def list_configured_skills_command() -> int:
```
xsh310: Can we move "skill list" logic to a new file? maybe skill_status.py?

src/ucode/skills_download.py
```diff
@@ -518,3 +530,102 @@ def remove_downloaded_skills_command(
...
+        return {}
+    workspace, locations_by_client = scope
+    agents_by_schema: dict[str, set[str]] = {}
+    for client, locations in locations_by_client.items():
+        for location in locations:
+            agents_by_schema.setdefault(location, set()).add(client)
+    if not agents_by_schema:
+        return {}
+
+    install_databricks_cli(minimum=SKILLS_MCP_MIN_DATABRICKS_CLI_VERSION)
```
xsh310: Does the existing version check happens in cli.py or where? Should we follow existing pattern?

src/ucode/skills_download.py
```diff
@@ -518,3 +530,102 @@ def remove_downloaded_skills_command(
...
+@dataclass(frozen=True)
+class ConfiguredSkill:
+    name: str
+    location: str
+    method: str
+    agents: str
+
+
+def _downloaded_skills() -> dict[str, tuple[str, str]]:
+    """Downloaded skills as ``fqn -> (bundle_name, "<catalog>.<schema>")``, one entry per fqn."""
```
xsh310: Why is bundle name needed here?

src/ucode/mcp.py
xsh310: Done. Renamed to `configured_skill_mcp_locations` so it no longer overloads "scope".

src/ucode/skills_download.py
xsh310: Moved it into a new `skills_list.py`. I kept the plural `skills_` prefix to match `skills_download.py` / `skills_state.py` / `skills_api.py`; happy to rename to `skills_status.py` if you prefer.

src/ucode/skills_download.py
xsh310: Good point. Moved `install_databricks_cli(minimum=SKILLS_MCP_MIN_DATABRICKS_CLI_VERSION)` into the `skills_list` wrapper in cli.py, matching `skills add` and `skills remove`.

src/ucode/skills_download.py
xsh310: `bundle_name` is the skill display name (the `name:` an agent reads from SKILL.md, and the on-disk directory name), which the securable leaf in the fqn need not match. It is the NAME column, and using it keeps the downloaded rows consistent with the MCP rows, which show `SkillRef.bundle_name`. Added that note to the docstring.

xsh310: You are right, thanks. Switched the NAME column to the securable name (the fqn leaf) for both downloaded and MCP rows, and dropped `bundle_name` from the module entirely.

src/ucode/skills_list.py
```diff
@@ -0,0 +1,112 @@
...
+
+@dataclass(frozen=True)
+class ConfiguredSkill:
+    name: str
+    location: str
+    method: str
+    agents: str
+
+
+def _downloaded_skills() -> dict[str, tuple[str, str]]:
```
xsh310: Is this helper actually still needed?

src/ucode/mcp.py
```diff
@@ -2111,6 +2111,18 @@ def skill_locations_for_client(entry: dict | None, client: str) -> list[str]:
     return _skill_locations_by_client(entry).get(client, [])
 
 
+def configured_skill_mcp_locations(state: dict) -> tuple[str, dict[str, list[str]]] | None:
```
xsh310: configured_skill_workspace_and_mcp_locations

src/ucode/skills_list.py
xsh310: Removed `_downloaded_skills` and inlined it as a set of fqns; the securable and location now derive from the fqn at the single row-building site.

src/ucode/mcp.py
xsh310: Renamed to `configured_skill_workspace_and_mcp_locations`.

sunishsheth2009 (approved):

PR #724, opened 2026-09-18
Unblock e2e CI: dismiss Codex update notice and pin Claude to 2.1.274

david-siqi-liu:
Two version-driven CI breakages from upstream agent releases, neither a ug bug.

1. Codex update notice. When Codex's pinned version is behind the newest npm release, it shows a blocking "Update available ... Press enter to continue" notice that hangs TUI boot, failing the Codex configure-and-launch integration lanes. `AgentTerminal.boot()` now dismisses it by moving off the default "Update now" to "Skip" before pressing enter (never a digit, so it can't npm-install mid-test). The non-interactive banner variant has no "Press enter to continue" line and renders the real prompt below it, so it falls through to the ready check. Versions stay pinned in integration.yml.

2. Claude opus-5 advisor tool. Claude Code 2.1.275 began sending the `advisor_20260301` tool, which `system.ai.claude-opus-5` rejects (HTTP 400), failing `test_launch_claude_per_model`. The e2e install floated latest; pin it to 2.1.274, the last release before `advisor_20260301`, so opus-5 launches and stays covered. Evidence: #728 (2.1.274) green vs #726 (2.1.275) red. Bump the pin once opus-5 accepts the tool.

This pull request and its description were written by Isaac.

PR #725, opened 2026-09-18
Skip opus-5 in the per-model Claude e2e launch test

david-siqi-liu:
`test_e2e.py::test_launch_claude_per_model` fails only for `system.ai.claude-opus-5`, which returns HTTP 400 "tool type 'advisor_20260301' is not supported for this model". `advisor_20260301` is a beta tool Claude Code sends; opus-5 on the gateway does not accept it.

Evidence this is gateway/model-side, not a version issue:
- The same Claude 2.1.275 passed on master (21:24) and failed on PRs (23:44+); 2.1.276 did not publish until 01:39, so the Claude version was constant across the flip. What changed is opus-5 on the gateway.
- `advisor_20260301` (dated March 2026) is sent by every ug-supported Claude; the minimum, 2.1.248, published Aug 27 2026. So no pinnable version avoids it without dropping below ug's own minimum.

Exclude opus-5 from the per-model launch loop until the gateway accepts that tool for it. Sonnet, haiku, and the provider/relayed launches still cover the launch path. Re-enable by removing the filter once opus-5 supports the tool.


david-siqi-liu:
Folded into #724 (combined into a single PR).

PR #729, opened 2026-09-18
Configure the bare skills MCP on `ug skills` and quiet download registration

xsh310:
## What changed and why

`ug configure skills` is being deprecated, so this adds a simple place to register the bare skills MCP connection: a bare `ug skills`. It also tunes what gets printed when the skills MCP is configured, so a successful registration never buries download failures.

* **Bare `ug skills`** registers the schema-less skills MCP connection (utility tools only) for the configured agents, keeps any client's existing `--mcp` scope, and then prints the group help it always showed. On the first run it also prints a short note on how to create a skill. When a connection already exists it re-registers (still printing the connection summary) but omits that create-skill note.
* **Download paths** (`ug skills add` without `--mcp` or `--skill`, and the interactive picker) no longer print the "Skills MCP registered" summary, which had been burying the per-skill download failures the download step already reports. `register_schemaless_skills_connection` gained a `print_summary` flag (default on) that these paths pass as `False`.
* **`--mcp`** stays exactly as it was.

## How do you know it works?

* New unit tests cover: bare `ug skills` first run (configures, shows help, prints the create-skill note); an already-configured run (re-registers, omits the note); a subcommand skipping the entrypoint's configuration; the three download paths suppressing the summary; and `configure_bare_skills_mcp_command` reporting first run versus existing.
* `uv run ruff check` is clean and `uv run pytest` passes. The one failing test is a pre-existing live-gateway e2e case unrelated to this change.

## Note on the stack

This sits on top of #723 (its base is that PR's `skills-list-command` branch, so the diff is just this one commit). GitHub does not allow a fork PR to join a gh stack, so the two carry no stack badge; once #723 merges, retarget this to `main`.

Screenshot 2026-09-17 at 8 49 56 PM


src/ucode/cli.py
```diff
@@ -1344,6 +1345,32 @@ def _stdin_is_interactive() -> bool:
...
+        first_time = configure_bare_skills_mcp_command()
+    except (RuntimeError, ValueError) as exc:
+        print_err(str(exc))
+        raise typer.Exit(1) from None
+    except KeyboardInterrupt:
+        print_err("Interrupted.")
+        raise typer.Exit(130) from None
+    if first_time:
+        print_note(
+            "To create a skill, start your agent (e.g. `ug claude`) and ask it to create a "
```
xsh310: nit: can we tighten this note and make it coherent and easily understandable by human

xsh310: Tightened to one sentence: "To create a skill, ask your agent to create one with the skills MCP tools, which register it in Unity Catalog." (d8fb371)

xsh310: Named it explicitly per your suggestion: "To create a skill, ask your agent to create one with the Databricks skills registry MCP, which registers it in Unity Catalog." (e368247)

README.md
```diff
@@ -106,6 +106,9 @@ Unity Catalog Skills can be registered as MCP tools or downloaded into local
 agent skill directories.
 
 ```bash
+# Register the skills MCP connection (utility tools only) for your agents.
```
sunishsheth2009: Do we need to mention this in the docs? Basically its quite everywhere, why mention it in the docs then?

sunishsheth2009: Also can ug skills display the help commands for skills?

xsh310: > Do we need to mention this in the docs? Basically its quite everywhere, why mention it in the docs then?
I think we will need to point to this command in the skill creation public doc after we remove `ug configure skills`

> Also can ug skills display the help commands for skills?
it's showing the help commands already

sunishsheth2009 (approved):

README.md
xsh310: Reworded to describe what bare `ug skills` is for rather than the mechanism: "Set up the Databricks skills MCP so your agents can create and manage skills through ug." I also dropped the duplicated "utility tools only" phrasing from the Commands table row so the same description is not repeated in two places. (a1ece72)

PR #741, opened 2026-09-18
ug status: show skill counts and point to `ug skills list` for detail

xsh310:
## What

`ug status` shows one combined `Skills` count per coding agent, alongside that agent's MCP servers count, in place of the `Skills MCP` scope row:

```
Coding Agents

╭─ Claude Code ────────────────╮
│ ...                          │
│ MCP servers: 3               │
│ Skills: 32                   │
│ ...                          │
╰──────────────────────────────╯
```

`Skills` is the distinct skills that agent can reach, combining the locally downloaded skills (visible to every agent) with the skills reachable through the schemas scoped to that agent in the skills MCP connection. `ug skills list` still shows the per skill detail.

Reading the MCP skills is best effort: `ug status` lists each scoped schema, and any schema it cannot read (including an auth failure that stops the read) is left out of the count, so status degrades to a partial count rather than failing.

## Why

Following review feedback: the MCP counts already live per agent in the Coding Agents list, so one combined skill count per agent reads more directly than a separate scope row or splitting local and MCP.

## Implementation

New `configured_skill_counts_by_agent(state, agents) -> dict[str, int]` in `skills_list.py` groups the reachable skills by agent (best effort), reusing the same listing `ug skills list` does. `status()` computes it once and renders `.get(tool, 0)` per agent. The previous `_status_skill_scope` helper is removed.

## Testing

`uv run pytest tests/test_cli.py tests/test_skills_list.py` passes. `TestConfiguredSkillCountsByAgent` covers downloaded skills reaching every agent, combined local and per agent MCP counts, an unreadable schema, and a token failure; `TestStatusSkillsSection` covers the per agent count and that only configured agents show it.

image


src/ucode/cli.py
```diff
@@ -957,6 +933,7 @@ def status() -> int:
             _print_managed_summary(managed, state, None)
 
     print_heading("Coding Agents")
+    skill_reach = configured_skill_reach(state)
```
xsh310: What does skill reach mean and why is this added here?

src/ucode/skills_list.py
```diff
@@ -31,14 +31,32 @@ class ConfiguredSkill:
     agents: str
 
 
+@dataclass
+class SkillReach:
```
xsh310: Why do we need this data class instead of just having a function to get the count

src/ucode/skills_list.py
```diff
@@ -89,7 +106,8 @@ def _configured_skills(state: dict) -> list[ConfiguredSkill]:
             rows.append(
                 ConfiguredSkill(securable, location, _MCP, ",".join(sorted(mcp_agents[fqn])))
             )
-    for location, clients in unlisted_schemas:
```
xsh310, line 92: We don't need to print this warning

src/ucode/cli.py
xsh310: Renamed to `configured_skill_counts_by_agent` (and the local var to `skill_counts_by_agent`). It returns the per agent skill count. It's computed once here because it does one best effort listing of the scoped schemas, then reused for every agent in the loop, the same as the MCP servers count above.

src/ucode/skills_list.py
xsh310: Done. Dropped the dataclass; it's now a plain function `configured_skill_counts_by_agent(state, agents) -> dict[str, int]`.

src/ucode/skills_list.py
xsh310: Removed the warning.

src/ucode/skills_list.py
```diff
@@ -59,7 +60,6 @@ def _mcp_skill_agents(
         catalog, schema = location.split(".")
         refs, reason = list_schema_skills(workspace, token, catalog, schema)
         if reason:
-            print_warning(f"Could not list skills in `{location}`: {reason}.")
```
xsh310, line 62: Add this line back

xsh310: Restored the `print_warning` line (and its import). `_mcp_skill_agents` now matches main again; the new count helper just reuses it.

sunishsheth2009 (approved):

PR #743, opened 2026-09-18
Auto-update downloaded skills at agent launch

xsh310:

**Change:** `ug <agent>` now refreshes a developer's UC-downloaded skills before the agent starts. At launch it reads the download attribution manifest, and for each of the current workspace's own (non-managed) downloads under the home or working directory it calls GetSkill and compares the skill's current `uc_update_time` against the value recorded at download. Any skill whose Unity Catalog source is newer is re-downloaded in place, with no overwrite prompt, since only manifest-attributed directories are touched.

**Why:** A downloaded skill previously drifted from its Unity Catalog source until the developer re-ran `ug skills add`. This keeps downloads current automatically, the way Isaac keeps installed plugins current.

The sweep is rate-limited to once every 24 hours through a new top-level `last_update_check` key in `~/.ucode/skills.json`, matching Isaac's plugin marketplace staleness window, so back-to-back launches make no network calls. It is best effort and fails open, so a network, auth, or fetch error is reported and the launch proceeds on whatever is already on disk. It is skipped under `--skip-preflight`.

## How do you know it works?

**Testing:** Added unit tests covering the new manifest accessors and that recording downloads preserves the `last_update_check` stamp (`test_skills_state.py`), plus the launch path (`test_skills_download.py`): eligibility filtering (home and working dir, current workspace, managed excluded), stale detection by `uc_update_time` (newer, equal, older, missing, deleted), the silent in-place overwrite that refreshes both skill roots and the manifest record, and the entry point's rate-limit gate, fail-open behavior, and end-to-end update. `uv run ruff check`, `ruff format --check`, `ty check src`, and `uv run pytest` all pass locally.


src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+
+
+def _eligible_launch_records(records: list[dict], workspace: str, cwd: str) -> list[dict]:
+    """The downloaded skills a launch from ``cwd`` loads and can refresh itself.
+
+    Only the current workspace's own (non-managed) downloads under the home dir or the
+    working dir qualify. Managed skills are left to ``ug configure``, and other workspaces'
+    downloads are skipped because the launch token authenticates only this workspace.
+    """
+    bases = {os.path.normpath(str(Path.home())), os.path.normpath(cwd)}
```
xsh310: Is this true? I think for claude code at least, the entire every directory in the directory path can have valid skills dirs right? Do you think we can just drop the base check for now for simplicity?

src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+    return [
+        record
+        for record in records
+        if record.get("scope") != "managed"
+        and record.get("workspace") == workspace
+        and os.path.normpath(record.get("base", "")) in bases
+    ]
+
+
+def _stale_launch_refs(
```
xsh310: nit: can you use a more intuitive name? Maybe "get_updated_refs" sth like that?

src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+            if ref is None:
+                continue
+            record = futures[future]
+            stored = record.get("uc_update_time")
+            if stored is None or (ref.uc_update_time or "") > stored:
+                pairs.append((record, ref))
+    return pairs
+
+
+def _apply_launch_updates(workspace: str, token: str, pairs: list[tuple[dict, SkillRef]]) -> int:
```
xsh310: nit: maybe call this update_stale_skills or something more intuitive and human understandable?

src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+    """
+    home = os.path.normpath(str(Path.home()))
+    refs_by_base: dict[str, list[SkillRef]] = {}
+    for record, ref in pairs:
+        refs_by_base.setdefault(os.path.normpath(record["base"]), []).append(ref)
+
+    updated = 0
+    for base, refs in refs_by_base.items():
+        path = None if base == home else base
+        roots = skill_dir_roots(path)
```
xsh310, line 477: Is there an existing download and write util function, if not can we extract one? I think we've rewritten this in a few places?

src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+            except OSError as exc:
+                print_warning(f"Skipping `{ref.fqn}`: {exc}.")
+                continue
+            written.append(ref)
+        record_downloads(_skill_installs(written, roots, path, workspace))
+        updated += len(written)
+    return updated
+
+
+def refresh_downloaded_skills_on_launch(state: dict) -> None:
```
xsh310, line 484: One edge case to consider, if a skill has been manually deleted by user but the record is still in manifest, should we still update this in this case? I think we should just remove the invalid record in that case, wdyt?

src/ucode/skills_download.py
```diff
@@ -412,6 +417,114 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+        roots = skill_dir_roots(path)
+        bundles = _fetch_bundles(workspace, token, refs, label="Updating skills")
+        written: list[SkillRef] = []
+        for ref in refs:
+            files, reason = bundles[ref.fqn]
+            if reason or files is None:
+                print_warning(f"Skipping `{ref.fqn}`: {reason}.")
+                continue
+            try:
+                write_skill(roots, ref, files)
```
xsh310: So this write_skill overwrite the existing files, right?

src/ucode/skills_download.py
xsh310: Good call. Dropped the base filter: `_eligible_launch_records` now refreshes every non-managed download for the current workspace regardless of base, since agents load `.claude/skills` from directories along the path, not only home and the working dir. Simpler, and it also covers project skills in ancestor dirs.

src/ucode/skills_download.py
xsh310: Renamed to `_get_updated_refs`.

src/ucode/skills_download.py
xsh310: Renamed to `_update_stale_skills`.

src/ucode/skills_download.py
xsh310: Extracted `_fetch_and_write(workspace, token, refs, roots, *, label)`, the fetch-then-write core with the per-skill failure skip. It now backs the download path (`_download_refs`), the managed reconcile, and this launch refresh.

src/ucode/skills_download.py
xsh310: Agreed. The refresh now partitions eligible records: any whose directories were deleted by hand are forgotten via `skills_state.forget` (record dropped, nothing re-downloaded), and only the ones still on disk are checked for updates. Reuses the existing `_record_dirs_missing` helper.

src/ucode/skills_download.py
xsh310: Yes. `write_skill` calls `write_bytes` per file, so it overwrites the bundle in place. One nuance: it writes the files present in the new bundle but does not remove files deleted upstream, so such a file would linger until the bundle dir name changes (that path is already reconciled). This is shared behavior across all download paths. Happy to add a clean-before-write in a follow-up if we want strict mirroring.

src/ucode/skills_download.py
```diff
@@ -391,25 +411,105 @@ def reconcile_managed_skills(managed: dict) -> tuple[list[str], list[str]]:
...
+        workspace, token, missing, roots, label="Fetching workspace skills"
+    )
+    record_downloads(_skill_installs(installed, roots, None, workspace, scope="managed"))
+    return [ref.bundle_name for ref in installed], removed
+
+
+# --- Launch-time refresh ---------------------------------------------------
+
+
+def _eligible_launch_records(records: list[dict], workspace: str) -> list[dict]:
```
xsh310, line 424: nit: better function name: maybe eligible_launch_refresh_records?

xsh310: Done, renamed to `_eligible_launch_refresh_records`. I also split this into a 2-PR stack per your other request, so it now lives in #764 (with the `_fetch_bundles_and_write` refactor pulled out into #763). Closing this PR in favor of that stack.

xsh310:
Superseded by a 2-PR stack (splitting the `_fetch_bundles_and_write` refactor out as requested):

- #763 — Extract shared `_fetch_bundles_and_write` skill-download helper (refactor)
- #764 — Auto-update downloaded skills at agent launch (feature, stacked on #763)

Both carry every change from this PR plus the review follow-ups (the two renames, forgetting hand-deleted records, and the shared fetch/write helper). Closing here in favor of the stack.

PR #753, opened 2026-09-18
Propagate managed Claude defaults with model discovery

andy-xu-db:
## Summary

- propagate managed Claude `default_model` with both Unity Catalog and Model Provider Service discovery
- write every explicitly configured Claude family default for both sources
- preserve MPS target IDs verbatim while retaining UC `[1m]` behavior
- keep local and automatically discovered family defaults suppressed for families the managed config omits
- preserve model precedence: explicit `--model`, managed budget recommendation, managed default, then provider fallback
- support Fable-only MPS configurations and relayed MPS private settings

## Context

PR #716 intentionally skipped automatic model resolution for managed discovery sources, but its guards also skipped defaults explicitly supplied by Coding Agent Config. This restores the managed-policy values without changing PR #744 picker behavior or unmanaged provider behavior.

## Testing

- `uv run --frozen pytest tests/test_agent_claude.py tests/test_cli.py tests/test_managed_resolve.py tests/test_databricks.py -q` (938 passed)
- Ruff check and format checks for all touched files
- independent implementation, risk, and test reviews: no remaining findings

Screenshot 2026-09-18 at 6 26 46 PM

https://github.com/user-attachments/assets/452c300d-b9bb-47df-9f8e-87b3075d29ee


src/ucode/agents/claude.py
```diff
@@ -919,6 +919,20 @@ def write_tool_config(
...
+        overlay_env = overlay["env"]
+        for key in CLAUDE_DEFAULT_MODEL_ENV_KEYS.values():
+            overlay_env.pop(key, None)
+        for family, model_id in coding_agent_config_defaults.items():
+            if key := CLAUDE_DEFAULT_MODEL_ENV_KEYS.get(family):
+                overlay_env[key] = (
+                    _maybe_add_1m_suffix(model_id)
+                    if parent_schema and family in ("opus", "sonnet")
+                    else model_id
+                )
```
lilly-luo: can you see if this is done elswhere in the codebase alr? i think we already do it for normal setting. can you reuse that func?

andy-xu-db: updated to reuse _enforce_model_default_hierarchy()

lilly-luo (approved):

PR #764, opened 2026-09-20
Auto-update downloaded skills at agent launch

xsh310:

**Change:** `ug <agent>` now refreshes a developer's UC-downloaded skills before the agent starts. For each of the current workspace's own (non-managed) downloads, the `uc_update_time` recorded at download is compared against the skill's current `update_time` via GetSkill, and any that changed are re-downloaded in place through `_fetch_bundles_and_write` (from #763). A record whose directories the user deleted by hand is forgotten instead of re-downloaded.

**Why:** A downloaded skill previously drifted from its Unity Catalog source until the developer re-ran `ug skills add`. This keeps downloads current automatically, the way Isaac keeps installed plugins current.

The sweep is rate-limited to once every 24 hours via a new top-level `last_update_check` stamp in `~/.ucode/skills.json`, matching Isaac's plugin marketplace staleness window, so back-to-back launches make no network calls. It is skipped under `--skip-preflight` and fails open, so any auth, network, or fetch error is reported and the launch proceeds on whatever is already on disk.

Stacked on #763 (the `_fetch_bundles_and_write` refactor).

## How do you know it works?

**Testing:** New unit tests cover the manifest accessors and that recording downloads preserves the `last_update_check` stamp (`test_skills_state.py`), plus the launch path (`test_skills_download.py`): eligibility filtering, stale detection by `uc_update_time` (newer, equal, older, missing, deleted upstream), a hand-deleted record being forgotten rather than re-downloaded, the silent in-place overwrite that refreshes both roots and the manifest record, and the entry point's rate-limit gate, fail-open behavior, and end-to-end update. `uv run ruff check`, `ruff format --check`, `ty check src`, and `uv run pytest` all pass locally.

https://github.com/user-attachments/assets/8095f2bb-35f6-4ae1-b548-ac5955baf59f

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


sunishsheth2009:
Reviewed with a focus on the launch CUJ, the async/blocking question, and general cleanliness. Overall this is well-structured — clean decomposition (`_eligible_launch_refresh_records` / `_get_updated_refs` / `_update_stale_skills` / entry point), fail-open with `--skip-preflight` bypass, the managed/other-workspace filtering, forget-vs-restore on manual deletion, and the `parse_update_time` de-dup into `string_utils` (with the RFC-3339 sub-second fix + tests) are all nice. A few things worth considering, in rough priority order:

### 1. "async" isn't safe here — but the sync path should be time-bounded
Recording why this is (correctly) synchronous: `launch_agent` ends in `exec_or_spawn`, which on POSIX does `os.execvp` — the agent process *replaces* ucode a few lines after `refresh_downloaded_skills_on_launch` returns. A background thread doing the refresh would be killed the instant `exec` fires; if killed mid-download it leaves **partially-written skill directories** on disk (strictly worse than a stale-but-complete skill). And since agents read skills from disk at startup, doing this synchronously *before* exec is what lets *this* launch pick up fresh skills — async would only ever help the next launch. So sync is the right call.

The real hardening is bounding it. The 24h rate-limit keeps ~every launch to a local file read (good), but the once-a-day launch that *does* run makes `get_databricks_token` + N concurrent `get_skill` + re-downloads on the hot path, and each HTTP call has a **30s timeout** (`skills_api.py`) with no overall budget on the sweep. If UC is slow/hung, that launch can stall for tens of seconds. Consider an overall wall-clock deadline (e.g. 2–3s) so a slow UC degrades to the existing fail-open "skipped" path rather than holding the terminal — 30s-per-call is too long as a launch gate.

### 2. Re-download doesn't actually restore a "mirror" (upstream deletions linger)
The refresh docstrings lean on "mirror" language ("restoring to mirror UC", "re-downloaded to restore the mirror"), but the write path (`_fetch_bundles_and_write` → `write_skill` → `_write_bundle`) writes each file in place with `write_bytes` and **never clears the destination dir first**. So when a skill's UC version *removes* a file, a re-download leaves the stale file on disk — the result is `old ∪ new`, not a true mirror. The partial-delete-restore case works (missing files get rewritten), but the version-change case doesn't fully sync deletions. Either clear the skill dir before writing for a true mirror, or soften the docstrings to "re-writes changed/added files." (Inherited from #763's writer, but this PR is what newly relies on mirror semantics.)

### 3. `set_last_update_check` is stamped before the work runs
The stamp is written before the fetch, and a per-skill `get_skill` failure returns `None` → that skill is silently skipped. Combined, a transient error on a skill means it stays stale for a full 24h with no signal. That's a defensible trade-off (avoids hammering a broken UC on every launch), just worth being intentional about — stamping only on a fully-successful sweep would retry sooner at the cost of retry-on-every-launch when UC is down.

### Minor / CUJ polish
- "Updated N downloaded skill(s)" is shown for a partial-delete *restore* too, where nothing changed upstream — slightly misleading wording for that path.
- When everything is already current, the user sees "Checking Unity Catalog for downloaded skill updates..." with no closing line before the agent boots. Fine (noise-avoidance), just noting the dangling message.
- Nit: `record_downloads` is called once per base inside the `_update_stale_skills` loop — each is a manifest read-modify-write. Could batch into one write; low priority.

None of these are blockers. #1 (bound the sync path) and #2 (mirror vs. write-in-place) are the two I'd most want addressed or explicitly acknowledged.

xsh310:
Thanks for the careful pass, Sunish. Responses on the three:

**1. Overall wall-clock budget on the sync sweep.** Worth having eventually, but not a blocker. Every network call already carries a per-call timeout (30s for get_skill / list_skill_files / fetch_skill_file, 15s for the token fetch), and a timeout is safe: downloads are all-or-nothing (fetch_skill_bundle returns None on any file failure) and the write runs only on a fully in-memory bundle, so a slow or failed call skips that skill and leaves the existing complete copy untouched, never a partial write. With the 24h rate-limit (almost every launch is a local file read with no network) plus the fail-open handler, the only residual is worst-case latency on the once-a-day run when UC is broadly slow. I'd rather land the overall budget as a follow-up if it proves annoying in practice than add the plumbing now.

**2. Mirror vs write-in-place.** It is already a true mirror. write_skill clears the destination first (unlink for a symlink, rmtree for a dir) and only then calls _write_bundle, so a file removed upstream does not linger. That clear-first came in with #767, now in the base, so the "mirror" wording holds. _write_bundle on its own does write in place, but its write_skill wrapper does the clearing, so the result is old fully replaced by new.

**3. Stamp written before the sweep.** Intentional. Stamping up front is what keeps a broken or slow UC from being re-hit on every launch; the alternative (stamp only on full success) means retry-on-every-launch during any UC outage, which reintroduces the per-launch network cost the design avoids. The trade-off is that a transiently-failed skill stays on its prior working version for up to 24h until the next sweep, which felt right for a best-effort background refresh.

The minor nits (the "Updated N" wording on a restore, the dangling "Checking..." line, batching record_downloads) are all fair and I'm happy to fold them in as small follow-ups since none affects correctness.

sunishsheth2009 (approved):

PR #767, opened 2026-09-20
Replace skill bundles atomically instead of merging over them

xsh310:

**Change:** `write_skill` now replaces a skill's bundle directory instead of writing the new files on top of it. Each bundle is staged in a sibling temp dir and swapped into place atomically (`_replace_dir`).

**Why:** The writer merged: it wrote every file from the new bundle but never removed files that were no longer in it, so a file deleted from a skill in Unity Catalog lingered on disk after a re-download. It also wrote in place, so an interrupted download could leave a half-written skill. Staging then swapping makes the on-disk copy a true mirror of UC (dropped files vanish) and guarantees an agent never sees a partial skill.

This is the shared writer for every download path, so `ug skills add` re-downloads get the same mirror semantics; the launch-time auto-update stacked on top (#764) relies on it.

## How do you know it works?

**Testing:** Added `TestWriteSkill` cases for the deleted-upstream-file case, an empty bundle leaving the existing copy untouched, replacing a symlinked bundle without following it to its target, and leaving no staging directories behind. `uv run ruff check`, `ruff format --check`, `ty check src`, and `uv run pytest` all pass locally.


src/ucode/skills_download.py
```diff
@@ -114,14 +116,33 @@ def should_download_skill(roots: list[Path], ref: SkillRef) -> bool:
...
+    written SKILL.md. Each bundle is staged in a sibling temp dir and swapped in atomically, so an
+    interrupted write never leaves a partial skill and a file removed upstream does not linger.
     """
+    if not files:
+        return
     for root in roots:
-        _write_bundle(root / ref.bundle_name, ref.bundle_name, files)
+        target = root / ref.bundle_name
+        target.parent.mkdir(parents=True, exist_ok=True)
+        staging = Path(tempfile.mkdtemp(prefix=f".{ref.bundle_name}.", dir=target.parent))
```
xsh310: Should we add the temp dir in the same root directory or in a specific tmp directory? Also should we add a timestamp to the tmp dir name to prevent the initial try and retry land into the same tmp dir?

src/ucode/skills_download.py
```diff
@@ -114,14 +116,33 @@ def should_download_skill(roots: list[Path], ref: SkillRef) -> bool:
     return True
 
 
+def _replace_dir(target: Path, source: Path) -> None:
```
xsh310: Is this replacement actually atmoic?

sunishsheth2009 (approved):

PR #770, opened 2026-09-21
Handle stale MCP cleanup failures when switching workspaces

rohita5l:
Switching workspaces carries the old skills MCP registration into the new workspace's state. Cleanup can then attempt the same removal twice, and a Claude CLI timeout or OS error escapes as a traceback after the stale-entry warning. Deduplicate removal attempts by client/server and report recoverable cleanup failures as warnings so configuration can continue.

Adds regression tests for both cleanup passes, continued cleanup of other clients, persistence of the new workspace, and preservation of the historical workspace bucket. The original screenshot omitted the final exception; these tests reproduce the confirmed code defects without claiming to identify that user's exact exception.

Adds an installed-product CUJ, `test_ug_configure_claude_cleans_stale_skills_mcp_on_workspace_switch`: configure the first real workspace, register its skills MCP, switch to a second real workspace, verify removal through `claude mcp get` and saved state, repeat configure, and complete a real Claude file-reading task. No mocks, fabricated ug state, or injected managed configuration. The CUJ covers the normal transition; deterministic timeout coverage remains in unit/component tests.

The runner accepts an explicit second workspace and bearer, redacts both credentials, and runs this CUJ in the existing **Managed config · Claude** lane using the two existing CI workspace credential sets. That lane retains its existing non-blocking policy.

Validation:

- Before the fix: six focused failures exposed duplicate attempts and uncaught timeout/missing-executable errors.
- After the fix: 805 targeted unit/component tests passed with inherited routing/workspace settings isolated.
- After CUJ wiring: 16 regression/contract tests passed; the new CUJ and its full CI lane collect successfully.
- Ruff lint/format checks, `git diff --check`, runner help, and both-workspace credential redaction checks passed.
- Live CUJ execution is pending CI; collection is not a live pass.


david-siqi-liu (approved):
Approved to unblock, can you check if skills has the same issue?

lilly-luo (approved):

